PR #135, opened 2026-03-21
fix: make decode timeout configurable, raise default to 5 minutes

ecmulli:
## Summary

The hard-coded 30-second timeout in `decodeAllCollectionsIsolated()` causes the decode worker to crash on databases over ~500MB. The server copies the entire LevelDB to `/tmp` then iterates and parses every protobuf entry — on a 2.2GB database, the `/tmp` copy alone takes ~10-15 seconds, leaving insufficient time for the actual decode.

This PR:
- **Raises the default timeout from 30s to 5 minutes** (300,000ms), which should handle most databases without any configuration
- **Makes the timeout configurable** via `DECODE_TIMEOUT_MS` environment variable and `--timeout <ms>` CLI flag, so users with very large databases can tune it further
- **Documents the configuration** in the README (Configuration section + Troubleshooting section), including an example Claude Desktop config with `--max-old-space-size=4096` for databases over 1GB

### Files changed
- `src/core/decoder.ts` — new `getDecodeTimeoutMs()` helper, default parameter reads from env
- `src/cli.ts` — `--timeout <ms>` flag with validation, sets `DECODE_TIMEOUT_MS` env var
- `README.md` — "Decode Timeout" config section + "Decode Worker Timed Out" troubleshooting section

## Test plan
- [x] `bun test` — 788 tests pass, 0 failures
- [x] Verified `--help` output shows new `--timeout` flag and `DECODE_TIMEOUT_MS` env var
- [x] Tested against a real 2.2GB database with `--timeout 600000` — server starts and responds to MCP queries successfully
- [x] Confirmed backward compatibility: omitting `--timeout` uses the new 5-minute default

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,6 +331,39 @@ COPILOT_CACHE_TTL_MINUTES=0 copilot-money-mcp
 
 You can also manually refresh the cache using the `refresh_database` tool.
 
+### Decode Timeout
+
+For large databases (500MB+), you may need to increase the decode timeout. The default is 5 minutes (300,000ms).
+
+**Via environment variable:**
+```bash
+DECODE_TIMEOUT_MS=600000 copilot-money-mcp
+```
+
+**Via CLI flag:**
+```bash
+copilot-money-mcp --timeout 600000
+```
+
+**In Claude Desktop config** (with increased Node.js memory for 1GB+ databases):
+```json
+{
+  "mcpServers": {
+    "copilot-money": {
+      "command": "node",
+      "args": [
+        "--max-old-space-size=4096",
+        "/path/to/copilot-money-mcp/dist/cli.js",
+        "--db-path",
+        "/path/to/your/database",
+        "--timeout",
+        "600000"
+      ]
+    }
+  }
+}
+```
+
 ## Known Limitations
 
 ### Local Cache Size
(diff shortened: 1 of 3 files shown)


ecmulli:
> Hey @ecmulli ,thanks for the PR. It's always motivating when people contribute to open source projects. There's a few minor non blockers on your code, but the one I need fixing before merging is that CI is failing with a typecheck error:
> 
> ```
> src/cli.ts(24,27): error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
>   Type 'undefined' is not assignable to type 'string'.
> ```
> 
> `args[i + 1]` is guarded by the `i + 1 < args.length` check, but TypeScript can't narrow array indexing. Fix:
> 
> ```diff
> -      const ms = parseInt(args[i + 1], 10);
> +      const rawValue = args[i + 1] as string;
> +      const ms = parseInt(rawValue, 10);
> ```
> 
> Also needs a Prettier reformat on the `console.error` line (long line). Running `bun run fix` will handle that.
> 
> Update the PR and I'll merge!

Apologies! I just saw this. I saw that you merged though. thanks!

ignaciohermosillacornejo:
No problem! thanks for the contribution. It's my intention to continue
investing in the project, so things like this are really nice.

Issue #260, opened 2026-04-14
.mcpb bundle follow-ups: CI refactor, size, and license hygiene


ignaciohermosillacornejo:
Tracks follow-up work surfaced during PR #251 and the subsequent research into MCPB native-module distribution.

## Context

PR #251 fixed a critical regression: the .mcpb bundle shipped broken from v1.6.0 onwards because \`classic-level\` was externalized at build time but \`node_modules/\` was excluded from the archive. The fix stages production deps into \`.mcpb-staging/\` via \`npm install --omit=dev --ignore-scripts\` and packs from there, producing a 6.9 MB self-contained bundle. A regression test extracts the packed bundle in isolation and verifies \`node dist/cli.js\` responds to \`initialize\` — reproducing Claude Desktop's runtime conditions exactly.

Deep research confirmed this is the correct strategy:
- MCPB has no install-at-extract hook. Claude Desktop never runs \`npm install\` on bundles.
- No pure-JS or WASM LevelDB reader handles Copilot's on-disk format.
- \`@anthropic-ai/mcpb\` manifest v0.3 has no \`dependencies_bundled\`, \`install_on_extract\`, or lifecycle fields.
- \`mcp-alphabanana\` (sharp) is the only known public .mcpb with native deps and takes the same bundling approach.

The items below didn't block the fix but are worth doing separately.

### CI simplification
- [ ] Refactor \`build-mcpb.yml\` to expose a \`workflow_call\` trigger, then replace the inline \`setup-bun + install + pack + gh release create ... bundle\` steps in \`auto-release.yml\` with a \`uses: ./.github/workflows/build-mcpb.yml\` call. Mirrors the \`npm-publish.yml\` pattern that was introduced to solve the GITHUB_TOKEN/\`push: tags: v*\` gap.

### Bundle size
- [ ] **Platform-specific \`.mcpb\` variants**: publish \`copilot-money-mcp-darwin-arm64.mcpb\`, \`-darwin-x64.mcpb\`, \`-linux-x64.mcpb\`, \`-win32-x64.mcpb\`. Strip non-target entries from \`node_modules/classic-level/prebuilds/\` during pack. Research estimates ~75% reduction on the native-binary portion. Uses manifest's \`platform_overrides\` / \`compatibility\`.
[...]

PR #267, opened 2026-04-14
ci: delegate .mcpb build/release to reusable workflow (#260)

ignaciohermosillacornejo:
## Summary

- Shrinks `auto-release.yml` from ~125 lines to 97 by moving the inline bun setup / install / pack / changelog / `gh release create` steps into `build-mcpb.yml` behind a `workflow_call` interface.
- Mirrors the existing `npm-publish.yml` `workflow_call` pattern.
- Hoists `concurrency` to workflow level in `auto-release.yml` so it serializes the whole run, not just one job.
- Removes the dead `push: tags: v*` trigger from `build-mcpb.yml` (never fires for GITHUB_TOKEN-created tags — same reason `npm-publish` had to be chained via `workflow_call`).
- Drops QA steps (typecheck/lint/format:check/test) from `build-mcpb.yml` — `test.yml` already guards every commit on main via PR CI.

No user-visible behavior change. No new secrets; still uses only `github.token`.

Spec: `docs/superpowers/specs/2026-04-14-ci-workflow-call-refactor-design.md`
Plan: `docs/superpowers/plans/2026-04-14-ci-workflow-call-refactor.md`
Tracks: #260

## Test plan

- [x] `actionlint .github/workflows/*.yml` — no new issues (only pre-existing `audit-reviews.yml:78` info)
- [x] Manual `workflow_dispatch` of `build-mcpb.yml` on this branch with `dry_run=true` → bundle uploaded as artifact (6.8 MB), no release created. Run: https://github.com/ignaciohermosillacornejo/copilot-money-mcp/actions/runs/24420501508
- [ ] Post-merge: next `chore: bump version` commit creates a GitHub release with `copilot-money-mcp.mcpb` attached and publishes to npm

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -6,20 +6,18 @@ on:
     paths:
       - 'package.json'
 
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: false
+
 jobs:
-  release:
-    name: Create release on version bump
+  detect-version:
+    name: Detect version bump
     runs-on: ubuntu-latest
-    # Only the repo owner can trigger releases
     if: github.actor == 'ignaciohermosillacornejo'
-    concurrency:
-      group: release-${{ github.ref }}
-      cancel-in-progress: false
-    permissions:
-      contents: write
     outputs:
       changed: ${{ steps.version.outputs.changed }}
-      version: ${{ steps.version.outputs.current }}
+      current: ${{ steps.version.outputs.current }}
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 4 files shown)


ignaciohermosillacornejo:
Addressed review:

- **#1** fixed in d78cfc6 — spec now shows `outputs.current`.
- **#2** (`bun-version: latest`) — leaving as-is. Matches `npm-publish.yml`'s existing stance and the reviewer flagged it as non-blocking. Happy to pin to a specific version in a follow-up if you'd rather.
- **#3** (`.mcpb` existence check) — the check already exists at `scripts/pack-mcpb.ts:76-79` which runs before any release step, so `pack:mcpb` fails fast with a clear message if the bundle isn't produced. Adding a second guard in the workflow would be redundant.

PR #269, opened 2026-04-15
Add Codex support for repo skills

randex:
## Summary

I noticed that Codex doesn't see the repo skills and tends to create a separate `AGENTS.md` copy from `CLAUDE.md`. This PR adds the symlinks Codex expects so it can discover the existing skills, while keeping `CLAUDE.md` and `skills/` as the source of truth.

## Changes

- add `.claude/skills` symlink pointing to `skills/` - recommended path in Claude Code [docs](https://code.claude.com/docs/en/skills)
- add `.agents/skills` symlink for Codex agent discovery
- add `AGENTS.md` symlink for Codex compatibility

## Testing

- `bun run check`

I couldn't validate the bundled `.mcpb` in Claude Desktop because adding the extension immediately shows a yellow `Server disconnected` popup, even on `main` without these changes. If you can, please test that flow on your end.


ignaciohermosillacornejo (approved):
LGTM, thanks for contributing! BTW the issue with not validating the bundle is due to https://github.com/modelcontextprotocol/mcpb/issues/229 and I added a work arround on #270 in the meantime.

Issue #249, opened 2026-04-14
Unable to attach to server


himb0y:
<img width="1251" height="882" alt="Image" src="https://github.com/user-attachments/assets/2dc79c13-d089-4e0b-9bdc-ef6a3ecdda2b" />

what am I doing wrong

ignaciohermosillacornejo:
Hey @himb0y, I was not including a node package in the bundle. Please download 1.6.1 and try again, the issue should be fixed

himb0y:
Downloaded 1.6.1, issue persists

ignaciohermosillacornejo:
I just retested with the version on Github and it works on my machine. Could you please:

1. Remove the connector, then re-add it. 

<img width="1051" height="200" alt="Image" src="https://github.com/user-attachments/assets/2b046458-45b0-4748-8c90-5578ce3c6905" />

2. If the issue still persists, please add here your log file (ask claude to produce it for you, or there's a link to view it in the app) to help me debug.

himb0y:
[mcp-server-Copilot Money MCP Server.log](https://github.com/user-attachments/files/26715857/mcp-server-Copilot.Money.MCP.Server.log)

himb0y:
Uninstalled, re-installed, issue persists. uploaded log

ignaciohermosillacornejo:
Hey @himb0y  — I've attached a one-off diagnostic build of v1.6.1 that adds an unconditional startup banner to stderr. Nothing else changed; it's still write-enabled and read-only safe but heads up that writes no longer work since the Copilot Money team disabled the workaround I was using for that. 

  Download the attached copilot-money-mcp-issue249-debug.mcpb, remove the existing connector in Claude Desktop, and install this one instead. When it fails,
  grab the log again — the first line should now look like:

  [copilot-money-mcp] startup platform=... arch=... node=v... write=true dbPath=...

  That tells me your OS/architecture and whether the Copilot Money database was found. Whatever crash happens next will also show up on stderr
  (module-not-found, native binding error, etc.), which the current build is silent about.

  If the banner line doesn't appear at all in the new log, that itself is useful info — it means the process is dying before our code even runs (e.g., a
  missing native dep at ESM resolve time).

[copilot-money-mcp-issue249-debug.mcpb.zip](https://github.com/user-attachments/files/26721236/copilot-money-mcp-issue249-debug.mcpb.zip)

himb0y:
[...]

PR #273, opened 2026-04-15
chore: release 1.7.1

ignaciohermosillacornejo:
## Summary
- Bumps to `1.7.1` to ship #270: fix for Claude Desktop 1.2581+ refusing to load the extension due to macOS hardened-runtime library validation rejecting `classic-level`'s ad-hoc-signed prebuild. Users on affected Claude Desktop versions currently see a silent "Server disconnected" status with no diagnostic output (process dies in `dlopen` before our JS runs).
- Includes the accompanying scope change from #270: `compatibility.platforms: ["darwin"]`, since Copilot Money itself only ships for macOS.
- Regenerates `package-lock.json` (version drift: lockfile was still on `1.6.1` — bumped to `1.7.1` in the same commit).

Triggers `.github/workflows/auto-release.yml` on merge (fires on `package.json` change under `main` + version diff). That workflow calls `build-mcpb.yml` via `workflow_call` to publish the `.mcpb` asset.

## Changes
- `package.json` / `package-lock.json` — `1.7.0` → `1.7.1`
- `manifest.json` — `1.7.0` → `1.7.1`
- `CHANGELOG.md` — new `## [1.7.1]` section under Keep-a-Changelog format, linking #249 (user report) and upstream [mcpb#229](https://github.com/modelcontextprotocol/mcpb/issues/229)

## Upstream
- User report: #249
- Root-cause + workaround write-up filed upstream: modelcontextprotocol/mcpb#229

## Test plan
- [x] `bun run check` passes (1509 pass, 0 fail)
- [x] Fresh `.mcpb` from `main` installed cleanly in Claude Desktop 1.2581.0, confirmed startup stderr banner now visible in logs and `tools/list` round-trips
- [ ] Auto-release workflow produces `.mcpb` asset on v1.7.1 tag (watches after merge)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.7.1] - 2026-04-15
+
+### Fixed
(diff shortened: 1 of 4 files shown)


ignaciohermosillacornejo:
Both minor observations resolved without code changes:

1. **CHANGELOG attribution for `platforms: ["darwin"]`** — verified against `git show v1.7.0:manifest.json`: the 1.7.0-tagged manifest has no `platforms` field. `platforms: ["darwin"]` genuinely first ships in 1.7.1, so attributing it there is correct.

2. **`entry_point` vs `command`** — per the [mcpb MANIFEST spec](https://github.com/modelcontextprotocol/mcpb/blob/main/MANIFEST.md), `entry_point` is described as "Path to the main server file" and every spec example sets it alongside a `mcp_config.command`, including when `command` is just `"node"` and `args` already contain the path. So `entry_point` is independent metadata (likely used for catalog/display), not the invocation path. Leaving it as `dist/cli.js`.

PR #280, opened 2026-04-15
fix(#278): read budgets from amounts[current_month], drop tombstones

### LINKED ISSUE #278: set_budget writes take minutes to appear in get_budgets (investigation)
## Context

During 2.0.0 smoke testing of the GraphQL write rewrite (#275), `set_budget` was observed to succeed server-side (HTTP 200, `editCategoryBudget: true`) but the new amount did not appear in `get_budgets()` local-cache reads within 60s of polling with `refresh_database`. Four other write types (tags, categories, transactions, recurrings) sync in ~3s.

## Investigation (so far)

Reviewed in the 2.0.0 release branch:

- **Filter in `getBudgets()` (`src/core/database.ts:995-1011`)** — only applies when `active_only=true`. Default is `false`, so all decoded budgets pass through. Not the cause.
- **Orphan filter in `tools.ts:1872-1876`** — excludes budgets whose `category_id` isn't in user/Plaid category maps. Only a factor if the write targets a category ID our cache doesn't know about.
- **Decoder parse failures in `processBudget` (`src/core/decoder.ts:1124-1173`)** — returns `null` on schema mismatch without logging. Theoretically possible that freshly-written budget docs have a shape the decoder drops, but unverified.

## Hypotheses

1. **Copilot native-app sync cadence** — the macOS app writes the `budgets` Firestore collection to local LevelDB on a slower schedule than other collections (most likely).
2. **Decoder silently drops newly-written budget docs** — `processBudget` catches Zod errors and returns `null`; would need debug logging to rule out.
3. **EditBudget creates docs at an unexpected path** — if `EditBudget` on a category that didn't previously have a budget writes to a new doc ID that doesn't match our decoder's collection matcher, we wouldn't see it.

## Gap found (separate)

`EditBudgetMonthly` writes land in the budget doc's `amounts` map (per-month overrides), but `getBudgets()` does not surface this field — only the all-months default `amount` is returned. Calls to `set_budget` with a `month` arg will never be observable via `get_budgets` regardless of sync timing.

## 2.0.0 shipping behavior

- Tool descriptions for `get_budgets` and `set_budget` now warn that budget writes sync on a slower cadence and that per-month overrides are not surfaced in the read view.
- No code-level fix shipped — reproduction steps needed to confirm the root cause.

## Next steps

- [ ] Reproduce with debug logging on `processBudget` to check hypothesis #2
- [ ] Capture the raw LevelDB doc for a freshly-written budget (via `bun run scripts/decode-coverage.ts` or similar) and diff against an old one
- [ ] If confirmed to be native-app sync cadence, leave the tool-description warnings as the shipped behavior
- [ ] Consider surfacing the `amounts` map in the `get_budgets` view so monthly overrides are observable

ignaciohermosillacornejo:
## Summary

Fixes the "set_budget writes never appear in get_budgets" behavior tracked in #278. Root cause was **not** a sync lag — we were reading the wrong field.

## Investigation

Started from issue #278 with three competing hypotheses (Copilot native-app sync cadence, decoder silently dropping budgets, EditBudget creating docs at an unexpected path) and instrumented the real LevelDB to confirm. Two concrete bugs found:

### 1. We were reading the stale legacy field

Copilot's macOS app stopped writing to the top-level `amount` field on budget docs ~2 years ago. Evidence from a real database:

| Budget | top-level `amount` (what we read) | `amounts["2026-04"]` (current month, actually fresh) |
|---|---|---|
| Home | 0 (stale) | 11316.40 (just synced from web-UI edit) |
| Utilities | 10 (stale) | 118 |
| Rent | 2601 (stale) | 5438 |
| Clothing | 157 (stale) | 156.75 |
| Household | 115.38 (stale) | 0 (cleared) |

Verified end-to-end: set Home = \$5,555.55 in the web UI, watched `003918.log` (Copilot's Firestore WAL) get written within minutes, then dumped the raw doc — the new value landed in `amounts["2026-04"]`, not in `amount`.

The Copilot app itself reads `amounts[current_month]`. That's the source of truth. Our `getBudgets` view now does the same.

### 2. 50/86 budget docs were ghost tombstones

Firestore represents deleted docs as empty-field entries. `processBudget` had no guard for this (unlike `processCategory`, which has `if (!name) return null`), so those empty docs surfaced as `{budget_id: "..."}` rows with no category or amount. 58% of the output was garbage rows, making it very hard to spot actual writes among the noise — which is likely why #278 was reported as "writes never appear" rather than "writes appear under wrong name".

## Changes

- \`src/core/decoder.ts\` — \`processBudget\` returns \`null\` for empty-field docs (mirrors \`processCategory\`).
- \`src/tools/tools.ts\`
  - \`getBudgets\` prefers \`amounts[current_month]\` over the stale top-level \`amount\`.
  - Exposes the full \`amounts\` map in the output so callers can inspect history.
  - \`total_budgeted\` uses the current-month effective amount.
  - Tombstone filter runs before orphan-category filter so counts/totals reflect only real budgets.
  - Tool descriptions: \`get_budgets\` documents the new shape (current-month effective, parent categories show resolved totals including children + rollovers); \`set_budget\` loses the "sync lag" caveat (it was never really a sync lag).

## Tests

- \`processBudget skips empty-field tombstone docs\` — decoder-level regression test (\`tests/core/decoder-coverage.test.ts\`).
- 6 new \`getBudgets\` tests in \`tests/tools/tools.test.ts\`:
  - prefers \`amounts[current_month]\` over stale top-level \`amount\`
  - treats \`amounts[current_month]=0\` as an explicit clear (not fallback)
  - falls back to top-level \`amount\` when \`amounts\` map is missing
  - falls back when \`amounts\` has no entry for current month
  - exposes the raw \`amounts\` map in the output
  - \`total_budgeted\` uses current-month override, not stale top-level
- Plus a tool-level tombstone test documenting the contract.

Full suite: 1380 pass / 0 fail / 21 skip.

## Test plan

- [x] \`bun run check\` green
- [x] Branch based on latest \`origin/main\` (f3409e0)
- [x] Reproduced the real-world bug against the developer's live database; confirmed the fix returns the expected current-month value

Closes #278.

### DIFF
--- a/src/core/decoder.ts
+++ b/src/core/decoder.ts
@@ -1122,6 +1122,10 @@ function processRecurring(fields: Map<string, FirestoreValue>, docId: string): R
  * Internal helper to process a budget document.
  */
 function processBudget(fields: Map<string, FirestoreValue>, docId: string): Budget | null {
+  // Firestore represents deleted docs as empty-field entries. Drop them
+  // rather than surfacing ghost `{budget_id}` rows (issue #278).
+  if (fields.size === 0) return null;
+
   const budgetId = getString(fields, 'budget_id') ?? docId;
 
   const budgetData: Record<string, unknown> = {
--- a/src/tools/tools.ts
+++ b/src/tools/tools.ts
@@ -1854,6 +1854,7 @@ export class CopilotMoneyTools {
       budget_id: string;
       name?: string;
       amount?: number;
+      amounts?: Record<string, number>;
       period?: string;
       category_id?: string;
       category_name?: string;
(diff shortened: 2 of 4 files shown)


ignaciohermosillacornejo:
Addressed the two valid nits in 5f5d422:

- **Stale comment (#3)** → rewritten to describe the filter as defense-in-depth now that `processBudget` drops true empty-field tombstones first.
- **Missing `total_budgeted=0` test (#6)** → added. Verifies an explicit `amounts[current_month] = 0` override produces `total_budgeted: 0` even when the stale top-level `amount` is non-zero.

Dismissed two other suggestions as non-issues after verification:

- **UTC vs. local month (#4)** — `Date.prototype.getFullYear()` and `getMonth()` return **local** time per MDN, not UTC. The code is already correct. (UTC-returning variants are `getUTCFullYear()` / `getUTCMonth()`, which I deliberately didn't use.)
- **`amounts` field ordering (#5)** — the spread `...(b.amounts ? { amounts } : {})` is positioned between `amount` and `period` in the object literal (tools.ts:1925-1927), which matches the type annotation order (tools.ts:1856-1858). No inconsistency.

PR #284, opened 2026-04-15
feat(build): local pack:mcpb:write for writes-enabled bundle

ignaciohermosillacornejo:
## Summary

Adds `bun run pack:mcpb:write`, a local-only build that produces `copilot-money-mcp-write.mcpb` — a writes-enabled variant advertising all 30 tools (17 read + 13 write) and passing `--write` to the CLI. Intended for self-installing the writes-enabled server in Claude Desktop alongside the published read-only bundle.

## What it does

- New script `scripts/build-write-manifest.ts` — pure transform that takes the read-only manifest and returns a writes-enabled variant (adds the 13 write tools, appends `--write` to `mcp_config.args`, renames the bundle so it doesn't collide with the read-only install).
- `scripts/sync-manifest.ts --write` writes the variant to a gitignored `manifest.write.json`. Default `bun run sync-manifest` behavior unchanged — still regenerates the read-only `manifest.json`.
- `scripts/pack-mcpb.ts --write` swaps `manifest.write.json` into staging as `manifest.json` and outputs `copilot-money-mcp-write.mcpb`. Default `bun run pack:mcpb` behavior unchanged.
- `pack:mcpb:write` npm script chains both with one command.
- TDD: 8 unit tests cover the transform (tool count, --write arg, no input mutation, custom-description preservation, etc.).

## Why local-only

The published Claude Desktop bundle is intentionally read-only. The writes-enabled CLI requires explicit user opt-in and direct authentication to Copilot Money's GraphQL backend; we don't want it discoverable through the catalog. The release workflow is unchanged and continues to ship only the read-only bundle.

## How to use

```bash
bun run pack:mcpb:write   # produces copilot-money-mcp-write.mcpb
```

Then double-click the `.mcpb` to install into Claude Desktop.

## Constraints honored

- Committed `manifest.json` is never modified — `git status` after a writes build is clean (artifacts are gitignored).
- No CI / release-workflow changes.
- No README mention; the only documentation lives in CONTRIBUTING.md.

## Context

Sibling to the recently-merged #281 (read-only bundle hardening / docs).

## Test plan

- [x] `bun run check` — typecheck, lint, format, 1389 tests pass
- [x] `bun run pack:mcpb:write` — produces a valid bundle; `unzip -p copilot-money-mcp-write.mcpb manifest.json` shows 30 tools, `--write` arg, distinct name (`copilot-money-mcp-write`)
- [x] `bun run sync-manifest` (no flag) — read-only manifest.json unchanged
- [x] `git status` after the writes build — only the (gitignored) artifacts exist; committed manifest untouched


ignaciohermosillacornejo:
Thanks for the review. Addressed in c056bde:

- **#1 hardcoded `17 read`** — fixed. Computed from `clone.tools.length - writeSchemas.length` and added a regression test that the description reports an accurate read/write split.
- **#2 `as never` cast** — fixed. Extracted the `Manifest`/`ManifestTool` types into a new `scripts/manifest-utils.ts` and imported them in both scripts. The cast is gone; `buildWriteManifest(manifest)` now type-checks directly.
- **#3 duplicated `truncateDescription`** — fixed. Same `manifest-utils.ts` module owns the helper; both scripts import it.
- **#4 absent/null `args` edge case** — intentionally not guarded. The MCP manifest spec requires `mcp_config.args` to be a string array (the read-only `manifest.json` we read from has had it since v0.1, and `bunx @anthropic-ai/mcpb` would reject a manifest without it). Adding a defensive check for an impossible state is anti-pattern per `CLAUDE.md` ("don't add error handling, fallbacks, or validation for scenarios that can't happen"). Happy to revisit if you'd rather have a guard.

Also rebased onto current main (17f1695) so the merge is clean. Setting auto-merge.

PR #302, opened 2026-04-18
fix(account): allow null vested_quantity and vested_value in holdings

har1kaush1k:
## Summary

`AccountHoldingSchema` declares `vested_quantity` and `vested_value` as `z.number().optional()`, which rejects `null`. Firestore stores these fields as literal `null` on investment accounts that aren't stock-plan accounts (plain brokerage, IRA, Roth, cash-management, etc.). Zod throws on parse, `processAccount`'s `catch` silently returns `null`, and those accounts are dropped entirely from `get_accounts` output.

Switched both fields to `.nullable().optional()` to match the existing `cost_basis` pattern in the same schema.

## Repro

Build a holding with `vested_quantity: null` (what Copilot actually writes for non-stock-plan accounts):

```ts
AccountHoldingSchema.parse({
  security_id: 'x',
  quantity: 1,
  institution_price: 1,
  institution_value: 1,
  cost_basis: null,
  vested_quantity: null,   // <-- throws before this patch
  vested_value: null,      // <-- throws before this patch
});
```

Before: `ZodError: Invalid input: expected number, received null` at `holdings[*].vested_quantity` / `vested_value`, account silently omitted from `get_accounts`.
After: parses, account appears in results.

## Changes

- `src/models/account.ts`: `vested_quantity` and `vested_value` → `.nullable().optional()`
- `tests/models/account.test.ts`: regression tests covering null, numeric, and missing cases

## Test plan

- [x] `bun test tests/models/account.test.ts` — 3/3 pass
- [x] `bun run check` (typecheck + lint + format + full suite) — 0 failures

### DIFF
--- a/src/models/account.ts
+++ b/src/models/account.ts
@@ -21,8 +21,8 @@ const AccountHoldingSchema = z
     institution_value: z.number().optional(),
     quantity: z.number().optional(),
     iso_currency_code: z.string().optional(),
-    vested_quantity: z.number().optional(),
-    vested_value: z.number().optional(),
+    vested_quantity: z.number().nullable().optional(),
+    vested_value: z.number().nullable().optional(),
   })
   .passthrough();
 
--- a/tests/models/account.test.ts
+++ b/tests/models/account.test.ts
@@ -1,0 +1,64 @@
+/**
+ * Schema tests for AccountSchema, focused on holdings field nullability.
+ *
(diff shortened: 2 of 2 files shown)


ignaciohermosillacornejo (approved):
LGTM, was able to reproduce and fix works, test is excellent. Thanks for contributing!

Will follow up with some more comprehensive approach to surface this kinds of silently failing bugs, probably some logging about dropped transactions by zod to catch this bugs and any drift from firestore

PR #321, opened 2026-04-22
feat(tools): add delete_transaction write tool

ignaciohermosillacornejo:
## Summary

Second of four planned write-tools based on the GraphQL recon. Wraps `deleteTransaction` — confirmed to exist in the schema but not in the web-session capture. Destructive: there is no soft-delete, no undo.

Requires all three IDs (`transaction_id`, `account_id`, `item_id`) — no cache lookup fallback — so a typo in any single field fails with "Transaction not found" rather than silently deleting a different transaction.

## Recon notes

Signature confirmed against the live endpoint:
- No-arg probe: lists exactly `itemId: ID!, accountId: ID!, id: ID!` as required — no optional args exist.
- Fake-ID probe: returns "Transaction not found" (NOT_FOUND) — confirms data-layer validation.
- Return type: `Boolean!` scalar — no output type, no subfield selection.

## Changes

- `docs/graphql-capture/operations/mutations/DeleteTransaction.md` — capture markdown
- `scripts/generate-graphql-operations.ts` — added `'DeleteTransaction'` to `IN_SCOPE_MUTATIONS`
- `src/core/graphql/operations.generated.ts` — regenerated (15 ops now)
- `src/core/graphql/transactions.ts` — `deleteTransaction()` wrapper + `DeleteTransactionArgs` type
- `src/core/database.ts` — `patchCachedTransactionDelete(transactionId)` helper mirroring `patchCachedTagDelete` / `patchCachedCategoryDelete` / `patchCachedRecurringDelete`
- `src/tools/tools.ts` — `CopilotMoneyTools.deleteTransaction()` method + schema (`destructiveHint: true, idempotentHint: true`)
- `src/server.ts` — switch case + `WRITE_TOOLS` registration (tool count 14 → 15)
- Tests: 4 wrapper tests + 10 tool tests (dispatch, bool mirror, cache evict on success only, no-lookup contract, per-field validation, error wrapping)
- Three lockstep updates: `server-protocol.test.ts`, `server-write-dispatch.test.ts`, `write-tool-schemas.test.ts`

## Design choice: require all three IDs verbatim

`updateTransaction` resolves `account_id` / `item_id` from the local cache via `transaction_id`. For `delete_transaction` we explicitly **do not** — a dedicated test passes a cached `transaction_id` alongside intentionally-wrong `account_id` / `item_id` and asserts the wrong values go through to the GraphQL wire unchanged. This makes a typo fail loudly at the server boundary rather than silently deleting a different transaction the cache happens to have.

## Cache eviction

On a `true` return from the server, the transaction is removed from the in-memory cache via the new `patchCachedTransactionDelete` helper. On `false` or error, cache is intentionally untouched so local state does not drift.

## Test plan

- [x] `bun run check` passes (typecheck + lint + format + 1488 tests)
- [x] Wrapper test: `mutate('DeleteTransaction', ...)` with correct variables, Boolean return
- [x] Tool tests: happy path, server returns false (cache not evicted), invalid doc IDs (all three paths), GraphQL error wrapping, no-cache-lookup contract test
- [ ] Smoke test against real endpoint (deferred — would actually delete)


ignaciohermosillacornejo:
Addressed review item 4 in f890514 — `DeleteTransaction.md` "Fires on" line now makes the synthetic/not-captured status explicit up front, with the intended-use narrative moved inside the parenthetical.

Items 1 (`idempotentHint` convention), 2 (literal `success: true` type), 3 (SAFE_ID_RE minimum length), and 5 (`DeleteTransactionArgs` exported for consumers) left as-is — all flagged as consistent with existing project conventions in the review itself.

PR #322, opened 2026-04-22
feat(tools): add add_transaction_to_recurring write tool

ignaciohermosillacornejo:
## Summary

Third of four planned write-tools. Wraps `addTransactionToRecurring` — the mutation Copilot uses when a user manually links a one-off transaction to an existing recurring series that auto-detection missed.

## Recon — filled output-shape gap

The previous recon (hidden-mutations.md) listed `AddTransactionToRecurringOutput` fields and optional input fields as "unknown". Closed both gaps via live-endpoint probes today:

- **Output**: exactly one field, `transaction: Transaction!`. Probed candidates rejected: `recurring`, `updated`, `id`, `success`, `errors`, `node`, `data`, `recurringTransaction` (the last returned "Did you mean 'transaction'" confirming the single-field shape).
- **Input**: `recurringId: ID!` is the only accepted field. Probed optional candidates rejected: `date`, `isReviewed`, `notes`, `tagIds` — server returned "not defined by type AddTransactionToRecurringInput" for each.

`hidden-mutations.md` updated in the first commit of this PR.

## Cache strategy

Unlike `create_transaction` (no pre-existing cache row to patch), this tool patches `recurring_id` onto the existing cached transaction via `patchCachedTransaction` — the same helper `update_transaction` uses. A subsequent `get_transactions` call will see the link without needing `refresh_database`.

## Changes

- `docs/graphql-capture/hidden-mutations.md` — output/input shape now documented concretely
- `docs/graphql-capture/operations/mutations/AddTransactionToRecurring.md` — synthetic capture markdown (uses the "(synthetic — not captured from real traffic...)" convention from PR #321)
- `scripts/generate-graphql-operations.ts` — added `'AddTransactionToRecurring'` to `IN_SCOPE_MUTATIONS`
- `src/core/graphql/operations.generated.ts` — regenerated (16 ops now)
- `src/core/graphql/transactions.ts` — `addTransactionToRecurring()` wrapper + types; wrapper unwraps the `{ transaction }` output level
- `src/tools/tools.ts` — `CopilotMoneyTools.addTransactionToRecurring()` + schema
- `src/server.ts` — switch case + `WRITE_TOOLS` registration (count 15 → 16)
- Tests: 3 wrapper tests + 8 tool tests (happy path, cache patch on recurring_id, verbatim ID forwarding, all four invalid-ID paths, GraphQL error wrapping, INTERNAL_TRANSFER mapping)
- Three lockstep test updates

## Test plan

[...]

PR #320, opened 2026-04-22
feat(tools): add create_transaction write tool

ignaciohermosillacornejo:
## Summary

Adds `create_transaction` as the first of four new MCP write tools wrapping Copilot's recently-discovered GraphQL mutations (see [hidden-mutations.md](docs/graphql-capture/hidden-mutations.md)).

Creates a brand-new manual transaction on an existing account. Seven required inputs: `account_id`, `item_id`, `name`, `date`, `amount`, `category_id`, `type`. The `type` field is constrained to the three enum values the live endpoint accepts.

## TransactionType enum discovery

The `TransactionType!` enum required by `createTransaction` wasn't documented in the recon. Verified against the live endpoint today:

- Probing with `INVALID_PROBE_VALUE` → server rejected without listing valid values.
- Probing `REGULAR`, `INCOME`, `INTERNAL_TRANSFER` → each passed enum validation and reached the data layer (failed on fake ID as expected).
- Probing typos (`REGULR`, `INCOM`, `INTERNA_TRANSFER`, `EXTERNAL_TRANSFER`) → "Did you mean REGULAR / INCOME / INTERNAL_TRANSFER" — no other values appeared across broad sweeps (`EXPENSE`, `TRANSFER`, `DEPOSIT`, `WITHDRAWAL`, etc. all rejected).

Conclusion: enum is exactly `REGULAR | INCOME | INTERNAL_TRANSFER`. Encoded as a TS union and as a JSON Schema enum so callers can't send garbage.

## Changes

- `docs/graphql-capture/operations/mutations/CreateTransaction.md` — mutation capture markdown (template-followed from EditTransaction.md)
- `scripts/generate-graphql-operations.ts` — added `CreateTransaction` to `IN_SCOPE_MUTATIONS`
- `src/core/graphql/operations.generated.ts` — regenerated (adds `CREATE_TRANSACTION`)
- `src/core/graphql/transactions.ts` — `createTransaction()` wrapper + `TransactionType` union
- `src/tools/tools.ts` — `CopilotMoneyTools.createTransaction()` method + schema
- `src/server.ts` — switch case + `WRITE_TOOLS` registration
- Tests: wrapper unit test, tool unit tests (happy + validation + error cases), dispatch/protocol lockstep updates

## Test plan

- [x] `bun run check` passes (typecheck + lint + format + 1467 tests)
- [x] Wrapper test: `mutate('CreateTransaction', ...)` called with mapped variables
[...]

PR #323, opened 2026-04-22
feat(tools): add split_transaction write tool

ignaciohermosillacornejo:
## Summary

Fourth and final write-tool from the GraphQL recon. Wraps `splitTransaction` — splits one parent transaction into N category-specific children. Exposes the mutation Copilot's desktop UI uses for "split transaction."

Input: `{ transaction_id, account_id, item_id, splits: [{ name?, date?, amount, category_id }] }`. Per-split `name` and `date` default to the parent's values if omitted. Server enforces that children amounts sum to the parent's amount; the tool also validates this client-side (1e-6 epsilon for IEEE-754 drift) so the caller gets a clear local error before a round-trip.

## Recon findings (fully confirmed against live endpoint)

- Signature: `splitTransaction(itemId: ID!, accountId: ID!, id: ID!, input: [SplitTransactionInput!]!)` — confirmed by no-arg probe.
- `SplitTransactionInput` required fields: `name: String!, date: Date!, amount: Float!, categoryId: ID!` — confirmed by empty-input probe. No optional fields (tags, notes, isReviewed all rejected as "not defined").
- Output: `SplitTransactionOutput!` with exactly two fields: `parentTransaction: Transaction!` and `splitTransactions: [Transaction!]!` — every other candidate subfield returned "Cannot query field".

## Cache strategy

On success the parent is evicted via `patchCachedTransactionDelete` (added in PR #321) — Copilot's UI hides the parent post-split, so cached reads should stop surfacing it. Children are NOT inserted into the cache; there's no insert helper and the next `refresh_database` picks them up. Callers needing children immediately can read them from the returned `children` array.

## Reversal

There is none — `unsplitTransaction`, `revertSplit`, `undoSplit` all probed and don't exist. Documented in the tool description and method JSDoc: to undo, delete each child via `delete_transaction` and edit the parent's category back via `update_transaction`.

## Changes

- `docs/graphql-capture/operations/mutations/SplitTransaction.md` — capture markdown
- `scripts/generate-graphql-operations.ts` — added `'SplitTransaction'` to `IN_SCOPE_MUTATIONS`
- `src/core/graphql/operations.generated.ts` — regenerated (17 ops now)
- `src/core/graphql/transactions.ts` — `splitTransaction()` wrapper + types
- `src/tools/tools.ts` — `CopilotMoneyTools.splitTransaction()` + schema
- `src/server.ts` — switch + `WRITE_TOOLS` registration (count 16 → 17)
- Tests: 3 wrapper tests + 18 tool tests (happy path, 3-way split, name/date defaulting, sum mismatch, epsilon tolerance, every validation axis including all ID paths and per-split fields, GraphQL error path with cache-not-evicted assertion, verbatim ID forwarding)
- Four lockstep updates (count 16→17): `server-protocol.test.ts`, `server-write-dispatch.test.ts`, `write-tool-schemas.test.ts`, `build-write-manifest.test.ts`
- Tool-count copy sync: `CLAUDE.md`, `CONTRIBUTING.md`, `package.json` (30 read+write → 31)

## Test plan

- [x] `bun run check` passes (typecheck + lint + format + 1529 tests)
- [x] Sum check with float epsilon (proof: `0.1 + 0.2` → parent `0.3` passes; mismatch by > 1e-6 rejects)
- [x] Defaulting from parent: no-name split inherits parent name; no-date split inherits parent date
- [x] Original-name fallback when parent has `original_name` but no `name`
- [x] Cache eviction on success only; GraphQL error leaves cache alone
- [x] All four ID validation paths exercised (transaction, account, item, each split's category)
- [ ] Smoke test against real endpoint (deferred — would actually split a transaction)

## Closes the write-tools-from-recon series

This is the last of the four PRs (#320 create, #321 delete, #322 add_transaction_to_recurring, this one). All hidden mutations documented in `docs/graphql-capture/hidden-mutations.md` that were safe to wrap are now exposed as MCP write tools. The remaining undocumented mutations (`bulkEditTransactions`, `bulkDeleteTransactions`, `createAccount`, `deleteAccount`, `deleteUser`) are explicitly out-of-scope per the recon doc — bulk ones because of the SQL-execution risk on probe, account/user ones because they aren't wanted.


ignaciohermosillacornejo:
Addressed review items 1, 2, 5 in 6c01483:

- **Item 1** (destructiveHint): flipped to `true` — reviewer's argument was compelling (parent permanently hidden post-split, no reversal mutation, non-idempotent). Updated the lockstep annotation table (`server-protocol.test.ts`) and schema assertion (`write-tool-schemas.test.ts`) to match.
- **Item 2** (type annotation): replaced `(typeof result.splitTransactions)[number]` with the imported `CreatedTransaction` alias in `toLocal`. Same type, clearer read.
- **Item 5**: added a comment on the `parentDefaultName!` non-null assertion explaining why the earlier guard makes it safe (TS can't narrow across the `.some` call).

Items 3 (mixed-sign splits) and 4 (date regex syntax-only) left as-is per reviewer's own triage — item 3 is intentional flexibility for refund-in-expense scenarios, item 4 is consistent with every other date-accepting tool in the codebase.

Issue #326, opened 2026-04-22
decoder: tombstoned (deleted) Firestore docs still returned after NoDocument sync


ignaciohermosillacornejo:
## Summary

`src/core/decoder.ts` returns deleted transactions (and presumably all other deleted Firestore docs) even after Copilot's desktop app has synced the delete to the local LevelDB cache. Deletes on Copilot's server succeed, the app sees and writes the tombstone to LevelDB, but our decoder ignores the tombstone and returns the doc as present.

Discovered 2026-04-22 while smoke-testing the four new write tools (PRs #320, #321, #322, #323) against a real Copilot account.

## Reproduction

1. Install + run Copilot desktop app (authoritative source for LevelDB writes).
2. From a script (real Firebase token + GraphQL client):

```ts
// pseudo-code
const tools = new CopilotMoneyTools(new CopilotDatabase(), client);

// Create
const created = await tools.createTransaction({
  account_id, item_id, name: 'GQL-TEST-unique',
  date: '2026-04-22', amount: 1.11,
  category_id, type: 'REGULAR',
});
console.log('id =', created.transaction_id);

// Sanity-check fresh decoder read sees the new row (this works — propagates in ~1ms)
const freshDb1 = new CopilotDatabase();
const present = await freshDb1.getAllTransactions();
assert(present.find((t) => t.transaction_id === created.transaction_id));

// Delete
await tools.deleteTransaction({
  transaction_id: created.transaction_id, account_id, item_id,
});

// Server confirms delete via idempotent re-call
const r = await deleteTransaction(client, {
  id: created.transaction_id, accountId: account_id, itemId: item_id,
});
assert(r === true);  // server says "yes, already deleted"

// Wait 30+ seconds, re-decode LevelDB fresh, expect row gone
await sleep(30_000);
const freshDb2 = new CopilotDatabase();
const stillThere = (await freshDb2.getAllTransactions())
  .find((t) => t.transaction_id === created.transaction_id);

// ACTUAL: stillThere is the full transaction object — name/amount/date match the original create
// EXPECTED: stillThere === undefined
```

### Actual run evidence

```
[1] mtime baseline…
    2026-04-22T17:45:27.405Z

[2] create_transaction "GQL-TEST-disk-proof-1776879927572"
    server returned id=YF943XqMCKtguESuk4cE

[3] wait for LevelDB .log mtime to advance…
    ✅ mtime changed after 1ms

[...]

PR #327, opened 2026-04-22
feat(smoke): add transactions-write section covering PR #320-#323 tools

ignaciohermosillacornejo:
## Summary

Extends `scripts/smoke-graphql.ts` with a new `transactions-write` section that exercises the four write tools shipped in PRs #320-#323 end-to-end against the real Copilot endpoint: `create_transaction`, `delete_transaction`, `add_transaction_to_recurring`, `split_transaction`.

Previously these four were only unit-tested with mocked GraphQL; this section closes the gap.

## What the section covers (7 steps, all pass on live run)

| Step | Latency |
|---|---|
| `create_transaction REGULAR + delete` | 769ms |
| `create_transaction INCOME + delete` | 587ms |
| `create_transaction INTERNAL_TRANSFER + delete` | 584ms |
| `split_transaction 2-way with name/date defaulting` | 4233ms |
| `split_transaction 3-way even split` | 4572ms |
| `split_transaction client-side sum mismatch rejected` | 3818ms |
| `add_transaction_to_recurring links a new txn to an active recurring` | 862ms |

**Run locally:**
```
bun run scripts/smoke-graphql.ts --section transactions-write
```

**Output from the author's live run:**
```
Section               Pass  Fail      Time
------------------------------------------
transactions-write       7     0   15.43s
Latency (successful ops): median=862ms max=4572ms
✓ no cleanup stragglers
```

## Design choices

**Manual account only.** Plaid-connected accounts are unsafe (Copilot re-syncs from the source and would overwrite our test writes). Section skips with a warning if no open manual account exists.

**Prerequisite gates** — section warns and exits cleanly if:
- No open manual account
- Fewer than 2 user-defined categories (split needs two)

**Add-to-recurring is skippable** (uses `stepSkippable`) — runs only if an active recurring exists, doesn't block the rest of the section.

**Cache-lag handling for split.** `tools.splitTransaction` reads the parent from the decoded LevelDB cache for name/date defaulting + client-side sum check. A just-created parent isn't visible until Copilot's desktop app syncs. Added a local `waitForTxnVisible(txnId)` helper (60s poll @ 2s intervals, using `pollUntil` + `tools.refreshDatabase()`) before every split call. Throws a clear "did not appear within 60s — Copilot desktop app sync lag" error if the app isn't running.

**Cleanup is ironclad.** Per-section cleanup registry of `{ id, account_id, item_id, label }`. Every created row is pushed before any state-producing mutation, so a mid-run throw still sweeps via try/finally. Cleanup uses raw `deleteTransaction` wrapper (bypasses MCP-tool validation). Server errors during cleanup log manual-cleanup instructions with a searchable `GQL-TEST-*` label.

**Decoder caveat documented inline.** The section docstring notes issue #326 (decoder doesn't respect Firestore NoDocument tombstones) and explicitly states it verifies **server-side state only** — doesn't assert deleted transactions disappear from `db.getAllTransactions()`. Once #326 is fixed, this section can be extended to also assert cache eviction.

**Gated on `!skipDestructive`.** The section mutates real data, so `--skip-destructive` correctly opts out.

## Implementation

One commit (`34e9775`), +505/-3 in `scripts/smoke-graphql.ts`, pure-additive:
- Added `'transactions-write'` to `VALID_SECTIONS`.
- Imported the four new GraphQL wrappers alongside the existing `editTransaction`.
- Added `smokeTransactionsWrites(client, db, tools)` function.
- Wired dispatch in `main()` gated on `!skipDestructive`.
- Updated header Usage block + Sections line.

## Test plan

- [x] `bun run check` passes (typecheck + lint + format + 1529 tests, 0 failures)
- [x] Live run `bun run scripts/smoke-graphql.ts --section transactions-write` — 7/7 pass, 15.43s, no stragglers
- [x] Live run proves the new write tools round-trip correctly for REGULAR / INCOME / INTERNAL_TRANSFER types on create, 2-way + 3-way splits with name/date defaulting, client-side sum-check rejection, and link-to-recurring

## Related

- Shipped tools: #320 create_transaction, #321 delete_transaction, #322 add_transaction_to_recurring, #323 split_transaction
- Known issue exposed during smoke test: #326 (decoder doesn't respect Firestore NoDocument tombstones). Referenced from the section docstring so the follow-up cache-assertion work is discoverable once that's fixed.

### DIFF
--- a/scripts/smoke-graphql.ts
+++ b/scripts/smoke-graphql.ts
@@ -16,9 +16,11 @@
  *   bun run scripts/smoke-graphql.ts --skip-mcp-e2e      # skip MCP tool dispatch
  *   bun run scripts/smoke-graphql.ts --include-state-gated # opt-IN (interactive)
  *   bun run scripts/smoke-graphql.ts --section tags
+ *   bun run scripts/smoke-graphql.ts --section transactions-write
  *
- * Sections: tags, categories, transactions, recurrings, budgets, accounts,
- *           bulk, errors, edge, consistency, mcp-e2e, state-gated
+ * Sections: tags, categories, transactions, transactions-write, recurrings,
+ *           budgets, accounts, bulk, errors, edge, consistency, mcp-e2e,
+ *           state-gated
  *
  * State-gated tests are opt-in because they require the user to toggle UI
  * state manually in the Copilot desktop app between prompts.
@@ -42,7 +44,13 @@ import {
   editCategory,
   deleteCategory,
 } from '../src/core/graphql/categories.js';
-import { editTransaction } from '../src/core/graphql/transactions.js';
+import {
+  addTransactionToRecurring,
+  createTransaction,
+  deleteTransaction,
+  editTransaction,
+  splitTransaction,
+} from '../src/core/graphql/transactions.js';
 import {
   createRecurring,
   editRecurring,
@@ -117,6 +125,7 @@ const VALID_SECTIONS = [
   'tags',
   'categories',
   'transactions',
+  'transactions-write',
   'recurrings',
   'budgets',
   'accounts',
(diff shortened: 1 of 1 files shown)


ignaciohermosillacornejo:
Addressed review feedback in 5300db2:

- **Bug fix**: replaced all 6 `cleanup.splice(cleanup.findIndex(...), 1)` call sites with a guarded `unregisterCleanup(id)` helper that checks `if (idx !== -1)` — no more silently removing the last element on a repeat/reordered delete.
- **Nit**: `CleanupEntry` lifted to module scope.
- **Nit**: `waitForTxnVisible` now logs observed propagation lag (uses the `observedLagMs` that `pollUntil` was already measuring). Handy diagnostic for CI / slow-sync environments.

Re-ran the section against live data — 7/7 pass, no cleanup stragglers, lag lines visible (~2.9-3.1s in this environment). `bun run check` green.

Left the other review observations (`if (quick) return` guard, straggler message including accountId/itemId) as-is per the reviewer's own triage ("not a blocker" / "low priority").

PR #405, opened 2026-05-27
feat(tools): fall back to live GraphQL for write-tool transaction resolution

aberhamm:
## Summary

When `--live-reads` is enabled, write tools (`update_transaction`, `split_transaction`, `review_transactions`, `create_recurring`) now fall back to the live GraphQL API when a transaction isn't found in the local LevelDB cache. Previously these tools threw "Transaction not found" for any transaction outside the ~30-day local cache window, even though the GraphQL API could update them.

Real-world scenario: tagging historical transactions (e.g., a vacation from 3 months ago) that `get_transactions_live` can read but `update_transaction` couldn't write to.

## Changes

- Added `resolveTransaction()` private helper with three-tier lookup: local cache → live window cache → GraphQL fetch
- Added `resolveTransactions()` batch variant for `reviewTransactions`
- Added `transactionNodeToLocal()` mapper from GraphQL shape to local Transaction type
- `CopilotMoneyTools` constructor accepts an optional `liveDb` parameter, wired up in `server.ts` when `--live-reads` is active
- No behavior change when `--live-reads` is off — the fallback only activates when `liveDb` is present

### Affected write methods
- `updateTransaction` 
- `splitTransaction`
- `reviewTransactions`
- `createRecurring`

`deleteTransaction` and `addTransactionToRecurring` already require explicit `account_id`/`item_id` parameters, so no cache lookup is needed.

## Test plan

- [x] 8 new tests in `tests/tools/live-fallback-writes.test.ts`
- [x] All 1927 tests pass, 0 failures
- [x] `bun run check` (typecheck + lint + format + tests) exits 0
- [ ] Manual verification: successfully tagged 43 transactions from February 2026 that were outside the local cache window

### DIFF
--- a/src/tools/tools.ts
+++ b/src/tools/tools.ts
@@ -36,6 +36,7 @@ import {
   deleteRecurring as gqlDeleteRecurring,
 } from '../core/graphql/recurrings.js';
 import { setBudget as gqlSetBudget } from '../core/graphql/budgets.js';
+import type { TransactionNode } from '../core/graphql/queries/transactions.js';
 import { graphQLErrorToMcpError } from './errors.js';
 import { parsePeriod } from '../utils/date.js';
 import { computeTotalReturnPercent, roundAmount } from '../utils/round.js';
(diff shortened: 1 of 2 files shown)


aberhamm:
Closing to rework. Will resubmit with cleaner implementation.

PR #406, opened 2026-05-27
feat(tools): resolve transactions via live GraphQL when local cache misses

aberhamm:
## Problem

Write tools (`update_transaction`, `split_transaction`, `review_transactions`, `create_recurring`) resolve `account_id` and `item_id` from the local LevelDB cache before sending GraphQL mutations. Transactions outside the ~30-day cache window throw "Transaction not found" even though the GraphQL API can update them.

Real-world trigger: tagging historical transactions (e.g., a vacation from months ago) that `get_transactions_live` can read but write tools can't reach.

## Solution

When `--live-reads` is active, write tools now fall back to the live transaction window cache (populated by prior `get_transactions_live` calls) before failing. An optional `dateHint` parameter enables a scoped GraphQL fetch for the transaction's month when the window cache also misses.

Three-tier lookup: local LevelDB cache -> live window cache -> GraphQL fetch (month-scoped, only with dateHint).

No behavior change when `--live-reads` is off.

## Changes

- `resolveTransaction(id, dateHint?)` — three-tier lookup with honest error surfacing (GraphQL failures are wrapped, not swallowed)
- `resolveTransactions(ids)` — batch variant for `review_transactions`
- `transactionNodeToLocal(node)` — maps camelCase GraphQL shape to snake_case local Transaction
- Four call sites replaced: `updateTransaction`, `splitTransaction`, `reviewTransactions`, `createRecurring`
- `deleteTransaction` and `addTransactionToRecurring` unchanged (already require explicit IDs)

## Test plan

- [x] 7 new tests in `tests/tools/live-fallback-writes.test.ts`
- [x] 1926 pass, 0 fail, 20 skip
- [x] `bun run check` clean (typecheck, lint, format)
- [x] Manual: tagged 43 transactions from February 2026 that were outside the local cache

### DIFF
--- a/src/tools/tools.ts
+++ b/src/tools/tools.ts
@@ -23,6 +23,7 @@ import {
   deleteCategory as gqlDeleteCategory,
 } from '../core/graphql/categories.js';
 import type { CategoryNode } from '../core/graphql/queries/categories.js';
+import type { TransactionNode } from '../core/graphql/queries/transactions.js';
 import type { TagNode } from '../core/graphql/queries/tags.js';
 import type { RecurringNode } from '../core/graphql/queries/recurrings.js';
 import {
@@ -37,7 +38,7 @@ import {
 } from '../core/graphql/recurrings.js';
 import { setBudget as gqlSetBudget } from '../core/graphql/budgets.js';
 import { graphQLErrorToMcpError } from './errors.js';
-import { parsePeriod } from '../utils/date.js';
+import { parsePeriod, getMonthRange } from '../utils/date.js';
 import { computeTotalReturnPercent, roundAmount } from '../utils/round.js';
 import {
   getCategoryName,
@@ -338,7 +339,7 @@ export class CopilotMoneyTools {
    *
    * @param database - CopilotDatabase instance
    * @param graphqlClient - Optional GraphQL client for write operations.
-   * @param liveDb - Optional LiveCopilotDatabase for write-through to live cache.
+   * @param liveDb - Optional LiveCopilotDatabase for read-fallback and write-through cache patching when --live-reads is active.
    */
   constructor(
     database: CopilotDatabase,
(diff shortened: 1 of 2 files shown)


ignaciohermosillacornejo:
Hey @aberhamm, thanks a lot for this PR and for the clear writeup.

Before reviewing I want to discuss this issue from a product perspective with you, because I think it points at a gap that we have in the MCP server. Right now we have two flags, `--live-reads` and `--write`, and a user can set any or none, so we end up with this combination of scenarios:

| | no `--write` | `--write` |
|---|---|---|
| **no `--live-reads`** | Cache reads, read-only — the default | Cache reads + live writes — **writes being affected by sparse cache** |
| **`--live-reads`** | Live reads, read-only | Live reads + live writes — full surface |

This status is an artifact of how the code developed. We initially had all cache reads and writes directly to Firestore, before the Copilot Money team enabled App Check. I then pivoted to add writes via the GraphQL endpoint, and later extended that to cover the entire GraphQL surface (`--live-reads`), since I was frustrated that the cache is pretty sparse for data older than 30 days.

The only reason writes are limited to the ~30-day window is that they resolve `account_id`/`item_id` from the local cache. But that whole limitation only exists in the top-right quadrant — `--write` without `--live-reads` — and I'm not sure it should exist at all. From a product perspective it doesn't make much sense.

Once a user has opted into writes, they're already consenting to us pulling their Firebase refresh token out of their browser's local storage (the `AMf-…` token, which we exchange for short-lived ID tokens) to make requests to Copilot's backend, and live reads are themselves cached, so there's no real privacy or perf cost left to justify keeping reads on the stale local cache. Do you see any benefit in keeping it?

So my instinct would be to make `--write` imply `--live-reads`, so the matrix collapses to:

| | no `--write` | `--write` |
|---|---|---|
| **no `--live-reads`** | Cache reads, read-only — the default | Live reads + live writes (implied) |
| **`--live-reads`** | Live reads, read-only | Live reads + live writes |

That removes the awkward quadrant entirely and closes the gap you hit for *every* historical transaction, not just recently-read ones — which would make this PR mostly unnecessary.

What do you think?

---

aberhamm:
Hey @ignaciohermosillacornejo, thanks for the thoughtful response — really appreciate you thinking through this at the product level rather than just reviewing the code.

You're right. The `--write` without `--live-reads` quadrant is an artifact, not an intentional design choice, and my PR is essentially patching around that gap rather than closing it. Once a user has opted into writes and the Firebase token extraction, there's no meaningful reason to keep them on stale cache reads. Making `--write` imply `--live-reads` is the cleaner fix — it removes the problem for all historical transactions, not just ones that happen to be in a recently-read window.

Happy to close this and open a smaller PR that wires up the flag implication instead, or if you'd prefer to handle it differently on your end, just let me know.

ignaciohermosillacornejo:
Happy if you take the lead on this.

Ignacio

aberhamm:
Closing in favor of the simpler approach discussed above — making `--write` imply `--live-reads`. Will open a new PR for that.

PR #407, opened 2026-05-30
Make --write imply --live-reads

aberhamm:
## Summary

- `--write` now automatically enables `--live-reads`, collapsing the flag matrix as discussed in #406
- Write tools resolved `account_id`/`item_id` from the local LevelDB cache (~30 days), causing failures on older transactions even though the GraphQL API handles them fine
- Since `--write` already requires the Firebase auth token, there's no privacy or perf reason to keep reads on the stale cache

## Changes

One-line logic change in `src/cli.ts` + help text update.

## Test plan

- [x] `bun run check` clean (typecheck, lint, format, 1919 pass / 0 fail / 20 skip)
- [x] Users with `--write` only get the same behavior as `--write --live-reads`
- [x] `--live-reads` without `--write` is unchanged
- [x] No flags (default) is unchanged

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -56,7 +56,7 @@ Usage:
 Options:
   --db-path <path>    Path to LevelDB database (default: Copilot Money's default location)
   --timeout <ms>      Decode timeout in milliseconds (default: 90000 = 90 seconds)
-  --write             Enable write tools (sends authenticated requests to Copilot Money's GraphQL API)
+  --write             Enable write tools (implies --live-reads)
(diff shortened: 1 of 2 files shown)


ignaciohermosillacornejo:
Hey @aberhamm, thanks for your PR. If you could just add a test for this I'll approve it and merge it.

aberhamm:
Added — two tests in `tests/unit/cli.test.ts` verifying that `--write` enables live reads. All 1941 tests pass.

ignaciohermosillacornejo (approved):
LGTM

PR #409, opened 2026-06-01
[codex] Discover Chrome profile token stores

konarkm:
## Summary

- discover Chrome token stores from the actual Chrome user-data directory instead of hardcoding only Default plus Profile 1 Local Storage
- scan Default and every Profile N directory, checking Copilot IndexedDB before Local Storage for each profile
- add unit coverage for multi-profile discovery and missing-user-data fallback

## Why

Chrome profile directory names are not stable. A user can have a single active profile stored as Profile 1, Profile 3, or another Profile N directory. Copilot's Firebase Web SDK session is stored in per-profile IndexedDB, so the previous static path list could miss a valid browser session.

## Validation

- bun test tests/core/auth/browser-token.test.ts
- bun run typecheck
- bun run format:check src/core/auth/browser-token.ts tests/core/auth/browser-token.test.ts
- bun run check via pre-push hook

### DIFF
--- a/src/core/auth/browser-token.ts
+++ b/src/core/auth/browser-token.ts
@@ -24,21 +24,51 @@ export interface TokenResult {
 
 /** Firebase refresh token regex: AMf- followed by 100+ URL-safe base64 chars. */
 const REFRESH_TOKEN_REGEX = /AMf-[A-Za-z0-9_-]{100,}/g;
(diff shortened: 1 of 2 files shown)


ignaciohermosillacornejo (approved):
This is a great find, one of those edge cases that's imposible to think of ahead unless you've experienced it lol. LGTM. Will also extend this to the other browsers using your helper.

PR #414, opened 2026-06-06
feat(tools): support name field in update_transaction

aberhamm:
## Summary

- Add `name` as a writable field on `update_transaction`, enabling programmatic transaction renaming
- Add `name` and `category_id` as writable fields on `update_recurring`, enabling programmatic renaming and category reassignment of recurring series
- Both fields discovered by probing the live Copilot Money GraphQL API — `EditTransactionInput` and `EditRecurringInput` accept these fields but they were never tested during the original hidden-mutations recon
- Empty/whitespace-only names are rejected before the write is issued; leading/trailing whitespace is trimmed
- `category_id` on recurring is validated against the local category cache before the write

## Test plan

- [x] `update_transaction` name: 4 new unit tests (dispatch, trim, empty rejection, whitespace rejection)
- [x] `update_transaction` e2e: updated from legacy-rejection to success assertion
- [x] `update_recurring` name: 3 new unit tests (dispatch, trim, empty rejection)
- [x] `update_recurring` category_id: 2 new unit tests (dispatch, non-existent rejection)
- [x] Existing recurring mock responses updated to include `name` and `categoryId`
- [x] Live API round-trip verified: renamed a transaction and a recurring via feature branch code against the production Copilot endpoint
- [x] Full suite passes (1947 pass, 0 fail)

### DIFF
--- a/src/core/graphql/recurrings.ts
+++ b/src/core/graphql/recurrings.ts
@@ -45,11 +45,15 @@ export interface EditRecurringInputRule {
 }
 
 export interface EditRecurringInput {
+  name?: string;
+  categoryId?: string;
   state?: string;
   rule?: EditRecurringInputRule;
 }
 
 export interface EditRecurringChanges {
+  name?: string;
+  categoryId?: string;
   state?: string;
   rule?: EditRecurringInputRule;
 }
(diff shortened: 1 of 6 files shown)


ignaciohermosillacornejo (approved):
LGTM. thanks to this PR, I'm now kicking an entire audit since it seems some other fields where lost when migrating the writes from Firestore to GrapQL

Issue #428, opened 2026-06-10
Epic A — Inner loop: single-source-of-truth enforcement


ignaciohermosillacornejo:
# Epic A — Inner loop: single-source-of-truth enforcement

## Invariant this epic installs

> **Two copies of a fact = a bug that hasn't fired yet.** Any literal list/count/name that exists in more than one place must be either derived from a single constant or guarded by a parity test that fails on drift.

## Why (first principles)

#419 was not an enum bug — it was a **duplication** bug. The wrong frequency list survived because 4 copies drifted independently and nothing compared them. The same pre-#420 pattern still exists for *internal* value-sets (`granularity`, `transaction_type` filters, `view`, `price_type`) and for *documented facts* (CLAUDE.md tool counts, which currently disagree with the code). This epic applies the #420 dedup pattern everywhere a fact has more than one home, and adds gates so new duplicates can't ship.

## Children (sub-issues)

- **A1** — Dedup remaining internal literal sets
- **A2** — Type-derived GraphQL mocks (compile-time conformance for unit tests)
- **A3** — Doc-sync gates (CLAUDE.md counts asserted against code)

## Sequencing

A1 and A3 are independent and parallel-safe. A2 lands **after** A1 (both edit `tests/tools/write-tools*.test.ts`).

All children are `dispatch:cloud-ok` — no live auth needed; cloud CI verifies everything.

---
_Part of the development-system program from the 2026-06-10 testing/dev-system audit. Siblings: Epic B (#421, boundary loop), Epic C (visibility), Epic D (meta loop), Epic E (architecture)._

ignaciohermosillacornejo:
All child issues of this epic are merged and closed (development-system program, 2026-06-10/11). The invariant this epic installs is now enforced by gates on main; ongoing upkeep is covered by the /boundary-audit skill (quarterly) and the weekly scheduled smoke.

PR #448, opened 2026-06-10
A3: Doc-sync gates — fix CLAUDE.md facts and assert tool counts against code

### LINKED ISSUE #434: A3: Doc-sync gates — fix CLAUDE.md facts and assert tool counts against code
Part of **Epic A (#428)**.

## Why

CLAUDE.md is agent-facing infrastructure: a wrong fact there is amplified into every future agent session. It has drifted (verified 2026-06-10):

1. Claims **"34 tools (17 read + 17 write)"** in 3 places — but `createToolSchemas()` returns **14** read tools (+ 17 write = 31 base; `manifest.json` itself says 14)
2. References `tests/fixtures/synthetic-db/`, which doesn't exist (`tests/fixtures/` is empty — the synthetic DB is built at runtime by `tests/helpers/test-db.ts`)
3. `tests/unit/manifest-sync.test.ts` carries a stale header comment ("write schemas… for a future GraphQL-based write path" — writes shipped long ago)

## Spec

1. Fix the three stale facts (decide & document the counting convention — base tools vs live-mode variants — so the number is unambiguous)
2. Add a unit test (suggest `tests/unit/doc-sync.test.ts`) that extracts the tool counts from CLAUDE.md (regex on a pinned phrase) and asserts they equal `createToolSchemas().length` / `createWriteToolSchemas().length` — same pattern as the existing manifest-sync test
3. Sweep README.md / CONTRIBUTING.md for the same counts while in there

## Definition of done

- CLAUDE.md facts true; doc-sync test fails in CI if counts rot again; `bun run check` green

## Agent charter
- `dispatch:cloud-ok`; branch off `main`; PII rules per CLAUDE.md

ignaciohermosillacornejo:
Closes #434 (part of Epic A #428)

## Why

CLAUDE.md is agent-facing infrastructure — a wrong fact there is amplified into every future agent session. It had drifted: it claimed "34 tools (17 read + 17 write)" while `createToolSchemas()` returns 14 read tools (31 base tools total), referenced a nonexistent `tests/fixtures/synthetic-db/` path, and `manifest-sync.test.ts` still described writes as "a future GraphQL-based write path".

## What

**Doc fixes**
- **Counting convention documented** in the CLAUDE.md headline: "base" = cache-mode read tools (`createToolSchemas()`) + write tools (`createWriteToolSchemas()`) = 31 (14 + 17); `--live-reads` swaps 6 cache reads for 13 live tools → 21 read tools in live mode. Live variants are counted separately so the numbers stay unambiguous.
- Fixed all three stale "34 tools (17 read + 17 write)" claims in CLAUDE.md and all three in CONTRIBUTING.md.
- Updated CLAUDE.md's Live Reads bullet, which had also drifted (it listed only 5 swapped reads and 1 net-new tool): now lists all 6 swapped reads + 7 net-new live tools.
- Replaced the `tests/fixtures/synthetic-db/` references (CLAUDE.md + CONTRIBUTING.md) with the real mechanism: synthetic DB generated at runtime by `tests/helpers/test-db.ts`.
- Added `src/tools/live/` to both project-structure trees, which omitted it entirely.
- Rewrote the stale `manifest-sync.test.ts` header — writes shipped long ago and are packaged via `manifest.write.json` (`pack:mcpb:write`).

**Gate**
- New `tests/unit/doc-sync.test.ts` (same pattern as `manifest-sync.test.ts`): derives read/write counts from `createToolSchemas()` / `createWriteToolSchemas()` and live-mode counts from `CopilotMoneyServer.handleListTools()` (default vs `liveReadsEnabled`), then asserts the pinned phrases in CLAUDE.md, CONTRIBUTING.md, and README.md match. No count is hardcoded in the test, so it tracks `src/server.ts`'s actual tool assembly. Also guards against any reintroduced `(X read + Y write)` drift and the dead fixtures path.
- README.md swept per spec — its counts were already correct, so it only gained gate coverage, no edits.

## Verification

- `bun run check` green (typecheck + lint + format:check + 1951 tests)
- Mutation-tested the gate: temporarily restoring the stale "34/(17 read" numbers makes 2 doc-sync tests fail with file-pointing messages

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,6 +1,6 @@
 # Copilot Money MCP Server
 
-MCP (Model Context Protocol) server that enables AI-powered queries and management of Copilot Money personal finance data. Reads come from the locally cached Firestore database (LevelDB + Protocol Buffers); writes go through Copilot's GraphQL API at `app.copilot.money/api/graphql`. 34 tools (17 read + 17 write). Read-only by default, write tools opt-in via `--write` flag.
+MCP (Model Context Protocol) server that enables AI-powered queries and management of Copilot Money personal finance data. Reads come from the locally cached Firestore database (LevelDB + Protocol Buffers); writes go through Copilot's GraphQL API at `app.copilot.money/api/graphql`. 31 base tools (14 read + 17 write); `--live-reads` swaps 6 cache reads for 13 live tools (21 read tools in live mode). Counting convention: "base" = cache-mode read tools (`createToolSchemas()`) + write tools (`createWriteToolSchemas()`); live-mode `_live` variants are counted separately. Read-only by default, write tools opt-in via `--write` flag.
 
 ## Quick Reference
 
@@ -39,7 +39,8 @@ src/
 │   ├── category.ts     # Category mappings (Plaid taxonomy)
 │   └── ...             # Other entity schemas (30+ models)
 ├── tools/
-│   └── tools.ts        # All MCP tool implementations (34 tools)
+│   ├── tools.ts        # Base tool implementations (14 read + 17 write)
+│   └── live/           # 13 GraphQL-backed live read tools (--live-reads mode)
 ├── utils/
 │   ├── date.ts         # Date period parsing (this_month, last_30_days, etc.)
 │   └── categories.ts   # Category name resolution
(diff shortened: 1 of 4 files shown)


ignaciohermosillacornejo:
Addressed all review observations in 370fbfd:

- **`liveToolCount` naming** — added an inline comment clarifying it counts ALL live tool schemas (cache-read replacements + net-new), not just net-new ones.
- **Live-mode `FirebaseAuth` construction** — now injects a stub `GraphQLClient` via the 5th constructor parameter (same pattern as `tests/integration/live-reads.test.ts`), so the gate never touches `FirebaseAuth`/session-token extraction even if the constructor becomes eager about auth.
- **README.md unguarded for `(X read + Y write)`** — added README.md to the generic drift check with `minOccurrences: 0`; any future occurrence of the pattern there is now validated against code counts.
- **Module-scope server construction** — left as-is per the review's own assessment (matches the existing `manifest-sync.test.ts` pattern; `handleListTools()` is synchronous). If a future refactor makes construction async, this will fail loudly, which is the desired behavior for a gate.

Issue #429, opened 2026-06-10
Epic C — Drift visibility: loud-by-default runtime posture


ignaciohermosillacornejo:
# Epic C — Drift visibility: loud-by-default runtime posture

## Invariant this epic installs

> **No silent absorption.** Every boundary failure must be (a) correctly attributed — our bug, server drift, user error, or transient network — and (b) visible at the right altitude: user-facing when data is affected, logged when not. And drift detection must be **time-triggered**, not only activity-triggered.

## Why (first principles)

The #419 symptom was a *misleading* error ("Copilot's API changed… report this issue") for a client-side bug — misattribution delayed diagnosis. Read-side schema mismatches silently drop documents from the cache with only a deduped stderr warning (`src/core/schema-warn.ts`), so a user sees incomplete finances with no signal. And every existing gate fires only when we push code: if Copilot changes the API during a quiet month, the first detector is a confused user.

## Children (sub-issues)

- **C1** — Scheduled smoke runs (time-triggered drift detection; local-only)
- **C2** — Error taxonomy: attribute boundary failures correctly, surface raw GraphQL errors
- **C3** — Read-side drop visibility in `get_cache_info` / `get_connection_status`
- **C4** — Bounded retry/backoff for network-level failures (never blind mutation retries)

## Sequencing

C2, C3 are independent and parallel-safe (`dispatch:cloud-ok`). C4 lands **after** C2 (same files: `client.ts`/`errors.ts`). C1 is `dispatch:local-only` (launchd on the machine with browser auth).

---
_Part of the development-system program from the 2026-06-10 testing/dev-system audit. Siblings: Epic A (inner loop), Epic B (#421, boundary loop), Epic D (meta loop), Epic E (architecture)._

ignaciohermosillacornejo:
All child issues of this epic are merged and closed (development-system program, 2026-06-10/11). The invariant this epic installs is now enforced by gates on main; ongoing upkeep is covered by the /boundary-audit skill (quarterly) and the weekly scheduled smoke.

Issue #419, opened 2026-06-08
create_recurring: frequency validation rejects valid cadences (ANNUALLY etc.) and accepts invalid YEARLY


ignaciohermosillacornejo:
## Bug: `create_recurring` frequency validation is wrong — rejects valid cadences, accepts an invalid one

`create_recurring`'s frequency allowlist is `['WEEKLY', 'BIWEEKLY', 'MONTHLY', 'YEARLY']` in **both** the runtime guard (`src/tools/tools.ts` ~3452) and the JSON-schema `enum` (~4986). The real GraphQL `RecurringFrequency` enum has **8** values and uses `ANNUALLY`, not `YEARLY`.

### Impact (confirmed live against production)

Running `create_recurring` today:

| frequency | What happens |
|---|---|
| `YEARLY` | Tool **accepts** it → server returns `SCHEMA_ERROR` (400) → surfaced to the user as the **misleading** *"Copilot's API changed in a way this tool doesn't handle yet. Please report this issue."* |
| `ANNUALLY` | Rejected **locally**: *"frequency must be one of: WEEKLY, BIWEEKLY, MONTHLY, YEARLY"* |
| `BIMONTHLY` / `QUARTERLY` / `QUADMONTHLY` / `SEMIANNUALLY` | Rejected locally — impossible to create these cadences |
| `WEEKLY` / `BIWEEKLY` / `MONTHLY` | Work |

So **you cannot create a yearly (or bimonthly/quarterly/quadmonthly/semiannual) recurring through the tool at all**, and the natural value `YEARLY` fails with an error that wrongly blames Copilot's API.

### Root cause

The list was hardcoded in `059739b` (the Firestore→GraphQL write migration — the same migration family that dropped `name` from `update_transaction`). The values were **assumed, never probed**: `YEARLY` matches the lowercase cache model (`KNOWN_FREQUENCIES` has `'yearly'`) but the GraphQL enum value is `ANNUALLY`, and only 4 of 8 values were listed. Introspection is disabled, so verifying would have required an error-leak probe (which now confirms the correct set).

A secondary issue: `graphQLErrorToMcpError` maps the server's `SCHEMA_ERROR` to *"Copilot's API changed… report this issue"*, which disguises a client-side validation bug as an upstream change.

### Correct values (probed + matches the Copilot UI + #417)

[...]

PR #449, opened 2026-06-10
feat(errors): C2 error taxonomy — classify GraphQL errors by content, surface raw server text

### LINKED ISSUE #441: C2: Error taxonomy — attribute boundary failures correctly, surface raw GraphQL errors
Part of **Epic C (#429)**.

## Why

#419's user-facing symptom was the generic `SCHEMA_ERROR` message — *"Copilot's API changed in a way this tool doesn't handle yet. Please report this issue."* — blaming the server for a client-side bug and hiding the server's actual error text. Misattribution delays diagnosis; this is the loud-by-default principle applied to error *quality*.

## Spec

- In `src/core/graphql/client.ts` + `src/tools/errors.ts`: classify GraphQL errors by content (parse-time validation rejections vs input/ownership errors vs auth vs transport) instead of mapping all unrecognized 400/500s to one bucket
- **Always include the server's raw error message** (truncated) in the surfaced error so the user/agent sees what the server actually said
- Three user-facing attributions: "this tool's model of the API may be outdated — run `bun run smoke` and report with the text below" vs "the server rejected the request: <server reason>" vs "transient network problem — retry"
- Keep error codes stable; while in here, prefer tests asserting **codes** over full message strings (reduces the error-string fragility flagged in the audit)

## Definition of done

- Unit tests per classification branch using realistic GraphQL error payloads; no path emits the old misleading blanket message; `bun run check` green

## Agent charter
- `dispatch:cloud-ok`; branch off `main`; PII rules per CLAUDE.md
- C4 (retry) builds on this classification — land C2 first

ignaciohermosillacornejo:
Closes #441 — part of Epic C (#429).

## Why

#419's user-facing symptom was the blanket `SCHEMA_ERROR` message — *"Copilot's API changed in a way this tool doesn't handle yet"* — which blamed the server for a client-side bug and hid the server's actual error text. This PR attributes boundary failures correctly and makes every error path loud about what the server actually said.

## What changed

**`src/core/graphql/client.ts`**
- Non-2xx responses are now classified by body content first (Apollo `errors[].extensions.code`), HTTP status as fallback:
  - `GRAPHQL_PARSE_FAILED` / `GRAPHQL_VALIDATION_FAILED` / `BAD_USER_INPUT` / persisted-query codes → `SCHEMA_ERROR` (our model of the API is wrong)
  - `NOT_FOUND` / `FORBIDDEN` → `USER_ACTION_REQUIRED` (server rejected for actionable reasons)
  - `UNAUTHENTICATED` → `AUTH_FAILED` at **any** HTTP status, including 2xx bodies (token expiry can surface as HTTP 200 + errors[])
  - 5xx → new `SERVER_ERROR` code (possibly transient), no longer lumped into `SCHEMA_ERROR`
  - bare 400 without a recognizable extension code → still `SCHEMA_ERROR`
- Every thrown `GraphQLError` carries the server's raw error message, truncated at 600 chars, plus the parsed `errors[]` as `serverErrors`

**`src/tools/errors.ts`** — three correct attributions per the issue spec:
- `SCHEMA_ERROR` → "This tool's model of Copilot's API may be outdated — run `bun run smoke` and report with the text below"
- `USER_ACTION_REQUIRED` → "Copilot's server rejected the request: \<server reason\>"
- `SERVER_ERROR` / `NETWORK` → transient, retry
- Every branch (including `AUTH_FAILED` and `UNKNOWN`) now embeds the underlying server text

**Tests** — per-classification-branch unit tests with realistic GraphQL error payloads (including the #419 enum-coercion shape); assertions check **codes + attribution fragments** rather than full message strings, reducing the error-string fragility flagged in the audit; an explicit test verifies no path emits the old blanket message.

## Definition of done

- [x] Unit tests per classification branch using realistic GraphQL error payloads
- [x] No path emits the old misleading blanket message
- [x] `bun run check` green (1952 pass / 0 fail)

Note for C4 (retry): `SERVER_ERROR` and `NETWORK` are the retryable buckets this classification sets up.

### DIFF
--- a/src/core/graphql/client.ts
+++ b/src/core/graphql/client.ts
@@ -58,18 +58,17 @@ function parseErrorBody(text: string): ParsedErrorBody | undefined {
 }
 
 /**
- * Classify a non-2xx response by CONTENT (Apollo `errors[].extensions.code`)
- * first, falling back to HTTP status. This is what attributes failures
- * correctly:
- *  - parse/validation/coercion rejections → SCHEMA_ERROR (OUR request shape
- *    or enum model doesn't match the server schema — a client-side bug)
- *  - ownership / resource errors → USER_ACTION_REQUIRED (server rejected
- *    the request for reasons the user/agent can act on)
- *  - auth → AUTH_FAILED; 5xx → SERVER_ERROR (possibly transient)
+ * Classify by Apollo `errors[].extensions.code` alone (shared by the non-2xx
+ * and 2xx-with-errors[] paths). Returns undefined when no recognizable code
+ * is present so callers can apply their own status-based fallback.
+ *
+ * Priority is intentional: auth first, then ownership/resource errors
+ * (USER_ACTION_REQUIRED) BEFORE schema codes — if a response carries both,
+ * the ownership error is the one the user/agent can act on, while the schema
+ * code on the same response is usually downstream noise. Don't reorder.
  */
-function classifyHttpError(status: number, parsed?: ParsedErrorBody): GraphQLErrorCode {
-  const codes = parsed?.extensionCodes ?? new Set<string>();
-  if (status === 401 || codes.has('UNAUTHENTICATED')) return 'AUTH_FAILED';
+function classifyExtensionCodes(codes: Set<string>): GraphQLErrorCode | undefined {
+  if (codes.has('UNAUTHENTICATED')) return 'AUTH_FAILED';
   if (codes.has('FORBIDDEN') || codes.has('NOT_FOUND')) return 'USER_ACTION_REQUIRED';
   if (
     codes.has('GRAPHQL_PARSE_FAILED') ||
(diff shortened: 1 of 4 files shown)


ignaciohermosillacornejo:
Addressed the review in 4441172:

1. **2xx + errors[] re-classification** — extracted `classifyExtensionCodes()` (shared by both paths); the 2xx-with-errors branch now routes through it, so 200 + `GRAPHQL_VALIDATION_FAILED` maps to `SCHEMA_ERROR` and only falls back to `USER_ACTION_REQUIRED` when no recognizable code is present. Added a test for exactly that shape.
2. **Redundant HTTP status in AUTH_FAILED message** — accepted as-is: the `${status}: ${text}` prefix is uniform across all non-2xx errors, and keeping it means `e.message` is self-contained wherever it's logged or re-wrapped (the duplication only appears in the one branch where the code already implies the status). Happy to revisit if it bothers in practice.
3. **Priority ordering** — documented in the `classifyExtensionCodes` doc comment (auth > ownership > schema codes, with the rationale and a "don't reorder" note).
4. **FORBIDDEN test gap** — added a dedicated `FORBIDDEN` → `USER_ACTION_REQUIRED` test.

`bun run check` green (1954 pass / 0 fail).

Issue #435, opened 2026-06-10
B1: Conformance ledger — machine-readable inventory of every external assumption and its oracle


ignaciohermosillacornejo:
Part of **Epic B (#421)** — the keystone child; **do this first**. (Promotion of the original Phase-5 "coverage manifest" to the spine of the program.)

## Why

The audit's core finding: bugs lived in the gap between our model and Copilot's reality, and nothing tracked *which* assumptions had independent verification. The ledger makes "we forgot to verify" a **red build** instead of a silent state, and gives every future PR a place to record evidence. It's also the cloud-visible shadow of the local smoke gate: cloud CI can't probe the server, but it *can* enforce that every assumption is accounted for.

## Spec

- **`conformance/ledger.ts`** (typed, checked in): one entry per external assumption:
  ```ts
  { surface: 'CreateTransactionInput.tagIds',
    kind: 'enum' | 'input-field' | 'response-shape' | 'operation',
    oracle: 'smoke:conformance' | 'smoke:input-fields' | 'runtime:zod-warn' | null,
    class: 'gated' | 'verified-once' | 'unverified',
    evidence: 'PR #418 live probe, 2026-06-08' }
  ```
- A plain unit test (cloud CI, no auth):
  (a) every param/enum reachable from `createWriteToolSchemas()` has a ledger entry;
  (b) every named smoke oracle maps to an existing `scripts/smoke/` script;
  (c) `class: 'gated'` requires a non-null oracle
- **Seed from the 2026-06-10 audit:** `RecurringFrequency`/`RecurringState`/`TransactionType` = `gated`; the ~11 input types (`CreateTransactionInput`, `EditTransactionInput`, `SplitTransactionInput`, `AddTransactionToRecurringInput`, `CreateRecurringInput`, `EditRecurringInput`, `CreateCategoryInput`, `EditCategoryInput`, `CreateTagInput`, `EditTagInput`, `EditAccountInput`) = `verified-once` (cite the probing PR: #414/#417/#418/#420 lineage) or `unverified`; response shapes = `unverified`; `TimeFrame` (`src/core/graphql/queries/_shared.ts`) = `unverified`
- `bun run smoke` prints the ledger's class distribution at the end — the "are we getting better" number

## Definition of done

- Ledger + enforcement test in CI; a new write-tool param without an entry fails the build; class distribution printable; `bun run check` green

## Agent charter
[...]

Issue #421, opened 2026-06-08
Epic B — Boundary loop: live conformance coverage (no fact without an oracle)


ignaciohermosillacornejo:
# Epic B — Boundary loop: no fact without an oracle

## Invariant this epic installs

> **Every assumption about a system we don't control is either gated by an independent oracle, or listed in the conformance ledger as dated, visible debt.** Unit tests validate the code against its own assumptions *by construction*; only checks against the live server (or recorded evidence from it) validate the assumptions themselves.

## Why (original analysis, still the core)

Two bugs (`update_transaction.name` dropped untested in the migration; `create_recurring` advertising invalid `YEARLY` while rejecting 5 valid cadences) shared one root cause: **our code's model of Copilot's GraphQL schema drifted from the real schema, and the only check on it was a mock that shared the wrong assumption.** Introspection is disabled server-side, so conformance comes from validation-only error-leak probes (`docs/graphql-capture/introspection-recon.md`) and reversible round-trips. These need a live browser-session token, so this is a **local gate**, not cloud CI — but the ledger test (B1) is the cloud-visible shadow of that gate: CI can't probe the server, but it *can* fail when an assumption isn't accounted for.

## Status of the original phases

- [x] **Phase 0 — scaffolding:** reusable harness `scripts/smoke/_conformance.ts`, `bun run smoke` runner with PASS/FAIL table (#420, #422)
- [x] **Phase 1 — enum conformance:** `RecurringFrequency`, `RecurringState`, `TransactionType`, each with a known-bad control (#420, #422, #424). `ColorName` moved to B5.
- [ ] Remaining work restructured as sub-issues below (Phase 5's "coverage manifest" promoted to keystone B1; response-shape validation added as new scope B3)

## Children (sub-issues)

| Child | Scope | Dispatch |
|---|---|---|
| **B1 #435** | **Conformance ledger** — machine-readable assumption inventory + CI enforcement. Do first; everything else records results into it | cloud-ok |
| **B2 #436** | Input-field conformance for all ~11 mutation input types (was Phase 2) | code cloud-ok, live gate local |
| **B3 #437** | Response-shape Zod warn-mode — the server→us direction (new scope) | code cloud-ok, live gate local |
| **B4 #438** | Reversible round-trips per write tool (was Phase 3) | local-only |
| **B5 #439** | Read-smoke completion + `TimeFrame`/`ColorName` stragglers (was Phase 4) | code cloud-ok, live run local |
[...]

Issue #438, opened 2026-06-10
B4: Reversible round-trip smokes per write tool (Phase 3)


ignaciohermosillacornejo:
Part of **Epic B (#421)** — Tier 2: prove writes *apply*, not just validate.

## Spec

- One reversible round-trip per write tool: create→verify→delete, or set→verify→revert (the #417 pattern: flip a recurring's frequency, confirm the echo, flip it back)
- **Guaranteed cleanup:** try/finally; idempotent re-run safety; refuse to start if a previous run left residue (marker convention, e.g. objects named `__smoke__<timestamp>`)
- Separate runner (`bun run smoke:roundtrip`) — **never** part of the default `bun run smoke` (Tier 1 stays non-mutating; C1's schedule runs Tier 1 only)
- Ledger (#435): per-tool "applies" entries → `gated` with oracle `smoke:roundtrip`

## Rules of engagement (hard constraints)

- Mutating — local-only, run consciously, never scheduled
- Cleanup verified in the same run via a final sweep query; any residue = FAIL loudly with the leftover IDs
- Synthetic names/amounts only — no real PII in created objects

## Definition of done

- Every write tool in `src/tools/tools.ts` has a round-trip; a full run leaves zero residue (sweep-verified)

## Agent charter
- `dispatch:local-only` (cloud agents may draft the scripts; execution and the merge gate are local)
- Sequencing: **after B2** — field conformance first, it's the cheaper detector for the same risk class

PR #467, opened 2026-06-11
feat(graphql): B3 — warn-mode Zod validation on mutation responses

### LINKED ISSUE #437: B3: Response-shape validation — Zod warn-mode on mutation responses (schema-warn for the write path)
Part of **Epic B (#421)**. New scope the original issue didn't cover: #421 gates what we *send*; nothing gates what comes *back*.

## Why

Write-path response shapes are hand-typed casts (e.g. `CreatedTransaction`, `src/core/graphql/transactions.ts:47-65`): if the server renames/removes a response field, code reads `undefined` and output degrades silently. The read path already solved this pattern — `src/core/schema-warn.ts` validates and warns with dedup. This is the same idea pointed at the GraphQL responses.

## Spec

- Zod schemas for mutation (and live-read) response payloads — derive field lists from the existing interfaces; keep passthrough semantics so *new* server fields still flow through
- Parse in `GraphQLClient` (`src/core/graphql/client.ts`) or per-module response handling, **warn-mode**: structured drift warnings (reuse/extend schema-warn's dedup), never hard-fail a user operation on shape drift
- Expose drift counters so C3's visibility work can surface them
- Ledger (#435): response-shape entries `unverified` → `gated` with oracle `runtime:zod-warn`

## Definition of done

- A drifted response field produces a loud structured warning on first occurrence (unit-tested with a deliberately wrong mock); zero behavior change on the happy path; `bun run check` green
- **Live gate:** [ ] one real write round-trip locally confirms no false-positive warnings

## Agent charter
- Code/tests: `dispatch:cloud-ok`; live confirmation local
- Mild file overlap with B2 (same graphql modules) — land sequentially relative to B2
- Branch off `main`; PII rules per CLAUDE.md

ignaciohermosillacornejo:
# Summary

Adds warn-mode response-shape validation for the write path (Epic B #421): every mutation response is validated against a Zod schema mirroring the hand-written response interface, so server-side renames/removals of fields we read surface as a loud structured warning + drift counter instead of silent downstream `undefined`s. Closes #437

**How it works**

- `src/core/graphql/response-validation.ts` — registry of 17 Zod schemas (one per mutation operation), derived field-for-field from the existing response interfaces in the per-domain modules. All objects are `z.looseObject`, so **new** server fields (and `__typename`) pass through silently — drift means a field we *read* going missing or changing type.
- `GraphQLClient.request()` calls `validateMutationResponse()` after every successful mutation. **Warn-mode**: it never throws and never alters the payload — consistent with the C3 drop-visibility pattern. Warnings are deduped per `(operation, path, code)` per process (same scheme as `schema-warn.ts`); per-surface drift counts are NOT deduped and are exposed via `getResponseDriftStats()` so C3's visibility work can surface them.
- An unregistered mutation operation name also gets a (deduped) warning, so a future mutation wrapper can't silently skip the gate.
- Ledger (#435): all 17 `Mutation.*:response` surfaces upgrade `unverified` → `gated` with oracle `runtime:zod-warn`; `zod-warn` is registered in `RUNTIME_CHECK_NAMES`. A new ledger test enforces the registry↔ledger mapping bidirectionally (a ledger entry claiming the oracle without a schema, or a schema without a ledger entry, fails the build).

**Tests**: valid synthetic fixtures for all 17 operations pass silently; deliberately wrong mocks (removed field, retyped field, garbage payloads) warn once, count every occurrence, never throw, and leave the payload untouched; client-level tests confirm a drifted mutation response still resolves with the raw payload and that query responses are not shape-validated.

**Live gate (pending)**: the DoD's real write round-trip (no false-positive warnings on a live mutation) is to be run by the maintainer via `bun run smoke` before merge — the smoke suite's write round-trips now exercise these validators on every run.

## External assumptions

No NEW assumptions about Copilot's API — this PR converts 17 *existing* response-shape assumptions (previously `unverified`, hand-typed TS casts) into runtime-checked ones:

- `Mutation.createTransaction:response`, `Mutation.editTransaction:response`, `Mutation.deleteTransaction:response`, `Mutation.addTransactionToRecurring:response`, `Mutation.splitTransaction:response`, `Mutation.createTag:response`, `Mutation.editTag:response`, `Mutation.deleteTag:response`, `Mutation.createCategory:response`, `Mutation.editCategory:response`, `Mutation.deleteCategory:response`, `Mutation.editCategoryBudget:response`, `Mutation.editCategoryBudgetMonthly:response`, `Mutation.createRecurring:response`, `Mutation.editRecurring:response`, `Mutation.deleteRecurring:response`, `Mutation.editAccount:response`
- Evidence class: **gated** (`runtime:zod-warn`) — schemas transcribed from the captured-response interfaces; every live mutation re-verifies the shape warn-mode at runtime. Maintainer's pre-merge `bun run smoke` run doubles as the live round-trip confirmation that there are no false positives.

### DIFF
--- a/src/conformance/ledger.ts
+++ b/src/conformance/ledger.ts
@@ -134,7 +134,7 @@ const FIELD_PROBE_GATED =
 const RESPONSE_SHAPE_GATED =
   'Hand-written TS interface mirrored into a Zod schema; every mutation response ' +
   'is validated warn-mode at runtime (src/core/graphql/response-validation.ts, ' +
-  'issue #437, PR #TBD)';
+  'issue #437, PR #467)';
 
 // ---------------------------------------------------------------------------
 // Entry factories (keep the inventory compact; pass overrides to upgrade an
--- a/src/conformance/ledger.ts
+++ b/src/conformance/ledger.ts
@@ -41,6 +41,7 @@
 import { TRANSACTION_TYPES } from '../core/graphql/transactions.js';
 import { RECURRING_FREQUENCIES, RECURRING_STATE_VALUES } from '../core/graphql/recurrings.js';
 import { ALL_TIME_FRAMES } from '../core/graphql/queries/_shared.js';
+import { RESPONSE_SHAPE_RUNTIME_CHECK } from '../core/graphql/response-validation.js';
 
 /** What kind of external surface the assumption is about. */
 export const SURFACE_KINDS = ['enum', 'input-field', 'response-shape', 'operation'] as const;
@@ -71,10 +72,13 @@ export type ConformanceOracle = `smoke:${string}` | `runtime:${string}`;
 
 /**
  * Registered always-on runtime checks that `runtime:<name>` oracles may
- * reference. Empty until B3 (#437) lands the zod warn-mode response
- * validators; B3 should register e.g. 'zod-warn' here when it ships.
+ * reference.
+ * - `zod-warn` (B3, #437): every mutation response is validated warn-mode
+ *   against a Zod schema mirroring the hand-written response interface
+ *   (src/core/graphql/response-validation.ts); drift logs a structured
+ *   warning and increments a per-surface counter, never failing the call.
  */
-export const RUNTIME_CHECK_NAMES: readonly string[] = [];
+export const RUNTIME_CHECK_NAMES: readonly string[] = [RESPONSE_SHAPE_RUNTIME_CHECK];
 
 export interface LedgerEntry {
   /** External assumption surface (see naming convention above). Unique. */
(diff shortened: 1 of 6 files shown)


ignaciohermosillacornejo:
Addressed both review rounds in b06cb27 (the `PR #TBD` placeholder was already fixed in fcd4ed6):

- **Multi-field drift / first-issue-wins** (round 1 issue 2, round 2 obs. 3): `validateMutationResponse` now warns for **every** Zod issue in a drifted response, each deduped per `(operation, path, code)` — the full drift inventory gets logged while repeats stay silent, and the dedup set still bounds total output. New unit test covers two simultaneously missing fields (2 warns, counted once per response, silent on repeat).
- **Retry-path coverage** (round 2 obs. 4): added a test where a mutation fails with `ECONNREFUSED` on attempt 1 and succeeds with a drifted payload on attempt 2 — validation fires on the retried response.
- **EditTransaction subset schema** (round 2 obs. 2): comment now states explicitly that gating only the read-back subset (no `type`, `amount`, …) is deliberate.
- **New-enum-value warns** (round 1 nit): documented on `CreatedTransactionSchema` — a new server `TransactionType` warning is the drift signal working, not a bug.
- **`__resetResponseDriftState` export** (round 2 obs. 1): kept as-is for consistency with `schema-warn.ts`'s `__resetWarnedKeys` convention; happy to follow up if a test-only-export marker lands repo-wide.
- **Registry keyed by operation name** (round 2 obs. 5): acknowledged — inherent to the design and documented at the registry; the bidirectional ledger test plus the unregistered-operation runtime warning are the existing guards.

ignaciohermosillacornejo:
Live gate completed 2026-06-11 (maintainer session): `bun run smoke` against production from this branch — **PASS** (all 3 enums + 12 input types). Ledger class distribution on this branch: 63 gated (75%) / 20 verified-once (24%) / 1 unverified (TimeFrame). The runtime zod-warn path gets its first real-write exercise during the #438 round-trip smokes.

PR #471, opened 2026-06-12
feat(smoke): Tier-2 reversible round-trip smokes per write tool (B4)

### LINKED ISSUE #438: B4: Reversible round-trip smokes per write tool (Phase 3)

ignaciohermosillacornejo:
# Summary

One reversible round-trip per write tool (17/17), each verifying the write by **independent re-read** — the detector for accepted-but-ignored writes. New `bun run smoke:roundtrip` runner (separate from `bun run smoke`; Tier 1 stays non-mutating), per-write-tool `applies` ledger surfaces gated by the `smoke:roundtrip` oracle, and a coverage ratchet so a new write tool cannot ship without a round-trip. Closes #438

> **MUTATING SUITE — drafted, not executed.** Per the rules of engagement this suite was written and unit-tested mock-only; nothing here was run against the live endpoint. **The maintainer's attended `bun run smoke:roundtrip` is the merge gate for this PR.** `--list` prints the full plan without auth or writes.

## Round-trip plan (run order)

| # | tool | flow | target |
|---|------|------|--------|
| 1 | create_tag | create → verify (Tags re-read) | run-created tag |
| 2 | update_tag | rename+recolor → verify | run-created tag |
| 3 | create_category | create → verify (Categories re-read) | run-created category |
| 4 | update_category | rename+toggle isExcluded → verify → untoggle | run-created category |
| 5 | set_budget | default 100 → verify; monthly 200 → verify; clear both in finally | run-created category |
| 6 | create_transaction | create (amount 100) → verify (Transactions re-read) | new txn on first account |
| 7 | update_transaction | rename+note → verify | run-created txn |
| 8 | review_transactions | capture isReviewed → flip → verify → restore in finally | run-created txn |
| 9 | create_recurring | create MONTHLY from run-created txn → verify (Recurrings re-read) | run-created recurring |
| 10 | update_recurring | MONTHLY→WEEKLY → verify | run-created recurring |
| 11 | set_recurring_state | capture state → flip ACTIVE↔PAUSED → verify → restore in finally | run-created recurring |
| 12 | add_transaction_to_recurring | create 2nd txn → attach → verify recurringId | run-created txn + recurring |
| 13 | split_transaction | create 3rd txn (200) → split 2×100 → verify children via re-read | run-created txn |
| 14 | delete_recurring | delete → verify absence | run-created recurring |
| 15 | delete_transaction | delete → verify absence | run-created txn |
| 16 | delete_tag | delete → verify absence | run-created tag |
| 17 | delete_category | delete → verify absence | run-created category |

**No check mutates pre-existing user data.** Every mutator/deleter operates on an object the run itself created (lazy prerequisites keep that true under `--only <domain>`). The single unavoidable contact with user data: created transactions must reference a real `account_id`/`item_id` (first account from the Accounts query, same pick as the Tier-0 read smoke); they are deleted and sweep-verified in the same run.

## Safety mechanics

- Marker `__smoke__<timestamp>` on every created object; synthetic names/amounts only (100/200; a source-scan test enforces the amount set).
- **Pre-flight residue check** refuses to start if marker objects from a prior run exist.
- **Guaranteed cleanup**: LIFO `CleanupRegistry` runs in `finally` (continues past failures, reports every leftover id); checks that delete their own target deregister only after verifying absence.
- **Final sweep** re-queries tags/categories/recurrings/transactions for marker residue and fails loudly with leftover ids. Recurrings additionally match `\bsmoke\b` (case-insensitive, word-bounded) because the server derives recurring names and may normalize the raw marker away.
- Set→revert flows capture the original BEFORE mutating and restore it in `finally` even when verification fails (unit-tested with a stub client).
- B3 zod-warn drift counters are printed in the run summary (come for free on every mutation the run sends).
- `bulkEditTransactions` is never used — enforced by a source scan in the ratchet test.
- Idempotent re-run safety: per-run timestamp marker + pre-flight refusal + name-based sweep (catches residue across crashed runs, not just tracked ids).

## Tools flagged for maintainer judgment (none blocked, two caveats)

- **split_transaction**: Copilot has no unsplit mutation. The flow is still safe — parent and both children are run-created and deleted in cleanup. Caveat: Copilot hides (not deletes) split parents in its UI; if the Transactions query also hides them, a failed parent delete would be invisible to the name sweep — the cleanup failure report (with id) is the loud signal in that case.
- **set_budget / update_account class**: `set_budget` avoids touching real budgets by budgeting the run-created category. There is no `update_account` tool (the `editAccount` wrapper has no MCP tool), so no account round-trip exists — noted in the ledger with instructions for when one ships.

## Ledger entries touched (src/conformance/ledger.ts)

- New surface kind `applies` (`Mutation.<x>:applies` = "the write is persisted and visible on independent re-read").
- 16 new entries, all `gated` / oracle `smoke:roundtrip`: createTransaction, editTransaction, deleteTransaction, addTransactionToRecurring, splitTransaction, createTag, editTag, deleteTag, createCategory, editCategory, deleteCategory, editCategoryBudget, editCategoryBudgetMonthly, createRecurring, editRecurring, deleteRecurring.
- `editAccount` deliberately has no applies entry (no MCP write tool) — comment in the ledger says when to add it.
- Ratchets: `tests/scripts/roundtrip-coverage.test.ts` enforces WRITE_TOOL_DEFS ↔ checks bijection and checks ↔ `:applies` surface coverage; the existing ledger test verifies the `smoke:roundtrip` oracle script exists.

## External assumptions

All of the below are per-tool "the write actually applies and is visible on re-read" assumptions. Evidence class for every one: **unverified — live gate pending** (the drafting agent cannot run the mutating oracle; the maintainer's attended `bun run smoke:roundtrip` upgrades them to live-verified in one run):

- `Mutation.createTransaction:applies`, `Mutation.editTransaction:applies` (update_transaction + review_transactions), `Mutation.deleteTransaction:applies`, `Mutation.addTransactionToRecurring:applies`, `Mutation.splitTransaction:applies`
- `Mutation.createTag:applies`, `Mutation.editTag:applies`, `Mutation.deleteTag:applies`
- `Mutation.createCategory:applies`, `Mutation.editCategory:applies`, `Mutation.deleteCategory:applies`
- `Mutation.editCategoryBudget:applies`, `Mutation.editCategoryBudgetMonthly:applies` (both branches of set_budget)
- `Mutation.createRecurring:applies`, `Mutation.editRecurring:applies` (update_recurring + set_recurring_state), `Mutation.deleteRecurring:applies`

Secondary unverified assumptions the attended run will confirm (assertions are written to fail loudly, not silently, if wrong): which of `budget.current.amount` / `resolvedAmount` reflects a freshly-set default budget; that the Transactions `matchString` filter matches the `__smoke__` marker substring; that server-derived recurring names retain a sweep-detectable form of the marker.

## Verification

- `bun run check` green (typecheck + lint + format + version-sync + 2130 tests, including 42 new mock-only tests — no network).
- `bun run sync-manifest` clean (no tool changes).
- `bun run smoke:roundtrip -- --list` exercised locally (no auth, no writes) — plan renders correctly; invalid `--only` rejected with usage.
- NOT executed: `bun run smoke:roundtrip` (mutating) — maintainer-run merge gate.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "fix": "bun run lint:fix && bun run format",
     "smoke": "bun run scripts/smoke/conformance.ts && bun run scripts/smoke/reads.ts",
     "smoke:conformance": "bun run scripts/smoke/conformance.ts",
+    "smoke:roundtrip": "bun run scripts/smoke/roundtrip.ts",
     "smoke:reads": "bun run scripts/smoke/reads.ts",
     "smoke:recurring-frequency": "bun run scripts/smoke/recurring-frequency-conformance.ts",
     "smoke:recurring-state": "bun run scripts/smoke/recurring-state-conformance.ts",
--- a/scripts/smoke/roundtrip-checks.ts
+++ b/scripts/smoke/roundtrip-checks.ts
@@ -414,22 +414,33 @@ export const ROUNDTRIP_CHECKS: readonly RoundtripCheck[] = [
       const categoryId = ctx.state.categoryId;
       if (!categoryId) return { skipped: 'no run-created category (create_category did not pass)' };
       const name = `${ctx.state.marker}-cat-edited`;
-      await editCategory(ctx.client, { id: categoryId, input: { name, isExcluded: true } });
-      const after = (await fetchCategories(ctx.client, { rollovers: false })).find(
-        (cat) => cat.id === categoryId
-      );
-      check(after, `update_category: category ${categoryId} missing from Categories re-read`);
-      check(
-        after.name === name,
-        `update_category: write accepted but re-read name is '${after.name}', expected '${name}'`
-      );
-      check(
-        after.isExcluded === true,
-        'update_category: write accepted but re-read isExcluded is still false'
-      );
-      // Restore isExcluded=false so the budget check below behaves like a
-      // normal category (the category itself is deleted at the end either way).
-      await editCategory(ctx.client, { id: categoryId, input: { isExcluded: false } });
+      try {
+        await editCategory(ctx.client, { id: categoryId, input: { name, isExcluded: true } });
+        const after = (await fetchCategories(ctx.client, { rollovers: false })).find(
+          (cat) => cat.id === categoryId
+        );
+        check(after, `update_category: category ${categoryId} missing from Categories re-read`);
+        check(
+          after.name === name,
+          `update_category: write accepted but re-read name is '${after.name}', expected '${name}'`
+        );
+        check(
+          after.isExcluded === true,
+          'update_category: write accepted but re-read isExcluded is still false'
+        );
+      } finally {
+        // Restore isExcluded=false even when verification failed, so the
+        // set_budget check downstream sees a normal (non-excluded) category
+        // (the category itself is deleted at the end either way).
+        try {
+          await editCategory(ctx.client, { id: categoryId, input: { isExcluded: false } });
+        } catch (err: unknown) {
+          ctx.log('update_category: WARNING — failed to restore isExcluded', {
+            categoryId,
+            error: err instanceof Error ? err.message : String(err),
+          });
+        }
+      }
       return undefined;
     },
   },
@@ -596,10 +607,12 @@ export const ROUNDTRIP_CHECKS: readonly RoundtripCheck[] = [
       ctx.registry.add({
         kind: 'recurring',
         id: created.id,
-        // The server derives the recurring's name from the transaction, so it
-        // should carry the marker — but that derivation is the server's, so
-        // the label records what the echo said.
-        label: created.name,
+        // Deliberately NOT `created.name`: the server derives the recurring's
+        // name (usually from the marker-bearing transaction, but that
+        // derivation is the server's), and labels are logged — so a
+        // server-derived name could leak real merchant data into output.
+        // A synthetic marker label keeps logs PII-safe regardless.
+        label: `${ctx.state.marker}-recurring`,
         cleanup: () => deleteRecurring(ctx.client, { id: created.id }).then(() => undefined),
       });
       const after = (await fetchRecurrings(ctx.client)).find((rec) => rec.id === created.id);
@@ -846,6 +859,12 @@ export interface ResidueReaders {
  * whose name was normalized away from the raw `__smoke__` marker. The
  * word-boundary regex catches 'Smoke 176...' style derivations without
  * matching e.g. 'Smokehouse BBQ'.
+ *
+ * Known approximation: a real recurring whose name contains the standalone
+ * word ("Smoke's BBQ", "Smoke Shop") WILL match, and the pre-flight check
+ * will refuse to start. That's the safe direction for an attended local
+ * gate — a false refusal is a one-line message the maintainer can act on,
+ * while a false negative would leave residue undetected.
  */
 const RECURRING_RESIDUE_RE = /\bsmoke\b/i;
 
--- a/scripts/smoke/roundtrip-checks.ts
+++ b/scripts/smoke/roundtrip-checks.ts
@@ -1,0 +1,924 @@
+/**
+ * Tier-2 reversible round-trip check definitions (issue #438, Epic B #421).
+ *
+ * One round-trip per MCP write tool: create→verify→delete, or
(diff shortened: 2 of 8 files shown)


ignaciohermosillacornejo:
Addressed all review findings in f9ad97b:

- **Medium (update_category try/finally)**: the isExcluded toggle + verification now sit inside `try`, with the restore in `finally` (warn-logged if the restore itself fails) — same pattern as `review_transactions` / `set_recurring_state`. A mid-flight verification failure can no longer leave the category excluded and cascade into `set_budget`.
- **Minor (set_recurring_state restore test)**: added a stubbed-client describe block mirroring `review_transactions`: happy-path flip/verify/restore (asserting both EditRecurring calls and their state values) and the restore-on-failed-verification case.
- **Minor (create_recurring label PII)**: the cleanup-registry label is now the synthetic `<marker>-recurring` instead of the server-derived echo name, so cleanup log lines can never carry real merchant data regardless of how the server derives the name.
- **Minor (residue regex false positive)**: documented the approximation in the source comment ("Smoke's BBQ" / "Smoke Shop" false-positive → pre-flight refuses to start, the safe direction for an attended gate) and added tests pinning that behavior.

`bun run check` green (2132 tests). The maintainer's attended `bun run smoke:roundtrip` remains the merge gate.

ignaciohermosillacornejo:
## Live gate completed (attended, 2026-06-12)

**Run 1:** 12 PASS / 1 FAIL / 4 SKIP. `create_recurring` failed with the server's own null-deref error text — the server INSERTED the recurring, then its `emoji`/`icon` resolvers crashed on the category-less recurring, so the mutation reported `USER_ACTION_REQUIRED` while it had applied. The orphan evaded the cleanup registry (never returned to the client) and **the final sweep caught it loudly with its id — the safety design worked exactly as specified.** Residue deleted manually.

**Root cause + fix (commit 26d8883):** our `createRecurring` wrapper only reads `id/name/state/frequency` but requested the app's full `RecurringFields` fragment. New minimal selection avoids the crash-prone server resolvers; mock test pins it. This is a real user-facing bug fix for `create_recurring` on transactions whose recurring derives no category — found by this suite on its first run.

**Run 2 (after fix):** **17/17 PASS — zero marker residue (sweep-verified), B3 response-shape drift counters: none.** All 16 `Mutation.*:applies` ledger surfaces are now live-verified.

Bug Response Ritual:
```text
Root cause:       CreateRecurring selection requested server resolvers (emoji/icon) that crash post-insert on category-less recurrings
Bug class:        mutation selection sets requesting fields the wrapper never reads (over-selection triggers server-side response failures on applied writes)
Detector added:   smoke:roundtrip itself (re-read + residue sweep caught it); regression test pins the minimal selection
Siblings checked: other write wrappers reviewed — editRecurring/editAccount return full fragments but their objects always have the fields resolvable (pre-existing objects); no other wrapper discards most of an app fragment on a CREATE path
Ledger updated:   Mutation.createRecurring:applies now live-verified via smoke:roundtrip (this run)
```

PR #498, opened 2026-06-18
fix(write): resolve account/item via live API so edits aren't limited to the local cache

scobbe:
## Problem

`update_transaction` and `review_transactions` resolve a transaction's `accountId`/`itemId` **exclusively from the local LevelDB cache** (`getAllTransactions()`), and throw `Transaction not found` for any id that isn't currently cached:

```ts
const allTxns = await this.db.getAllTransactions();
const txn = allTxns.find((t) => t.transaction_id === transaction_id);
if (!txn) throw new Error(`Transaction not found: ${transaction_id}`);
```

Because the local cache only auto-fetches ~30 days and Firestore LRU-evicts older docs, this makes it **impossible to edit older transactions** — even with `--write` (which implies `--live-reads`). This directly contradicts the README:

> Write tools resolve transaction metadata (account/item IDs) against the live GraphQL surface so they can edit any transaction the API exposes, not just the ~30 days the local LevelDB cache happens to hold.

Repro: read an older transaction via `get_transactions_live`, then `update_transaction` on it → `Transaction not found`, even though the live API just returned it. In one real account, `decode_health` reported 524 transactions decoded cleanly but only ~462 were resolvable for writes; the ~62 difference were silently un-editable.

## Fix

Adds a shared `resolveTransactionMeta(ids)` helper that prefers the local cache and, for ids not present locally, falls back to a **single live GraphQL window fetch** (via the already-wired `liveDb.getTransactions`). `update_transaction` and `review_transactions` now use it, so they can edit any transaction the live API exposes — matching the documented behavior.

- Recent transactions still resolve from the local cache with **no extra network calls**.
- The live window defaults to the **last 13 months** and is overridable via `COPILOT_WRITE_RESOLVE_WINDOW_MONTHS` for accounts needing deeper history.
- For a bulk `review_transactions` call, the live window is fetched **once** and shared across all ids.

## Testing

- `npm run check` green (typecheck + lint + prettier + 2202 tests).
- Two unit tests that asserted the old "missing account_id or item_id in local cache" message were updated: a locally-incomplete/uncached transaction now attempts live resolution and, when unresolved, throws `Transaction not found` (plural for bulk review).
- Verified end-to-end against a live account: ~70 transactions that previously failed with `Transaction not found` (older than the local cache window) now edit successfully.

## Notes / possible follow-ups

- The same local-only resolution exists in `createRecurring`/`splitTransaction`/`deleteTransaction`; left out of this PR to keep it focused, happy to extend if you'd like.
- Window default (13mo) is a balance between coverage and a cold-start full-window fetch; open to a different default or a smarter per-id bounded fetch.

### DIFF
--- a/src/tools/tools.ts
+++ b/src/tools/tools.ts
@@ -2413,6 +2413,51 @@ export class CopilotMoneyTools {
   }
 
   /**
+   * Resolve accountId/itemId for a set of transaction ids — required to build
+   * EditTransaction mutations. Prefers the local LevelDB cache; for ids not
+   * present (or missing account/item) locally, falls back to a single live
+   * GraphQL window fetch so write tools can edit any transaction the API
+   * exposes, not just the subset the local cache holds. (The local cache only
+   * auto-fetches ~30 days and Firestore LRU-evicts older docs, so cache-only
+   * resolution silently fails for older transactions.) The live window defaults
+   * to the last 13 months and is overridable via the
+   * COPILOT_WRITE_RESOLVE_WINDOW_MONTHS env var. Returns a map of only the ids
+   * that resolved; callers treat absence as "not found".
+   */
+  private async resolveTransactionMeta(
+    ids: string[]
+  ): Promise<Map<string, { accountId: string; itemId: string }>> {
+    const out = new Map<string, { accountId: string; itemId: string }>();
+    const local = await this.db.getAllTransactions();
+    const localById = new Map(local.map((t) => [t.transaction_id, t]));
+    const missing: string[] = [];
+    for (const id of ids) {
+      const t = localById.get(id);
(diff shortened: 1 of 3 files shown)


ignaciohermosillacornejo (approved):
Verified the claims and reviewed the diff. Ran the full check suite locally — typecheck, lint, format, version-sync all green; 2202 tests pass. Also exercised the new live-fallback path against a live session: an older transaction absent from the local cache (where the old cache-only path threw "Transaction not found") now resolves via the live window fetch to the correct account/item. Correct and a real improvement.

Follow-ups (non-blocking, will be handled in upcoming work on the same resolver): add an automated test covering the live-fallback resolution; fill the External assumptions section; consider a more specific error when an id is unresolved but outside the resolution window.

PR #504, opened 2026-06-23
docs(tools): clarify live finance tool descriptions for agent tool-selection

jogan-r:
## What

Improves the `description` text on three live tools so an LLM agent picks and uses them correctly. **Wording only — no behavior, schema, or code-path changes.**

- **`get_categories_live`**: state explicitly that a category's `amount`/`resolvedAmount` is the *budget*, not actual spend, and point to `get_transactions_live` for real spending.
- **`get_transactions_live`**: lead with its primary use — filtering/summing transactions by category, merchant, date, or amount — and call out that a `category` ID (from `get_categories_live`) is how you total category spend. The prior text opened with cache/historical caveats and a list of *unsupported* filters, which buried the fact that this is the "how much did I spend on X" tool. All caveats are retained, just moved after the purpose.
- **`get_monthly_spend_live`**: note it's the all-categories total, not per-category.

## Why

Surfaced while driving this server with a small local model. Twice the model read `get_categories_live`'s `resolvedAmount` (the monthly budget) as money spent and answered the wrong number, because the field is unexplained; and it nearly missed `get_transactions_live` for a category-spend question because the description led with what it *can't* do rather than what it's for. Clearer descriptions fix both without any code change.

## Notes

- Each new description's first sentence leads with purpose, since that's what agents (and embedding/keyword tool-routers) match on.
- No `manifest.json`/`server.json` changes needed — these descriptions aren't mirrored there.

---

_This PR was written by Claude (Claude Code), opened from @jogan-r's account on his behalf. Authorship is also noted via the `Co-Authored-By: Claude` trailer on the commits._


ignaciohermosillacornejo (approved):
LGTM

PR #570, opened 2026-07-23
feat(tools): out-of-window writes via caller-supplied routing ids

heyltsjay:
# Summary

Write tools fail with `Transaction not found` for transactions the local resolution window cannot see — historical rows beyond the window and freshly-posted rows the local snapshot hasn't picked up yet. This PR lets callers holding routing ids from a live read write anyway, removing the local cache as a write dependency.

- **update_transaction**: optional `account_id` + `item_id` (passed together, doc-id validated) are forwarded verbatim and skip local resolution; the not-found error now points at the bypass. Routing ids are excluded from the mutable-field count.
- **review_transactions**: new optional `rows` mode — an array of `{transaction_id, account_id, item_id}` entries supplying the mutation's routing ids directly — as the out-of-window bulk path. `transaction_ids` stays the locally-resolved path and is no longer schema-required; one of the two must be provided (`rows` wins when both are). Both modes normalize to one entries list feeding the existing bounded-concurrency worker loop (cap unchanged at 5).
- **split_transaction**: an unresolvable parent no longer blocks the split when every entry carries an explicit `name` and `date` — no parent-derived defaults are needed then and the amount-sum check is deferred to the server (which enforces it anyway). Resolved parents keep the client-side sum check.
- **add_transaction_to_recurring**: already takes the full caller-supplied triple with no cache lookup, so the bypass is inherent — regression test added, no code change.
- **conformance ledger**: new routing params registered on the `Mutation.editTransaction` entry.

Tests cover bypass dispatch, half-a-pair and malformed-id errors, rows-mode batching (25 rows → 25 mutations), mode precedence, unchanged cache-only paths, split name/date requirements, and the recurring no-lookup regression. `bun run check` is green.

## External assumptions

None new at the GraphQL surface — no new operations or input fields; the added params only change where the routing ids come from. The one behavioral assumption — that `editTransaction`-family mutations accept caller-supplied `accountId`/`itemId` for rows absent from the local cache and validate the binding server-side (wrong pairs fail loudly) — is **live round-trip** verified: this feature has run for weeks as a local patch doing production writes (single updates, bulk review sweeps of historical rows, splits).

### DIFF
--- a/src/conformance/ledger.ts
+++ b/src/conformance/ledger.ts
@@ -380,7 +380,13 @@ export const CONFORMANCE_LEDGER: readonly LedgerEntry[] = [
 
   operation('editTransaction', [
     'update_transaction.transaction_id',
+    'update_transaction.account_id',
+    'update_transaction.item_id',
     'review_transactions.transaction_ids',
+    'review_transactions.rows',
+    'review_transactions.rows[].transaction_id',
+    'review_transactions.rows[].account_id',
+    'review_transactions.rows[].item_id',
   ]),
   gatedInputField('EditTransactionInput.name', ['update_transaction.name']),
   gatedInputField('EditTransactionInput.categoryId', ['update_transaction.category_id']),
--- a/src/conformance/ledger.ts
+++ b/src/conformance/ledger.ts
@@ -409,7 +409,10 @@ export const CONFORMANCE_LEDGER: readonly LedgerEntry[] = [
       'binding, not mere existence — fabricated accountId → "accountId … Not Found"; ' +
       'real-but-wrong pair → "Transaction not found" (server scopes the txn lookup under ' +
       'account/item); correct pair → edit applied. Routing ids therefore cannot be ' +
-      'defaulted or faked; resolveTransactionMeta must supply the true pair.',
+      'defaulted or faked; the true pair must come from resolveTransactionMeta or, on the ' +
+      'out-of-window bypass paths, from the caller (update_transaction account_id/item_id or ' +
+      'review_transactions rows entries, taken from a live read) — a wrong pair fails loudly ' +
+      'either way.',
   },
 
   operation('deleteTransaction', [
@@ -443,6 +446,18 @@ export const CONFORMANCE_LEDGER: readonly LedgerEntry[] = [
   gatedInputField('SplitTransactionInput.categoryId', ['split_transaction.splits[].category_id']),
   responseShape('splitTransaction'),
   appliesSurface('splitTransaction'),
+  {
+    surface: 'Mutation.splitTransaction:sum',
+    kind: 'operation',
+    oracle: null,
+    class: 'verified-once',
+    evidence:
+      'Error-leak recon (docs/graphql-capture/hidden-mutations.md, SplitTransactionInput.amount): ' +
+      '"children must sum to parent.amount (server enforces)". Load-bearing for the ' +
+      'out-of-window split bypass, which skips the client-side sum check and leaves the server ' +
+      'as the sole enforcer on that path. Routine live splits with matching sums succeed ' +
+      '(supporting use only — no wrong-sum rejection transcript has been captured).',
+  },
 
   // ----- Tags ---------------------------------------------------------------
   // No top-level args beyond the input object (covered by CreateTagInput.*).
--- a/src/tools/registry/transactions.ts
+++ b/src/tools/registry/transactions.ts
@@ -332,7 +332,10 @@ export const splitTransactionTool = defineTool({
       'least 2 entries. Each split entry requires `amount` and `category_id`; `name` and ' +
       "`date` default to the parent's values if omitted. The sum of all children's `amount` " +
       "fields must equal the parent's `amount` (server-enforced; this tool also validates " +
-      'client-side before dispatching). After success the parent transaction is hidden but ' +
+      'client-side before dispatching). If the parent cannot be resolved locally (outside the ' +
+      'resolution window), the split can still proceed when every split entry carries an ' +
+      'explicit `name` and `date` — the amount-sum check is then deferred to the server. ' +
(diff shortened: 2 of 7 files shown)


ignaciohermosillacornejo (changes requested):
Thanks @heyltsjay — this is a well-built PR and the direction is right. I validated the claims against the codebase and ran the suite locally (`bun run check` green: 2,351 pass / 0 fail; `sync-manifest` clean — no manifest drift). Verdict: the approach is approved; one in-PR ask, then follow-ups/nits.

## What checks out

- The load-bearing safety property — the server validating the full `(id, accountId, itemId)` binding so a wrong caller-supplied pair fails loudly instead of editing a different row — is already ledger-verified (`Mutation.editTransaction:routing`, 2026-07-05 live probe). That's what makes handing callers the routing ids safe: worst case is a loud error, not corruption.
- The `add_transaction_to_recurring` claim is accurate: no resolution call anywhere in that method, and the poisoned-cache regression test pins it.
- Bypass paths keep all per-field validation (doc-id shape, half-a-pair rejection, split name/date/amount checks) — only the split amount-sum check is deferred.
- Cache patches are no-op-safe for uncached rows; response shapes are backward compatible; making `transaction_ids` optional is non-breaking.
- `rows` mode is the right bulk primitive for out-of-window sweeps — O(1) local state instead of resolving every id.

## One ask before merge

**Register the split sum-enforcement assumption in the conformance ledger.** On the bypass path the client-side sum check is gone, making the server the *only* enforcer — this PR turns a previously redundant assumption into a load-bearing one. Today it's only asserted in `docs/graphql-capture/hidden-mutations.md` ("children must sum to parent.amount (server enforces)") with no ledger entry. Please add a `Mutation.splitTransaction:sum` entry (kind `operation`, class `verified-once`, evidence citing the recon probe and/or your live round-trips). If your weeks of live use included an observed wrong-sum rejection you can cite, even better — that's exactly the evidence the entry should carry.

## Follow-ups (fine outside this PR)

- **`create_recurring` is a missed sibling** (`src/tools/tools.ts` — `createRecurring` resolves routing ids via `resolveTransactionMeta` with no bypass), so "make this old transaction recurring" still fails out-of-window. The same optional `account_id` + `item_id` pair would close the class.
- The `Mutation.editTransaction:routing` ledger evidence ends "resolveTransactionMeta must supply the true pair" — now stale, since the caller may supply it. One-line tweak.

## Nits

- Naming: the title/body say "cache-bypass" / "out-of-cache", but no cache is involved in the live path — this is an out-of-**window** resolution bypass. The registry descriptions get this right ("outside the resolution window"); consider aligning the title.
- `rows: []` silently falls through to the `transaction_ids` path, so the error message talks about `transaction_ids` even when the caller passed `rows`.
- Narrative nit: "freshly-posted rows the local snapshot hasn't picked up yet" mostly applies to degraded (no-liveDb) mode — in live mode the window cache uses `liveTtlMs: 0` for recent months, so fresh rows already resolve.

CI is approved to run. I'll run the local live smoke before merging per repo convention.

heyltsjay:
Thanks for the thorough review — all three in-PR items are addressed in 0078108, `bun run check` green (2352 pass / 0 fail).

**Split-sum ledger entry (the ask):** added `Mutation.splitTransaction:sum` (kind `operation`, class `verified-once`), placed with the other splitTransaction entries. Evidence cites the error-leak recon in `docs/graphql-capture/hidden-mutations.md` ("children must sum to parent.amount (server enforces)"). To be precise about evidence class: our weeks of live use are matching-sum splits only — we have **no captured wrong-sum rejection transcript** — so the entry frames live use as supporting evidence and leans on the recon for the enforcement claim. If you'd rather see a deliberate wrong-sum probe before merge, happy to leave that to your live-smoke pass or a follow-up.

**`rows: []` fall-through:** now errors immediately with a rows-shaped message ("rows must be a non-empty array of {transaction_id, account_id, item_id} when provided — omit it to use transaction_ids instead") instead of falling through; covered by a new test asserting no write dispatches, including the `rows: []` + valid `transaction_ids` case. The neither-provided error is unchanged.

**Naming:** agreed on out-of-window — title updated, and the added strings audited: schema descriptions and errors already said "resolution window"/"out-of-window"; the stragglers were test describes/comments, now renamed. Remaining "cache" mentions in the diff are the degraded-mode paths where the cache genuinely is the resolver.

Also took the one-line follow-up while in the file: `Mutation.editTransaction:routing` evidence no longer implies resolveTransactionMeta is the only legitimate source of the pair (probe facts untouched). `create_recurring` sibling left for a follow-up PR as suggested — glad to send it once this lands.

Fair point on the freshness narrative — in live mode recent months resolve fine (`liveTtlMs: 0`); the fresh-row case really bites in degraded mode. The PR body's framing is yours to discount accordingly.

ignaciohermosillacornejo (approved):
Probe run (attended, synthetic data only — created, split, and deleted a $10 marker transaction): **the server does enforce the sum.**

- Wrong-sum attempt (5 + 4 on a 10 parent) → rejected: `Split amounts (9) must sum to parent amount (10)`
- Matching-sum control (5 + 5) on the same parent → accepted, so the rejection is sum-caused, not a malformed probe
- Cleanup verified: children + parent deleted, no residue

That closes the one open evidence gap — the enforcement your bypass leans on is now an observed rejection, not just recon. I'll upgrade the `Mutation.splitTransaction:sum` evidence to cite the transcript in a small follow-up PR on our side, so nothing further is needed here.

All review items addressed, `bun run check` green locally on 0078108, CI green. Approving — thanks for the quick turnaround and for flagging the missing wrong-sum transcript honestly rather than papering over it. The `create_recurring` sibling is filed as #571; would be glad to have your PR for it once this lands.

ignaciohermosillacornejo:
thanks for working on this! contributions are always much appreciated

PR #577, opened 2026-07-26
fix(check-skills): scan all of src/tools/ and gate on parser health

gmhoward9289-ops:
# Summary

`bun run check:skills` has failed on a clean `main` since the tool-registry
refactor: it reports all 38 tool references across the 6 skills as unknown
tools, every one a false positive. The linter scraped tool names from a
hardcoded `src/tools/tools.ts`, which no longer contains any.

## External assumptions

None — this PR touches only repo tooling. No new assumptions about Copilot's
API, data, enum values, or response shapes.

## Bug fix?

Root cause:       `collect_all_tool_names()` scraped a hardcoded `src/tools/tools.ts`; the schemas moved to `src/tools/registry/*.ts` in 8230a63, so the parse returned one incidental `name: 'name'` literal and the emptiness-only guard let the run continue against a phantom tool list.
Bug class:        Tooling that reconstructs a fact the codebase already knows — here, by text-scraping TypeScript for tool names — and then treats a degenerate reconstruction as a valid answer, failing open and blaming its inputs instead of reporting itself.
Detector added:   `tests/scripts/check-skills.test.ts` drives the script against synthetic repo trees covering every way the tool lookup can fail (missing, non-zero exit, invalid JSON, wrong shape, empty list); each must report a linter fault and validate no skill. A separate test asserts the lookup equals the registry exactly and spans read, write, and live tools — the coverage gap a `manifest.json`-based tripwire would have left.
Siblings checked: `scripts/sync-manifest.ts` — immune, it imports `createToolSchemas` from source so a move fails typecheck loudly rather than silently. `scripts/smoke-graphql.ts` — mentions `tools.ts` only in a comment. No other script locates tool definitions by text-scraping a path.

## What changed

**`scripts/dump-tool-names.ts`** (new)
- Imports `ALL_TOOL_DEFS` — the same registry the server builds dispatch from —
  and prints the 48 tool names as a JSON array.

**`scripts/check-skills.py`**
- Resolve skill tool references against that dump instead of scraping
  TypeScript. `parse_tool_names`, the recursive collector, `manifest_tool_names`
  and `check_parser_health` are deleted: asking the registry removes the failure
  mode rather than guarding it, and gives the linter the same immunity
  `sync-manifest.ts` has — a move that breaks the import fails typecheck loudly.
- Every lookup failure raises `ToolLookupError` and is reported as
  `FAIL: linter self-check: … Skill references were NOT validated.`, naming the
  real cause (bun missing, dump missing, non-zero exit, invalid JSON, wrong
  shape, empty list) rather than a guessed one.
- Keep the `CHECK_SKILLS_REPO_ROOT` override so tests can point the linter at a
  synthetic tree.

**`tests/scripts/check-skills.test.ts`** (new, follows `check-pr-sections.test.ts`)
- Each way the lookup can fail reports a linter fault and blames no skill.
- Failures are clean `FAIL:` output, never a traceback.
- The dump equals the registry exactly, and covers read, write and live tools.
- A genuinely nonexistent tool reference is still rejected.
- The real repository validates clean.

## Verification

- `bun run check` — exit 0, 2381 pass / 20 skip / 0 fail across 135 files
  (2377 before, minus the 6 tests this replaces, plus 10 new)
- `bun run check:skills` — exit 0, `OK: 6 skills validated`
- Mutation-tested rather than asserted — restoring the pre-fix script fails 8 of
  10; making the lookup fail open fails 5 of 10; accepting an empty registry
  fails 1 of 10.

## Note, not addressed here

No CI workflow invokes `check:skills` by name (#578). Merging this does put it
in CI indirectly — the `the real repository` test runs under `bun test`, which
the `tests` job executes on `ubuntu-latest` with python3 preinstalled — but that
is coverage by side effect: it surfaces as a test failure rather than a named
check, and it depends on python3 staying in the runner image. #578 is still
worth doing as an explicit step.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,6 +71,7 @@ scripts/*
 !scripts/check-version-sync.ts
 !scripts/check-server-json.ts
 !scripts/check-skills.py
+!scripts/dump-tool-names.ts
 !scripts/scheduled-smoke.ts
 !scripts/install-scheduled-smoke.sh
 !scripts/uninstall-scheduled-smoke.sh
--- a/scripts/check-skills.py
+++ b/scripts/check-skills.py
@@ -4,16 +4,18 @@
 Checks:
 1. Frontmatter — every skills/<name>/SKILL.md has YAML frontmatter with
    `name:` matching <name> and `description:` non-empty.
-2. MCP tool references resolve against the tool definitions under src/tools/.
+2. MCP tool references resolve against the server's own tool registry.
 3. Profile path is canonical (~/.claude/copilot-money/user-profile.md),
    except for the literal template reference skills/user-profile.template.md.
 
-Before check 2 runs, the parser is validated against manifest.json: every
-tool the manifest declares must be visible to the parser. Without that gate,
-a parser that silently stops seeing tool definitions — because they moved,
-or because the literal format changed — reports every skill reference as an
-unknown tool. That reads as dozens of skill bugs when it is really one
-linter bug.
+Check 2 gets its tool list from `scripts/dump-tool-names.ts`, which imports
+`ALL_TOOL_DEFS` — the same registry the server builds dispatch from. This
+script used to scrape `name: '...'` literals out of TypeScript instead, which
+failed open: when the schemas moved to `src/tools/registry/`, the parse
+collapsed to one incidental literal and every skill reference was reported as
+an unknown tool. That reads as dozens of skill bugs when it is really one
+linter bug. Asking the registry directly removes the failure mode rather than
+guarding it — a move that breaks the import fails typecheck loudly.
 
 Exit 1 on any failure with a clear per-check message.
 """
@@ -23,6 +25,8 @@ from __future__ import annotations
 import json
 import os
 import re
+import shutil
+import subprocess
 import sys
 from pathlib import Path
 
@@ -32,8 +36,7 @@ REPO_ROOT = Path(
     or Path(__file__).resolve().parent.parent
 ).resolve()
 SKILLS_DIR = REPO_ROOT / "skills"
-TOOLS_DIR = REPO_ROOT / "src" / "tools"
-MANIFEST_FILE = REPO_ROOT / "manifest.json"
+DUMP_SCRIPT = REPO_ROOT / "scripts" / "dump-tool-names.ts"
 
 # Prefixes that look like MCP tool names in our skills (kept narrow to
 # avoid matching English words that happen to share a prefix).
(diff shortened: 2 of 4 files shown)


ignaciohermosillacornejo (commented):
Thanks for this — the premise is real and the diagnosis is exact. I checked out the branch and verified the claims independently rather than taking the description at face value.

## Verified

| Claim | Result |
|---|---|
| Broken on clean `main` | ✅ exactly 38 false positives |
| Root cause | ✅ `src/tools/tools.ts:2828` — `name: 'name'` in a GraphQL→MCP field map, the single surviving literal |
| "six weeks" | ✅ `8230a63` is 2026-06-11 |
| `check:skills` → exit 0 | ✅ `OK: 6 skills validated` |
| `bun run check` → 2377 pass / 0 fail | ✅ matches exactly |
| New collector correctness | ✅ finds all 48 real tools + 1 harmless junk entry (`name`, which no `TOOL_PREFIXES` entry matches) |
| Branch state | ✅ current with `origin/main`, `check-pr-sections.sh` passes locally |
| Tests are detectors, not decoration | ✅ reverting the script fails 5 of 6 (see nit 5) |

The bug-response framing is right: hardcoded-path discovery plus a degenerate-parse-is-a-valid-answer guard is exactly the class, and the sibling sweep matches what I found.

## Findings

**1. [Main] The tripwire covers 14 of 48 tools — and misses the 9 write tools the skills actually reference.**

`manifest.json` is the read-only bundle surface: `createToolSchemas()`, 14 cache-mode reads. It contains none of the 17 write tools and none of the 17 `_live` tools. The skills reference 16 distinct tools, and 9 of them are write tools — `update_transaction`, `create_recurring`, `set_recurring_state`, `update_recurring`, `review_transactions`, `create_category`, `update_category`, `delete_category`, `create_tag`.

So if write schemas go dark while the read registry stays parseable, `check_parser_health()` passes clean and the identical false-positive storm recurs for those 9. The docstring acknowledges manifest is "not every tool," but for a change framed as a class-level detector this is the gap worth closing.

The stronger move is to stop scraping altogether. A ~10-line `scripts/dump-tool-names.ts` printing `createToolSchemas() ∪ createWriteToolSchemas() ∪ live` as JSON, invoked via subprocess, deletes `parse_tool_names` entirely and gives the linter the same immunity you correctly credit `sync-manifest.ts` with — a move that breaks the import fails typecheck loudly instead of silently. Bun is already in the loop, since `check:skills` runs under `bun run`.

If you'd rather keep this PR narrow, widening the tripwire to the write tools is the minimum I'd want, since those are the ones the skills lean on hardest.

**2. [Med] A stale `manifest.json` produces a confidently wrong diagnosis.**

`manifest.json` is generated by `bun run sync-manifest` and verified by nothing — it is not in `bun run check` and no workflow runs it. Rename a read tool without regenerating, and the gate fires *"Tool definitions have likely moved or changed shape — fix parse_tool_names/collect_all_tool_names"*, sending the next person after the wrong file. Worth naming stale-manifest as the alternative cause in the message (`…or run bun run sync-manifest`).

**3. [Low] `manifest_tool_names()` doesn't guard malformed JSON.** `json.loads` raises an uncaught `JSONDecodeError` → Python traceback, where every other failure path in this script prints a clean `FAIL:` and returns 1.

**4. [Low] `expect(stderr).toBe('')`** (`tests/scripts/check-skills.test.ts:171`) is brittle — a future python deprecation warning turns this red for an unrelated reason. `expect(stderr).not.toContain('FAIL')` preserves the intent.

**5. [Nit] "Restoring the pre-fix script makes all 6 fail"** — I measured 5 of 6. The survivor is the guard test that *should* pass on both scripts, so the substance holds; the claim just overstates slightly.

## One thing the PR undersells

The new `the real repository > check:skills passes` test runs inside `bun test`, which CI's `tests` job executes on `ubuntu-latest` (python3 preinstalled). So merging this **does** put `check:skills` in CI — which partly resolves #578 as a side effect. Worth saying so in the "Note, not addressed here" section, and it's useful context for how #578 gets scoped.

None of the above is blocking except (1), and even that is a "widen it or replace it" rather than a rewrite. Nice work — the mutation-testing discipline in particular is the right instinct.

gmhoward9289-ops:
Thanks — the manifest gap is real, and the independent verification is appreciated. I took the "replace it" branch rather than widening the tripwire.

## 1 [Main] — replaced the scraping

New `scripts/dump-tool-names.ts` imports `ALL_TOOL_DEFS` and prints the names as JSON; `check-skills.py` shells out to it under bun and resolves skill references against that. `parse_tool_names`, the recursive collector, `manifest_tool_names` and `check_parser_health` are all deleted.

Coverage goes **14 → 48 tools**, so all 9 write tools the skills lean on are in scope, along with the 17 `_live` ones. `ALL_TOOL_DEFS` rather than `createToolSchemas() ∪ createWriteToolSchemas() ∪ live` — same set (it *is* `READ ∪ LIVE ∪ WRITE`), one import, and it's what the server builds dispatch from. A test asserts the dump equals the registry exactly and spans all three families, so the write tools can't quietly drop out again.

This also gets the immunity you credited `sync-manifest.ts` with: a move that breaks the import fails typecheck loudly instead of degrading into a plausible wrong answer.

I kept the linter itself in python rather than porting the whole thing to TypeScript — everything left in it is markdown parsing where the regexes already exist and work, and a rewrite would bury this fix in unrelated diff. Happy to do that separately if you'd rather have one language here. One `.gitignore` entry was needed since `scripts/*` is ignored by default.

## 2 [Med] — dissolved

No manifest read at all now, so a stale `manifest.json` can't produce a confidently wrong diagnosis. Failures name the actual cause: bun not on PATH, dump script missing, non-zero exit (quoting its last stderr line), invalid JSON, wrong JSON shape, or an empty list.

## 3 [Low] — dissolved

No `json.loads` on the manifest. The one remaining decode — the dump's stdout — is wrapped, so `JSONDecodeError` surfaces as `FAIL: linter self-check: dump-tool-names.ts did not print valid JSON (...)`. There's a test asserting the output contains `FAIL:` and no traceback.

## 4 [Low] — fixed

`expect(stderr).toBe('')` → `expect(stderr).not.toContain('FAIL')`, with a comment explaining why.

## 5 [Nit] — you measured right

5 of 6, not 6 of 6. Corrected in the body, and the new numbers are measured rather than asserted — three separate mutations:

| Mutation | Result |
|---|---|
| The original pre-fix scraping script | 8 of 10 fail |
| Fail-open: lookup errors become an empty tool set | 5 of 10 fail |
| Accept a degenerate empty registry | 1 of 10 fails |

The third is deliberately narrow — one behaviour, one test.

## On the CI point

Good catch, and you're right that it partly resolves #578 as a side effect: `the real repository > check:skills passes` runs under `bun test`, which the `tests` job executes on `ubuntu-latest` with python3 preinstalled. Now stated in the "Note, not addressed here" section. It's coverage-by-side-effect rather than a real gate though — it fails inside a test job rather than as a named check, and it dies if the runner image ever drops python3 — so I'd still keep #578 open for an explicit step.

ignaciohermosillacornejo (approved):
LGTM, thanks for working on this.

PR #587, opened 2026-07-27
feat(update_transactions): bulk transaction edit tool

gmhoward9289-ops:
# Summary

`update_transaction` writes one row per MCP tool call, and one tool call costs one agent turn. A cleanup pass over a few hundred transactions therefore spends a few hundred turns — re-reading the whole context each time — to issue a few hundred sub-second mutations. The GraphQL cost was never the bottleneck; the turn count was.

`update_transactions` takes an array of the same edits and applies them in one call. Each entry gets identical validation, field mapping, and optimistic cache patching — arity is the only difference.

## External assumptions

None — reuses `Mutation.editTransaction` and `EditTransactionInput`, both already ledgered. Existing entries' `toolParams` were extended with the new `update_transactions.*` paths rather than adding new surfaces.

Note: Copilot does expose a real server-side `bulkEditTransactions` mutation, but its input shape is unreversed and `docs/graphql-capture/hidden-mutations.md` marks it DO-NOT-PROBE against a live account. This PR deliberately does not use it — client-side fan-out captures the same turn-count win at zero risk. If it is ever adopted it needs its own operation entry, and a comment in the ledger says so.

## What's in it

- **`update_transactions`** — array of edits, max 200 per call, 5 writes in flight.
- **Shared edit plumbing extracted** so the singular and plural paths cannot drift: unknown-key rejection, per-field validation, `EditTransactionInput` mapping, optimistic cache patch. `update_transaction`'s error strings are unchanged (existing tests pin this).
- **Batched routing resolution** — `resolveTransactionMeta` already takes an array and issues at most one windowed fetch, so N edits cost the same lookup as one. Resolving per row would re-fetch the window N times.
- **All validation completes before the first write.** A malformed entry at index 40 fails the call without index 0 having been written — the batch-level extension of the singular tool's atomicity property.
- **`review_transactions` refactored onto the same bounded pool**, now built on the existing `pLimit` util rather than a second hand-rolled cursor loop. Cap stays at 5; its partial-failure contract is unchanged.
- **`continue_on_error`** (default `false`) governs write failures only; `failures[]` reports per-row rejections so callers retry the failures, not the batch. Validation errors are always all-or-nothing.
- **Duplicate `transaction_id` rejected** — the pool runs 5 wide with no ordering guarantee, so two edits to one row would race and the optimistic cache patch would disagree with the server.

## Verification

- 21 new unit tests (`tests/tools/update-transactions-batching.test.ts`): per-edit input isolation, the 5-concurrency cap, validate-before-write, partial-failure accounting in both modes, and cache-patch correctness on partial batches.
- Mutation-tested the headline property: deleting the validation phase makes 3 tests fail.
- Round-trip smoke check added — two throwaway transactions, different edits each, verifying the edits apply independently with no cross-talk.
- `bun run check` green (2403 pass / 0 fail), `bun run check:skills` green.
- Verified over real stdio MCP against the built server: 43 tools, schema correct.

## Skills

`finance-cleanup`, `finance-trip`, and `amazon-sync` now issue one bulk call per approved batch instead of looping per row.


ignaciohermosillacornejo:
I'll review this soon. I'm having some thoughts though on the approach. I think the right track is to use the bulk update endpoint in graphql and enable this on top of it. Only reason to keep using the bulk update via single update endpoint would be if there's some param's that can be edited on the single edit endpoint that can't on the bulk update. Let me run a capture session this weekend and I'll add support for bulk update graphql. You can the rebase this PR and use that instead.

Originally I didn't want to probe much further the bulk update since the UI didn't have the functionality so I couldn't capture a session, and didn't want to risk my own account and mess it up probing the endpoint blindly. That's no longer the case so we can support it properly.

gmhoward9289-ops:
Sounds good. I did a lot of review and scrubbing on it to make it worthy. I want to say I got rid of the harness I had built and moved everythign into graphql, and I was really happy with the result. Happy to bang away more if you have any suggestions. I am using the build without issues, but it deserves a thorough review. Thanks!!

ignaciohermosillacornejo:
## Review + rebase notes

Reviewed this in full, then rebased it onto `main` after #595 landed. Posting both here so the changes I made to this branch are on the record — @gmhoward9289-ops, a chunk of your PR was rewritten and you should see exactly what and why.

**Verdict: high-quality PR, ready to merge.** The findings below are minor and none block.

---

### Original review

I read the whole diff and ran the full gate locally. The claims in the description hold up:

- **The shared plumbing is real, not aspirational.** Validation, field mapping and cache patching were genuinely extracted (`validateEditFields`, `buildEditInput`, `buildEditCachePatch`, `resolveEditRoutes`), and `update_transaction` now runs through them with identical error strings. Nothing was dropped in the refactor — the singular path still validates `transaction_id` format.
- **Validate-before-write is enforced and tested.** Structural checks → one batched routing fetch → per-field validation, all before the first mutation. The tests pin that a bad entry at the end prevents the first write.
- **Batched routing resolution is the right call.** `resolveTransactionMeta` takes an array and issues at most one windowed fetch, so a 200-edit batch costs the same lookup as one. Resolving per row would have re-fetched the window N times.
- **Duplicate `transaction_id` rejection is correct** — with a 5-wide unordered pool, two edits to one row would race and the optimistic cache patch would disagree with the server.
- **Repo rituals followed**: ledger `toolParams` extended, tool counts bumped, CHANGELOG entry, round-trip check, skills updated, `readOnly: false` gating correct, and `manifest.json` rightly untouched since the shipped bundle is read-only.

Supply chain: no dependency, workflow or network-surface changes. It reuses the already-ledgered `editTransaction` mutation. It also explicitly declines to use the then-unreversed `bulkEditTransactions` — which was the right call at the time (see the rebase section).

**Minor findings (non-blocking, all still present):**

1. The `edits` JSON Schema declares neither `minItems: 1` nor `maxItems: 200`. Runtime enforces both, but schema-level bounds would let MCP clients pre-validate instead of round-tripping to an error.
2. `results[]` comes back in completion order, not input order, because entries are pushed from concurrent tasks. Harmless since each carries its `transaction_id`, but sorting would make responses deterministic.
3. In the phase-1 loop, `validateDocId(edit.account_id, 'account_id')` throws without naming the `edits[i]` index, unlike every other batch error. Small inconsistency in an otherwise carefully indexed error surface.

---

### What changed in the rebase

#595 landed first and added `bulk_edit_transactions`, wrapping Copilot's **native** `bulkEditTransactions` mutation — which has since been captured, probed and ledgered (see [`docs/bulk-edit-transactions.md`](../blob/main/docs/bulk-edit-transactions.md)). That created one genuine semantic collision with this PR plus some mechanical conflicts. Resolutions, all judgement calls I made on your behalf:

- **`review_transactions` keeps the native bulk path.** This PR refactored it onto the new `runBoundedPool` fan-out (N requests at 5-wide); `main` now issues **one** `bulkEditTransactions` request for the whole set. Main's version wins — it is strictly fewer round trips. `runBoundedPool` survives and is used only by `update_transactions`, where per-row heterogeneous edits genuinely need it.
- **Ledger**: `editTransaction` keeps `update_transaction.*` + `update_transactions.*`; `review_transactions.*` moved onto the `bulkEditTransactions` entry. I also rewrote the comment this PR added claiming `bulkEditTransactions` is *"unreversed and unsafe to probe against a live account"* — it is now captured, ledgered and adopted.
- **CHANGELOG**: both entries kept. Corrected a wrong PR link, and removed the now-false *"`review_transactions` now shares the same bounded write pool"* claim.
- **Counts**: 18 → 19 write tools, 33 base.

### One defect fixed

The tool description told the model to use `update_transactions` for *"recategorizing a merchant across months"* and *"retagging a trip"*. Those are now precisely the cases `bulk_edit_transactions` does in a single request — and the second is actively harmful: `update_transactions`' `tag_ids` **replaces** a row's whole tag list, so following that advice to "add a trip tag" silently drops any tags the row already had. `bulk_edit_transactions.add_tag_ids` is additive.

Both descriptions now open with the contrast and cross-reference each other, and a ratchet test in `tests/unit/write-tool-schemas.test.ts` pins the cross-references plus the tag hazard. Mutation-tested: removing the warning fails the test.

The division of labour between the two tools is now:

| | `update_transactions` (this PR) | `bulk_edit_transactions` (#595) |
|---|---|---|
| Shape | **N different edits**, one per row | **1 edit** applied to N rows |
| Fields | all 8, incl. `name`/`date`/`amount`/`note` | the 5 Copilot's schema allows |
| Tags | `tag_ids` **replaces** the list | `add`/`remove` — cannot replace |
| Requests | N, at 5 in flight | **1** |

They only overlap when every edit is identical *and* touches only the five bulk-supported fields. Outside that, each does something the other structurally cannot — which is why both are shipping.

### Verification

Run locally against the rebased branch:

- `bun run check` — **2435 pass / 0 fail**
- `bun run check:skills` — `OK: 6 skills validated`
- `bun run smoke:conformance` — PASS
- `bun run smoke:roundtrip` — **19/19 PASS**, zero marker residue, no response-shape drift

That last one matters: when I first reviewed this PR I flagged that its round-trip and stdio verification claims needed a live account to confirm and that I'd want them exercised before merge. They now are — `update_transactions` passes its round-trip against the real endpoint.

Note the automated `review` workflow reported **skipping** on this PR (fork head), so this is the only review on it.

### Follow-ups, not blockers

- The three minor findings above.
- A design pass concluded we should **not** add silent homogeneous-batch routing inside `update_transactions` (switching transport based on runtime value-equality would silently change the error contract, and full observational equivalence is unachievable because a silently-skipped row has no server error text). Steering via descriptions is the chosen approach; a response `hint` when a batch *would* have been bulk-eligible is the proposed measurement before revisiting.
- Pre-existing doc drift unrelated to this PR: `skills/amazon-sync/SKILL.md` still says `update_transaction` can't set `reviewed` (it can), and `CLAUDE.md` still describes the retired `WRITE_TOOLS` set in `server.ts`.

ignaciohermosillacornejo:
## Second review — independent pass (different model), findings addressed

Because the automated `review` workflow skips on fork heads, and because I'd rewritten a third of this branch during the rebase and then verified my own work, I had an independent reviewer go over it — with instructions to scrutinise **my** rebase commit hardest and to mutation-test rather than eyeball.

**Its verdict: approve-with-nits.** It found five issues, **two of them mine**. All five are now fixed in `5509b2d`.

---

### 1. MED — stop-on-first-failure had no detector at all

The most valuable finding, and it was invisible to every other signal we had.

`continue_on_error: false` promises the batch stops at the first failure. The reviewer deleted **both** `stopOnError` guards in `runBoundedPool` and **all 2435 tests still passed**. The only default-mode test asserted:

```
/update_transactions failed at transaction_id=txn-05 \(\d+\/8 succeeded\)/
```

`\d+` matches whether the batch stopped after 4 writes or ran all 8. The stop property was never asserted.

**Failure scenario:** someone refactors the pool and drops the guard. A 200-edit batch with `continue_on_error: false` fails at row 3 — and the remaining 197 writes still go out against real financial data. The thrown message ("failed at X, N/200 succeeded") is indistinguishable from correct stopped-early behaviour. Silent blast-radius regression.

Note this was **not** caught by coverage: codecov reported 98.7% patch coverage on the original PR, and both guards were *executed* by tests. Line coverage answers "did it run", not "would anything notice if it were wrong".

Fixed by pinning the actual property — 20 edits, 5-wide pool, entry 0 fails fast, assert the tail was never attempted:

```
const leaked = attempted.filter((id) => tail.includes(id));
expect(leaked).toEqual([]);
expect(attempted.length).toBeLessThanOrEqual(5);
```

Mutation-checked: deleting the guards now fails this test.

### 2. LOW — my ratchet assertion was vacuous, and my commit message overclaimed

In the previous commit I added a steering ratchet and wrote that it was "mutation-tested". I had verified the tag-hazard assertions — those are genuine detectors — and stated that of the whole test. One assertion wasn't:

```
/name.*note.*date.*amount|name, note, date or amount/
```

That matched the description's field *enumeration* ("transaction_id plus any of name, category_id, note, tag_ids, type, reviewed, date, amount"), not the steering sentence it was meant to pin. Deleting the entire "Copilot cannot bulk-edit those four" sentence left it green.

Re-anchored on `/Copilot cannot bulk-edit/` and verified by deleting that sentence.

### 3. LOW — stale comments left by my conflict resolution

`runBoundedPool`'s contract comment still said it was "shared by review_transactions and update_transactions" and that it "Preserves the pre-existing review_transactions behaviour exactly". After the rebase, `review_transactions` uses the native `bulkEditTransactions` single request and never touches the pool. Exactly the half-merged residue the review was asked to hunt for. Corrected.

### 4. LOW-MED — `results[]` claim was false in default mode

The tool description and doc comment said `updated_count` and `results[]` "reflect exactly what was written" either way, so "retry the failures". In default mode the call **throws**, so the client receives only an error string and its counts — no `results[]`. And under a 5-wide pool the succeeded set is completion-order-dependent, not an index prefix, so "retry the failures" wasn't actionable.

Corrected to describe both modes accurately, and to note that every edit is an absolute assignment, so re-running the whole batch is safe (which is why `idempotentHint: true` is right).

### 5. LOW — lockstep tables not extended

`tests/unit/server-protocol.test.ts` says, two lines above the count it pins, that the annotation and rejection tables "must be extended in lockstep". `update_transactions` was absent from `ALL_WRITE_TOOLS` and `WRITE_TOOL_SPECS`; `bulk_edit_transactions` was absent from `ALL_WRITE_TOOLS` — **my debt from #595**. `ALL_WRITE_TOOLS` is asserted with `arrayContaining`, so both omissions passed silently. Both added.

Gating was never actually broken — it's registry-derived and `tool-registry.test.ts` pins `readOnly` classification generically — but the per-tool belts had been quietly narrowing.

---

### What the independent pass confirmed

Worth recording, since these are the parts where my own verification was least trustworthy:

- **The conflict resolution lost nothing.** It recovered the contributor's true pre-rebase head (`8368c82`, from the `head_ref_force_pushed` timeline event) and diffed it against HEAD: the only `tools.ts` differences are main's #595 content. `updateTransactions` and all five extracted helpers came through byte-identical. No dangling code.
- **Ledger placements match `origin/main` exactly**, and `tests/conformance/ledger.test.ts` is genuinely bidirectional — a missing param fails one test, a stale one fails another. Not vacuous.
- Validation phase ordering (structural → routing → per-field → write) is real and tested with zero-call assertions; `pLimit` is FIFO and releases slots on rejection; category/tag validators are TTL-cached so a 200-row batch doesn't fan out validation fetches.

### Verification after fixes

`bun run check` — **2437 pass / 0 fail** · `check:skills` OK · `smoke:conformance` PASS · `smoke:roundtrip` **19/19** against the live endpoint, zero marker residue.

### Remaining nits, non-blocking

- `edits` schema still lacks `minItems: 1` / `maxItems: 200` (runtime-enforced only).
- `results[]` order is completion-order, not input-order, and undocumented.
- `validateDocId(edit.account_id, …)` doesn't name the `edits[i]` index, unlike every other batch error.

ignaciohermosillacornejo (approved):
Approving on the strength of: an independent review by a different model (findings posted above, all five addressed in 5509b2d), the round-trip smoke passing 19/19 against the live endpoint, and `bun run check` green at 2459 tests.

Disclosure: I authored the rebase resolution and the subsequent fix commits on this branch, so this approval is not independent of the changes it covers. The independent signal is the second review — which found two real defects in my own commit, including a vacuous test assertion I had incorrectly described as mutation-tested.

Follow-up filed as #596: a mutation-guard registry, so the class of bug found here (an assertion satisfied by both correct and broken behaviour) has a standing detector rather than depending on a reviewer thinking to check.

ignaciohermosillacornejo:
Ended up merging, as there's product reasons to keep this fan out approach. Bulk edit can't do everything this approach has, so there's value in having both.

Issue #521, opened 2026-07-06
Live caches retain the previous login's data across mid-session re-auth (all-caches uid-transition sweep)


ignaciohermosillacornejo:
Found by #511's final review. When the session re-authenticates as a DIFFERENT account mid-process (token refresh failure → cold re-extract picking up another browser login), the in-memory live caches retain the previous login's data: the window cache and snapshot caches hold the prior account's full nodes (categories, tags, accounts, transactions), and served reads/validations can mix identities until TTLs expire.

#511 fixed this class for the meta index (memory tier cleared on non-null→different-non-null uid transition, per-uid files on disk). The remaining sweep: clear/invalidate ALL live caches (transactionsWindowCache, accountsCache, categoriesCache, tagsCache, recurringCache, upcomingRecurringsCache, userCache, holdingsCache, balance-history/investment caches) on the same transition — likely one guard at a shared chokepoint rather than per-cache patches. Note the pre-existing severity: full nodes are strictly more sensitive than routing ids.

Also absorb here if convenient: store-side line-skip warnings and valve hardening polish beyond what #511 shipped.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #470, opened 2026-06-11
E1: tool registry — one definition per tool drives schema, dispatch, gating, and manifest

ignaciohermosillacornejo:
# Summary

Implements the E1 tool registry (Epic E #431): every MCP tool is now a single `ToolDefinition` — `{ name, schema, handler, readOnly, requiresLiveReads?, swappedOutInLiveMode?, formatError? }` — in `src/tools/registry/`. The server's tool list, dispatch, write gating, live-reads gating, and live-mode swapping are all derived from the registry; `createToolSchemas()` / `createWriteToolSchemas()` (and through them `sync-manifest` and the conformance-ledger walk) are pure projections of the same definitions. The legacy `server.ts` switch, the `WRITE_TOOLS` parallel set, and the hardcoded live-swap name list are gone. Closes #446

All 44 tools migrated, one domain per commit: transactions → categories/tags → recurring → budgets/goals → investments → accounts/system → live (13 GraphQL-backed tools).

## Behavior preservation

The MCP-visible tool list (names, descriptions, inputSchemas, annotations) is **byte-identical** to `origin/main` under all four flag combinations (default, `--write`, `--live-reads`, `--write --live-reads`): `handleListTools()` was dumped on both revisions with the same script and `diff` reports zero differences (3,658-line JSON dump). Dispatch semantics are locked by the existing e2e suite (write gating, per-tool live-reads errors, `refresh_cache` verbatim error text via `formatError`) plus a new `tests/unit/tool-registry.test.ts` asserting registry invariants (unique names, `name === schema.name`, `readOnly` ⇔ `readOnlyHint`, exact live/swap classification, list sizes 14/13/17).

Gates: full `bun run check` green (typecheck, lint, format, 2,100 tests incl. doc-sync, manifest-sync, ledger walk, read-smoke coverage, mcpb-bundle); `bun run sync-manifest` reports zero diff.

## Notable internals

- `CopilotMoneyServer` shrinks 671 → 319 lines; the 13 private live-tool fields collapse into one `live?: LiveToolContext`.
- `tools.ts` shrinks ~5,100 → 3,891 lines; all schema literals now live in per-domain registry modules (live schemas stay with their `tools/live/*` modules, wrapped by `registry/live.ts`).
- `normalizeMerchantName` moved to leaf `src/utils/merchant.ts` (re-exported from `tools.ts`) to keep the import graph acyclic.
- `get_balance_history` remains deliberately un-swapped in live mode; the rationale comment moved onto its definition.

## External assumptions

None

[...]

PR #588, opened 2026-07-31
test: add registry-derived tool-count drift check

gmhoward9289-ops:
## Summary
- Adds `scripts/check-tool-counts.ts`, which derives real tool counts from `READ_TOOL_DEFS`/`LIVE_TOOL_DEFS`/`WRITE_TOOL_DEFS` in `src/tools/registry/index.ts` and asserts that `package.json`, `README.md`, `CLAUDE.md`, `docs/index.html`, and `docs/graphql-live-reads.md` all restate the same counts.
- Wires it into `bun run check` as `check:tool-counts` **and into `.github/workflows/test.yml` as an explicit CI step**, so fork PRs (which bypass the local pre-push hook) are covered.
- Adds `tests/scripts/check-tool-counts.test.ts` following the `check-skills` sibling pattern: a `CHECK_TOOL_COUNTS_ROOT` override drives the script against synthetic doc trees (pass, corrupted count, multiple mismatches all reported, unreadable file reported cleanly).
- Motivation: the doc-facing tool-count description in `package.json` went stale twice in a row (49f267d, then a wrong-split follow-up in 1038ec7), and `docs/index.html` / `docs/graphql-live-reads.md` drifted separately (fixed in b22bacc). Five prose surfaces restate registry counts with nothing enforcing agreement — this check closes that gap so future registry changes (or copy edits) fail CI instead of silently shipping wrong numbers.

## Scope note (review response)
- The earlier `fail_ci_if_error: true → false` Codecov change has been **reverted out of this PR**; it will be proposed separately with its own rationale so it gets deliberate review.
- Remaining doc commits (README/CONTRIBUTING/CLAUDE.md rewrites, SECURITY.md version table, audit-plan archive) stay: the CLAUDE.md/CONTRIBUTING.md edits are entangled with the substrings this check asserts, and the rest were spot-checked in review as individually fine. Noted for next time — tighter PRs.
- Hardcoded `src/server.ts:154`/`:181` line references in CLAUDE.md/CONTRIBUTING.md replaced with function-name references per review.

## External assumptions

None — this PR adds a CI check over in-repo files only; it makes no new assumptions about Copilot's API, data shapes, or enum values.

## Test plan
- [x] `bun run check:tool-counts` passes against current registry state
- [x] New unit tests: 5 pass (pass case, corrupted count, all-mismatches, unreadable file)
- [x] `bun run check` (typecheck, lint, format, version-sync, server-json, tool-counts, full suite) green: 2407 pass, 0 fail
- [x] Workflow diff vs main is now only the added `check:tool-counts` step


scripts/check-tool-counts.ts
```diff
@@ -0,0 +1,141 @@
...
+ * Assert that every doc-facing tool count matches the registry.
+ *
+ * The registry (`src/tools/registry/index.ts`) is the single source of truth
+ * for tool counts, but several human-written surfaces (package.json's npm
+ * description, README.md, CLAUDE.md, docs/index.html, docs/graphql-live-reads.md)
+ * restate those counts in prose. Nothing enforced agreement, so a registry
+ * change (or a copy edit that "helpfully" rewords a count) went stale twice in
+ * a row (commits 49f267d and 1038ec7 shipped two different wrong splits of
+ * the same description string before b22bacc caught the other two files).
+ * Run as part of `bun run check`.
```
ignaciohermosillacornejo, line 12: **This never runs in CI, so the PR's stated goal ("fail CI instead of silently shipping wrong numbers") isn't met yet.** `bun run check` only runs via the local `.husky/pre-push` hook — fork PRs (like this one) bypass it entirely. `.github/workflows/test.yml` runs each quality gate as an explicit step (`check:version-sync` and `check:server-json` both have one) but no `check:tool-counts` step was added, even though this PR touches that workflow. Please add:

```yaml
      - name: Tool count drift check
        run: bun run check:tool-counts
```

Separately, a minor gap: the repo has an established pattern for unit-testing check scripts against synthetic inputs (`tests/scripts/check-skills.test.ts`, and #586's `check-deps-pinned` tests via an env-var override), and this script has none — its only exercise is the live repo state. Low priority given how simple the script is, but worth noting it breaks the sibling pattern.

---

.github/workflows/test.yml
```diff
@@ -84,4 +84,7 @@ jobs:
         with:
           files: coverage/lcov.info
           token: ${{ secrets.CODECOV_TOKEN }}
-          fail_ci_if_error: true
+          # Coverage upload is best-effort: a Codecov outage or missing token
+          # must not fail an otherwise-green build. Set back to true once a
+          # CODECOV_TOKEN secret is configured, if a hard gate is wanted.
+          fail_ci_if_error: false
```
ignaciohermosillacornejo: This flips a CI gate from hard-fail to best-effort, but it's not mentioned anywhere in the PR title, summary, or test plan — a reviewer reading the description would never know a gate got loosened. The inline rationale (a missing `CODECOV_TOKEN` or Codecov outage shouldn't fail a green build) is reasonable, but please either disclose it in the PR body or split it into its own PR so it gets deliberate review rather than riding along in a "test:" PR about tool counts.

---

ignaciohermosillacornejo (changes requested):
The core detector is well-built and I verified it hands-on in a worktree — but it needs the CI wiring to deliver what the PR promises, and one undisclosed CI change should be surfaced. Details in the inline comments; summary below.

## Verified

- `bun run check:tool-counts` passes on the PR head; derived counts (14 read, 17 live, 17 write, 6 swapped → 25 live-mode, 31 base, 48 total) match the registry.
- Mutation test: corrupting a count in `package.json` fails the check with a clear error naming the surface and printing the derived truth. It reports **all** mismatches, not just the first — nice.
- Typecheck (script is covered by `tsconfig.scripts.json`), lint, format, `check:version-sync`, `check:server-json` all pass. `.gitignore` whitelist handled.
- Full suite: 2401 pass / 1 fail — the failure (`transaction-meta-store.test.ts`, the #521 cap-valve suppression test) **also fails on main** in my environment, so it's pre-existing, not from this PR.

## What's good

- Correctly follows the fix-the-class ritual: three stale-count incidents cited, and the detector covers all five prose surfaces at once with no hardcoded expected numbers — `survivingCache`/`liveModeTotal`/`baseTotal` are all derived from the registry, mirroring CLAUDE.md's counting convention.
- Exact-substring matching is brittle by design, but the failure message explicitly invites updating the script when wording legitimately changes — the right trade-off for prose assertions.

## Blocking

1. **`check:tool-counts` doesn't run in CI** — only in the local pre-push hook, which fork PRs bypass. One workflow step fixes it (inline comment).
2. **Undisclosed Codecov gate weakening** (`fail_ci_if_error: true` → `false`) — disclose in the PR body or split out (inline comment).

## Non-blocking

- **Scope creep:** ~half the 10 commits are unrelated to tool counts (docs/README.md rebuild, `REVIEW_AUDIT_PLAN.md` archive, SECURITY.md version table, the Codecov change). I spot-checked them and they're individually fine — no stale links to the moved audit plan, the registry description matches what shipped in #470, the SECURITY.md claim matches the changelog — but "External assumptions: None — this PR adds a CI check over in-repo files only" undersells the diff. The CLAUDE.md/CONTRIBUTING.md rewrites are legitimately entangled (the check asserts substrings in CLAUDE.md); the rest ideally travels separately next time.
- "48 tools" on the landing page counts all definitions, but no single mode exposes 48 (write mode maxes out at 42, since swapped cache tools and their `_live` replacements are both counted). Now drift-proof, which is the point, but the number slightly overstates the surface.
- Hardcoded line references in the docs (`src/server.ts:154`, `:181`) will silently rot; function-name references would age better.
- No unit test for the new script, unlike its `check-skills` / `check-deps-pinned` siblings (inline comment).

---

scripts/check-tool-counts.ts
gmhoward9289-ops: Done in 9daec5d + 39e13f8: added the `Tool count drift check` workflow step next to the other quality gates, and added `tests/scripts/check-tool-counts.test.ts` on the `check-skills` sibling pattern (new `CHECK_TOOL_COUNTS_ROOT` override, synthetic doc trees covering pass / corrupted count / all-mismatches-reported / unreadable-file cases).

.github/workflows/test.yml
gmhoward9289-ops: Reverted in 9daec5d — the workflow diff vs main is now only the added `check:tool-counts` step. The `fail_ci_if_error` loosening will be proposed as its own PR with the rationale spelled out.

ignaciohermosillacornejo:
Confirmed both blockers are resolved — verified in the diff, not just the replies: the `Tool count drift check` step is in `test.yml` next to the other quality gates (so CI wiring needs no separate PR), the codecov loosening is fully reverted out of this diff (it got its own review over in #591), and the synthetic-tree unit tests for the script look good. Nice work — once this merges, your detector becomes one of the repo's standing CI gates.

Last step: a rebase onto main (this repo is rebase-merge only). Heads-up that it's more than mechanical — main has gained two write tools since your branch point (`bulk_edit_transactions` from #595 and `update_transactions`), so the derived counts your docs commits assert (17 write / 31 base / 48 total) are now stale against main. Expect `check:tool-counts` to fail right after the rebase until the prose numbers are re-derived — your own detector proving its worth on its own PR.

If you'd rather not chase the re-derivation, just say so and we'll handle the rebase ourselves, keeping your commits and authorship intact. Otherwise, once it's green this is ready to approve.

ignaciohermosillacornejo (approved):
Re-reviewed after the update — everything checks out, verified hands-on in a worktree: `check:tool-counts` derives 14 read / 17 live / 19 write / 33 base / 50 total and all five doc surfaces agree; the codecov revert survived the rebase (zero `fail_ci_if_error` in the diff); your new CI step ran and passed on this very PR; full `bun run check` green including the context-budget ratchet that landed since your branch point; and the line-reference nit is fixed nicely with function-name references. The branch is a few commits behind main but merges clean — no re-rebase needed. Thanks for sticking with this one through the split-out and re-derivation; from here on, stale tool counts are a CI failure instead of a doc bug.

PR #590, opened 2026-08-01
Exclude base64 account logos from get_accounts by default

### LINKED ISSUE #589: get_accounts embeds full base64 institution logos, bloating responses ~20-30x
## Summary

`get_accounts` embeds each account's institution `logo` as a full base64-encoded PNG (`logo` + `logo_content_type` fields), straight from the cached Firestore document, with no way to opt out. For a normal set of linked accounts this dominates the response size — in my case, 15 accounts came back as ~66KB of JSON, and grep'ing the response showed the actual financial data (balances, names, masks, institution names) was only ~2-3KB of that. The rest was inline PNG bytes.

For an MCP client (an LLM) this is pure waste: the logo image isn't something a text-based tool caller can use, and it can blow past per-tool-call token/size limits well before the useful data does (I hit the MCP client's own response-size cap on a `get_accounts` call with only 15 accounts).

## Steps to reproduce

1. Run the server in default (cache) mode with a Copilot Money install that has several linked accounts.
2. Call `get_accounts`.
3. Inspect the response — each account object contains a multi-KB base64 `logo` string.

## Suggested fix

Exclude `logo` / `logo_content_type` from the `get_accounts` response by default, with an `include_logos: false`-by-default parameter to opt back in for callers that actually want the image data. This mirrors the existing `include_hidden` pattern the tool already has.

Happy to open a PR for this if useful — I have a working patch (strip fields in `CopilotMoneyTools.getAccounts`, add `include_logos` to the tool's input schema, plus tests).

NikhilGaddam:
## Summary

Fixes #589 — `get_accounts` was embedding each account's cached `logo` (a base64-encoded PNG, several KB per account) inline on every call, with no way to opt out. For a normal set of linked accounts this dominated the response: in the reporter's case, 15 accounts came back as ~66KB of JSON, of which only ~2-3KB was actual financial data (balances, names, masks, institution names).

For an MCP client this is pure waste — the image bytes aren't usable by a text-based tool caller and can trip response-size limits well before the useful data does.

## Changes

- `CopilotMoneyTools.getAccounts` now strips `logo` / `logo_content_type` from each account by default.
- Added `include_logos` (default `false`) to `get_accounts`'s input schema, mirroring the existing `include_hidden` pattern, for callers that do want the image data.
- Updated the tool description and `docs/tools-by-mode.md` to document the new default.
- Added unit tests covering both the default (stripped) and `include_logos: true` (preserved) behavior.
- Added a CHANGELOG entry.

## Test plan

- [x] `bun run typecheck` — passes
- [x] `bun run lint` — passes (0 errors; pre-existing unrelated warnings only)
- [x] `bun run format:check` — passes
- [x] `bun run check:version-sync` / `check:server-json` — pass
- [x] `bun test` (full suite via pre-push hook) — 2384 pass, 0 fail
- [x] Manually verified against a real local Copilot Money database: `get_accounts` response dropped from ~66KB to well under the previous size with `logo`/`logo_content_type` absent, and reappeared correctly with `include_logos: true`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`get_accounts` no longer embeds base64 institution logos by default.** Each account's cached `logo` (a base64-encoded PNG) was being returned inline on every call, bloating responses roughly 20-30x for no benefit to an MCP client. Logos are now stripped by default; pass `include_logos: true` to opt back in. Fixes [#589](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/issues/589).
+
 ## [2.2.2] - 2026-06-14
 
 ### Fixed
--- a/docs/tools-by-mode.md
+++ b/docs/tools-by-mode.md
@@ -7,7 +7,7 @@ This server exposes different tools depending on which CLI flags you enable. The
 | Tool | Status | Notes |
 |---|---|---|
 | `get_transactions` | ✅ | Query transactions with filters (date range, category, merchant, amount, account, text search, etc.) |
-| `get_accounts` | ✅ | List accounts with balances; filter by type |
+| `get_accounts` | ✅ | List accounts with balances; filter by type. Logo images are omitted by default (`include_logos: true` to include) |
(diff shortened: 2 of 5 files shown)


ignaciohermosillacornejo (approved):
LGTM, thanks for catching this, as I had not used the get accounts surface in my personal use.

PR #591, opened 2026-08-01
ci: make Codecov upload best-effort (fail_ci_if_error: false)

gmhoward9289-ops:
## Summary
One-line CI change: set `fail_ci_if_error: false` on the Codecov upload step.

Today a Codecov outage or a missing `CODECOV_TOKEN` secret fails an otherwise-green build. Coverage feedback on PRs still arrives via the `codecov/patch` status check; the upload step itself shouldn't be a hard gate. A comment in the workflow records the rationale and how to restore the hard gate.

Split out of #588 per review, so the gate loosening gets deliberate review instead of riding along in an unrelated PR.

## External assumptions

None — CI configuration only; no assumptions about Copilot's API or data.

## Test plan
- [x] Pre-push `bun run check` green (2404 pass, 0 fail)
- [ ] CI on this PR runs the upload step and stays green

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -84,4 +84,8 @@ jobs:
         with:
           files: coverage/lcov.info
           token: ${{ secrets.CODECOV_TOKEN }}
-          fail_ci_if_error: true
+          # Coverage upload is best-effort: a Codecov outage or a missing
+          # CODECOV_TOKEN secret must not fail an otherwise-green build. The
+          # codecov/patch status check still reports coverage on PRs; set back
+          # to true if a hard upload gate is ever wanted.
+          fail_ci_if_error: false


ignaciohermosillacornejo:
Thanks for the work here @gmhoward9289-ops — and apologies, this one's on us. The ask to split this out came from review feedback on #588, and on closer look it wasn't a good request. Before deciding we went through the CI history: zero codecov-upload failures in ~7 months of retrievable runs (the one 2026-06-08 infra 504 also failed a job that has no codecov step), and the historical token-missing cases were already solved surgically by skipping the upload for fork/dependabot PRs. So the hard `fail_ci_if_error: true` gate has had zero cost — and it's currently the only signal that would surface a broken upload or dead token, since the codecov status contexts aren't required checks. Making the upload best-effort would trade a flake that hasn't happened for a silent-failure mode that would go unnoticed indefinitely.

Closing rather than asking you to rework something we shouldn't have requested in the first place. If Codecov ever starts genuinely flaking we'll revisit — most likely as `continue-on-error` scoped to PR runs with the hard fail kept on pushes to main, so failures stay visible. Sorry for the churn, and thanks for the careful split-out.

PR #593, opened 2026-08-01
Add field selection (fields/compact) to get_transactions

### LINKED ISSUE #592: get_transactions has no field selection — full ~35-40 field document per row is mostly unused bloat
## Summary

`get_transactions` returns the full Firestore transaction document per row — ~35-40 fields including internal IDs (`item_id`, `user_id`, `plaid_category_id`, `parent_transaction_id`), Plaid metadata (`plaid_category_strings`, `plaid_transaction_type`, `plaid_deleted`), intelligence-suggestion arrays (`intelligence_suggested_category_ids`), and assorted flags (`is_amazon`, `from_investment`, `account_dashboard_active`, `_origin`, etc.). For most spending-analysis use cases only a handful of these matter (transaction_id, date, name, amount, category).

There's currently no way to ask for less. Pulling a few months of transaction history multiplies this bloat across hundreds/thousands of rows — a 6-month pull can come back as 600K+ characters, most of it fields the caller never uses, forcing a save-to-file + grep-and-reconstruct workflow just to get clean rows.

## Suggested fix

Add field selection to `get_transactions`:
- `fields: string[]` — return only the named fields per transaction (an explicit allowlist, like a GraphQL/SQL field selection).
- `compact: true` — a curated 7-field preset (`transaction_id`, `date`, `name`, `amount`, `category_name`, `account_id`, `pending`) for the common case, without having to spell out `fields` every time. `fields` takes priority when both are given.

Default behavior (neither option passed) stays exactly as-is — fully backward compatible.

Happy to open a PR — I have a working patch (projection applied post category_name/normalized_merchant enrichment, covering both the filtered-list and single-transaction-by-ID lookup paths) plus tests.

NikhilGaddam:
## Summary

Fixes #592 — `get_transactions` returns the full ~35-40 field Firestore document per transaction (internal IDs, Plaid metadata, intelligence-suggestion arrays, misc. flags), most of which go unused for typical spending-analysis queries. Pulling a few months of history multiplies this bloat across hundreds/thousands of rows — the reporter hit 600K+ characters for a 6-month pull, forcing a save-to-file-and-grep workflow just to reconstruct clean rows.

## External assumptions

None — this is pure post-processing of already-cached data (no new Copilot API/GraphQL surface touched).

## Changes

- Added `fields: string[]` to `get_transactions` — an explicit field allowlist, applied per-transaction after `category_name`/`normalized_merchant` enrichment (so both are selectable alongside raw document fields).
- Added `compact: true` — a curated 7-field preset (`transaction_id`, `date`, `name`, `amount`, `category_name`, `account_id`, `pending`) for the common case without spelling out `fields` every time. `fields` wins when both are given.
- Applied to both the filtered-list path and the single-transaction-by-ID lookup path, so behavior is consistent across `get_transactions`'s modes.
- Default behavior (neither option passed) is unchanged — fully backward compatible.
- Updated the tool description, `docs/tools-by-mode.md`, and added a CHANGELOG entry.
- Added unit tests: default full-width response, `compact: true`, `fields: [...]` (including an unknown field name, which is silently ignored), `fields` taking priority over `compact`, and the single-lookup path honoring `compact`.
- Rebased onto latest `main` and raised `SCHEMA_BUDGETS.get_transactions` (4,145 → 5,100 chars, measured value + ~10% headroom per the ratchet's own convention) in `tests/context-budget.test.ts` — this PR's new `fields`/`compact` schema properties are a deliberate capability addition, not bloat, per the ratchet's "raising a budget requires justifying the increase in the PR that needs it" policy.

Going with Option A from the review — smallest lift to land what's already reviewed, and I'd like to see the primitive used for a bit before committing to the bigger `"default"`/`"all"`/`_field_warning` v3 shape. Happy to help scope that out in #597 whenever you're ready.

## Test plan

- [x] `bun run typecheck` — passes
- [x] `bun run lint` — passes (0 errors; pre-existing unrelated warnings only)
- [x] `bun run format:check` — passes
- [x] `bun run check:version-sync` / `check:server-json` — pass
- [x] `bun run check` (full suite via pre-push hook, post-rebase) — 2523 pass, 0 fail
- [x] Manually verified against a real local Copilot Money database with `compact: true` and a custom `fields` list, confirming the returned rows only carry the requested keys


ignaciohermosillacornejo:
Thanks for #592 and this PR — the problem statement is exactly right, the implementation is clean (projection post-enrichment, both lookup paths, good tests), and we want this in. Your field-selection engine is the core of where the whole tool surface is heading (#597 tracks the broader context-diet effort), so here are two ways forward — pick whichever you prefer:

**Option A — land this PR now (smallest lift).** Two mechanical items and it's merge-ready:

1. Add the `## External assumptions` section the CI gate requires — for this change the answer is simply `None` (it's pure post-processing of already-cached data).
2. Rebase onto main (the branch has picked up conflicts).

We'll take it from there: a follow-up on the v3.0.0 train flips the default to a curated preset and extends selection to the live-mode tools, building directly on your projection engine — with credit to you in the changelog.

**Option B — if you'd enjoy going deeper, the v3 shape is yours to take.** Where we want to land eventually: a single `fields` param (no separate `compact`), where omitted → a curated 10-field preset, `"default"` expands the preset (e.g. `["default", "plaid_category_id"]`), `"all"` returns the full document, and unknown names come back in a `_field_warning` instead of being dropped silently. Happy to split the work — e.g. you take the cache tool, we take `get_transactions_live` parity and the docs/skills updates — and to answer questions along the way.

Either way, thank you — this is exactly the kind of contribution that moves the project, and review turnaround here will be fast.

ignaciohermosillacornejo (changes requested):
let's allign on this, see previous comment. Happy for you to lead this implementation, or I can rework this PR my self.

NikhilGaddam:
Thanks for the fast turnaround on #590, and for the detailed writeup here!

Let's go with **Option A** — I'd rather see the simpler `fields`/`compact` primitive get used for a while before committing to the bigger `"default"`/`"all"`/`_field_warning` redesign, and landing what's already reviewed now beats sitting on it.

Both items are done:
1. Added the `## External assumptions` section (`None` — pure post-processing of already-cached data).
2. Rebased onto `main` and pushed. That also picked up the new context-budget ratchet from #599/#600/#602 — `get_transactions`'s schema legitimately grew with the new `fields`/`compact` properties, so I raised `SCHEMA_BUDGETS.get_transactions` from 4,145 → 5,100 (measured value + the same ~10% headroom every other budget uses) with a comment pointing back to this PR as the justification. Full `bun run check` is green (2523 pass, 0 fail).

Whenever you're ready to scope the v3 shape in #597, happy to help — could take the cache-mode tool while you (or whoever) handles `get_transactions_live` parity and docs, or however you'd rather split it.

ignaciohermosillacornejo (approved):
Everything checks out — External assumptions in place, clean rebase, and the schema-budget bump is exactly how the ratchet is meant to be used (visible, justified, in the same diff). Approving and merging. Thanks again for this one — and yes, would love the help on the v3 shape; let's coordinate on #597 when that train starts moving.

ignaciohermosillacornejo:
Follow-up: the shared field-selection engine generalizing your projection helper just merged (#607), so the v3 reshape is now unblocked and scoped in #604 — earmarked for you as discussed. #605 and #606 are neighboring cache-mode items if you want more. Thanks again for kicking this whole thread off!

PR #594, opened 2026-08-01
PRIVACY.md: correct the network destinations, and detect the drift class

gmhoward9289-ops:
# Summary

PRIVACY.md describes the Firestore-REST architecture the server no longer has. It states in five places that network traffic goes to `firestore.googleapis.com`, and that it happens only in `--write` mode. Both are false — and the second is the one that matters.

## What's actually true

`src/core/graphql/client.ts:22`

```ts
const ENDPOINT = 'https://app.copilot.money/api/graphql';
```

Every live read and every write POSTs full transaction rows, balances, merchants and holdings to Copilot Money's own API. `firestore.googleapis.com` appears nowhere in `src/`. The only Google endpoint is `securetoken.googleapis.com`, used solely to exchange a refresh token — no financial data goes to Google.

More importantly, **`--live-reads` is a read-only flag that turns on all of this traffic**, and PRIVACY.md did not mention it at all. So "read-only mode makes zero network requests" was false for a documented, supported configuration. A privacy-conscious reader could conclude read mode was offline and run `--live-reads` on a shared or monitored machine believing nothing left it.

`README.md:227` already describes this correctly. Only PRIVACY.md was behind.

## What changed

- Three modes instead of two: default (cache-only, no network), `--live-reads` (read-only **but online**), `--write` (implies `--live-reads`)
- A table of the exact destinations contacted per mode
- The Firebase token exchange separated from the data path, so "Google is contacted" isn't confused with "Google receives your finances"
- `Last Updated` refreshed

No code behaviour changes.

## External assumptions

None. This PR documents the endpoints the code already uses; it adds no assumption about Copilot's API. The destination set is now derived from `src/` by the checker below rather than asserted.

## Bug fix?

Root cause:       PRIVACY.md was written against the Firestore-REST design and never revised when the GraphQL client replaced it; `--live-reads` was added later and never documented as network-bearing.
Bug class:        Documentation that describes an architecture the code no longer has — invisible to every existing check, since tests exercise behaviour and lint reads syntax, and neither opens a Markdown file.
Detector added:   `scripts/check-privacy-endpoints.ts`, wired into `bun run check`. It compares two sets: every `https` host reachable at runtime in `src/` (comments stripped, so `@see` links to Plaid/protobuf/GitHub don't count) against every host named in PRIVACY.md. A host in code but not the doc is **undisclosed traffic**; a host in the doc but not the code is a **stale claim**; both fail. Adding a new destination now requires saying so in the privacy policy in the same PR.
Siblings checked: `README.md` (already correct), `CLAUDE.md` (already correct), `manifest.json`/`server.json` (no endpoint claims). The checker covers PRIVACY.md, which was the only drifted document.
Ledger updated:   n/a — not an external-assumption bug; nothing about Copilot's API changed, only our description of which host we call.

### Detector verified against the real bug

Against the pre-fix PRIVACY.md:

```
  Contacted by src/ but NOT named in PRIVACY.md:
    app.copilot.money
    securetoken.googleapis.com
  Named in PRIVACY.md but NOT contacted by src/:
    firestore.googleapis.com
  exit 1
```

Against an injected `https://analytics.example.net` in the GraphQL client: reported as undisclosed, exit 1. Against the corrected document: exit 0.

`bun run check` passes — 2384 tests, 0 fail.


ignaciohermosillacornejo (commented):
This is careful, verifiable work — thank you. We checked every factual claim against the code and they all hold: `src/` has exactly two network call sites (`src/core/graphql/client.ts:354` → `app.copilot.money/api/graphql`, `src/core/auth/firebase-auth.ts:103` → `securetoken.googleapis.com`), no `firestore.googleapis.com` anywhere, and the credential-flow description matches the implementation exactly. We also reproduced your detector-vs-real-bug transcript verbatim — it catches the actual drift in both directions. The doc rewrite is merge-ready as written, and once this lands, this whole class of doc drift stays caught for good.

**One must-have before merge:** the detector currently only runs in the local pre-push hook — `.github/workflows/test.yml` runs individual steps rather than `bun run check`, so CI (including on this PR) never executes `check:privacy-endpoints`. A fork PR or GitHub-web edit could reintroduce exactly the drift this PR catches. The fix is one workflow step mirroring the existing `check:version-sync` step:

```yaml
      - name: Privacy endpoints drift check
        run: bun run check:privacy-endpoints
```

Add that (plus a quick rebase — you're 1 commit behind main), and this is approved. **Or, if you prefer, say the word and we'll push that line and the rebase to your branch ourselves** — your commits and authorship stay intact either way.

**Optional hardening, zero pressure** — fold in if you feel like it, otherwise we'll file them as small follow-up issues (they'd be nice second PRs if you want them):

- `HOST_RE` only matches `https://` literals, so a plain `http://` fetch or a template-built URL passes silently; `https?:\/\/` plus a header sentence disclosing the dynamic-URL limitation would close most of that.
- Two of the three `DOC_ONLY_HOSTS` entries (`plaid.com`, `protobuf.dev`) are already dead — they only occur in block comments, which `stripComments` removes — so today they only add blind spots. Only `github.com` is genuinely needed (error-message literal in `src/core/database.ts:595`).
- Latent edge case: the block-comment regex can be blinded by `/*` and `*/` appearing inside separate string literals (e.g. globs like `'src/*'`).

Separately (not this PR's burden): PRIVACY.md still doesn't mention that live/write modes read browser profile storage to extract the Firebase refresh token — README discloses it, PRIVACY.md doesn't. We'll file that as its own issue unless you'd like to fold a sentence into the Data Access section here.

Thanks again — between this and #588, the repo's drift-detection story is meaningfully stronger.

gmhoward9289-ops:
Thank you — genuinely. This is the first repo where I've done work like this, and having every factual claim independently checked against the code, and the detector transcript reproduced verbatim rather than taken on trust, is more care than I expected. It's taught me more than the PR did.

I've taken the must-have myself, plus the first two hardening items:

- **`Privacy endpoints drift check` step** in `test.yml`, next to `check:version-sync` and `check:server-json`. No secrets, so it runs on fork `pull_request` events — which was the actual hole.
- **`HOST_RE` now matches `http://` as well as `https://`.** Nothing in `src/` uses plain http today, so it catches nothing right now; it's there so a future cleartext endpoint — the case most worth disclosing — can't pass a TLS-only scan.
- **Dropped `plaid.com` and `protobuf.dev` from `DOC_ONLY_HOSTS`.** You were right that they're dead: both appear only inside `/** */` blocks that `stripComments()` removes, so they never reached the set and were pure blind spots. `github.com` stays — it's a real string literal in `src/core/database.ts:595`.
- **Documented the blind spot instead of pretending it isn't there.** The scan matches URL *literals*, so a host assembled at runtime is invisible to it. The header now says plainly that a green run means "no new destination was hardcoded in `src/`", not "the server contacts nothing else", and flags that a constructed URL needs disclosing by hand.
- **Rebased onto current main** — twice, since #588 merged partway through and main gained the very lines this PR touches. Three conflicts the second time, all the same additive shape: the `check` chain, the `scripts/` allowlist, and the workflow, where "Tool count drift check" and "Privacy endpoints drift check" wanted the same slot. Both kept in every case; the two drift checks now sit next to each other in the quality job.

I also added the unit test I should have written in the first place. Every sibling gate has one and this didn't; you flagged exactly that on #588 and I didn't carry the lesson across. It drives the script end-to-end against synthetic trees via a new `CHECK_PRIVACY_ENDPOINTS_ROOT` override, matching `CHECK_TOOL_COUNTS_ROOT` and `CHECK_DEPS_PINNED_PACKAGE_JSON`. Ten cases: agreement, undisclosed traffic, stale claims, both directions reported at once, block and line comments ignored, an undisclosed `http://` host, and the real repository. One test deliberately pins the dynamic-URL blind spot as a *known* gap, so it stays documented rather than being mistaken for coverage.

The block-comment edge case (`/*` and `*/` in separate string literals) I've left alone — happy to take it as its own issue, along with the browser-profile-storage sentence for PRIVACY.md's Data Access section, if you'd still like that one folded in here instead.

---

One thing I should raise rather than let you find it. On #588 you noted that "48 tools" overstates the real surface, since no single mode exposes all of them. My rebase there updated that number to **50** — so I made your objection worse, not better, and `check-tool-counts.ts` now mechanically enforces the number with two `allTotal` assertions. The largest any one configuration actually exposes is **44** (`--live-reads --write`: 8 surviving cache + 17 live + 19 write), because the 6 swapped cache tools and their `_live` replacements are both counted.

50 is a true count of *definitions*, but "50 tools covering every aspect of your personal finances" reads as a capability claim, and no user can ever have 50. I didn't want to change the framing unilaterally on your landing page, or widen a diff you'd already called scope-crept. Say the word and I'll switch it to 44 (or "50 tool definitions across three modes") and update the two assertions to match — either here or as a follow-up, whichever you prefer.

ignaciohermosillacornejo (approved):
Approved — the CI step closes the actual hole (your read is exactly right: fork `pull_request` events were the gap, and the step runs there because the check needs no secrets). Everything beyond the must-have is a welcome bonus: the `http://` widening rationale, the dead-allowlist cleanup, the honest blind-spot documentation — and especially the ten-case unit test with the ROOT-override pattern. Carrying the #588 lesson across without being asked is the thing code review hopes for and rarely gets.

Your three questions:

- **Landing-page framing:** you've independently arrived at what we filed earlier today as #610 — same conclusion, same 44 arithmetic. It's yours if you want it, as its own PR (per-mode framing preferred, but "50 tool definitions across three modes" is acceptable too — see the issue). Don't fold it here; this PR's scope is right as-is.
- **Block-comment edge case:** agreed, its own issue — now filed as #614, claim it if you like.
- **Browser-profile-storage disclosure:** filed as #615 per the review's promise; also standalone, also yours if you want it.

And thank you for the kind words. The verification-heavy style is deliberate and it'll stay — contributions here get checked, not guessed at. Merging shortly.

PR #603, opened 2026-08-03
fix(budgets): anchor live budget windows to current month (#598)

### LINKED ISSUE #598: get_budgets_live: months_window anchored to end of fetched history, not current month
## Summary

During a read-only response-size audit (2026-08-02), `get_budgets_live` called with `months_window: 1` returned budget rows whose `amounts` month-map contained a single key — a `YYYY-MM` month roughly **12 months in the future** of the current month. Expected: a 1-month window anchored to the current month.

## Suspected cause

The windowing logic appears to slice from the **end of the fetched history** (Copilot's budgets payload includes future months, e.g. scheduled/rollover projections) instead of anchoring to today's month. If the server returns N months spanning past→future, taking the last `months_window` entries yields the most-future months, not the current one.

## Impact

Any caller using `months_window` to scope "current budget vs spend" silently gets a future month's numbers — plausible-looking amounts, wrong period. This is a correctness bug independent of response size.

## Repro

1. Run the server with `--live-reads` (authenticated browser session).
2. Call `get_budgets_live` with `months_window: 1`.
3. Inspect the `amounts` keys on any returned budget — observed a month ~12 months ahead of the current date instead of the current month.

## Notes for the fix (Bug Response Ritual applies)

- Root cause: confirm where the window slice happens (`src/tools/live/budgets.ts` / the budgets query mapper) and what ordering the wire payload guarantees.
- Bug class: "window/limit applied to an array whose ordering or anchor differs from the caller's mental model" — siblings to check: every other live tool that slices a server-returned series by count or window (`months_window`-style params, history caps, `time_frame` handling).
- Detector: a conformance/smoke assertion that a current-anchored window always contains the current period.
- Ledger: add/update the relevant entry in `src/conformance/ledger.ts` for the budgets payload ordering assumption.

joyheroes:
## Summary

Fix `get_budgets_live` month window selection so that `months_window` is anchored to the current month instead of selecting the last entries from the payload.

This prevents future projected budget months from incorrectly replacing the current budgeting period.

## Changes

- Update live budget window trimming logic to use the current month as the anchor.
- Preserve existing `months_window: 0` behavior for full history output.
- Add regression coverage for payloads containing past, current, and future months.
- Add smoke and conformance coverage for the month anchoring assumption.

## Testing

Verified with:

- `bun test tests/tools/live/budgets.test.ts tests/conformance/ledger.test.ts tests/scripts/read-smoke-coverage.test.ts`
- `bun run typecheck`
- `bun run lint`
- `bun run format:check`
- `bun test tests/integration/mcpb-bundle.test.ts`
- `bun run check`

Fixes #598

---

## External assumptions

**None new.** This corrects how an already-fetched payload is windowed; it adds no new assumption about Copilot's API or data.

It does make an existing, previously-unstated assumption explicit and wrong-proof: Copilot's `categories` response **can contain future-dated budget months** (scheduled/rollover projections). The old code implicitly assumed it could not. That is now recorded as an evidence note on the existing `Query.categories:response` ledger entry rather than as a new entry, since the response *shape* is unchanged — only our reading of it was wrong. Guarded by `tests/tools/live/budgets.test.ts` and a new assertion in `scripts/smoke/budgets.ts`.

## Bug fix?

```text
Root cause:       trimAmountsToWindow picked the trailing N keys by lexicographic order
                  from the payload and never consulted the clock. Copilot returns
                  future-dated budget months, so those won the trailing-N selection and
                  displaced the current month — "the last 12 keys we were sent" is not
                  "the last 12 months".
Bug class:        Proposing a new one: `data-relative-anchoring` — deriving a "current"
                  reference point from the extent of returned data instead of from an
                  external truth (here, the clock). Safe exactly while the data happens
                  to end where the caller assumes it does.
Detector added:   Instance-level only, honestly. Unit tests pin the anchoring against a
                  mocked clock (which is what makes them deterministic rather than
                  passing by luck of the calendar), and scripts/smoke/budgets.ts gains a
                  4th assertion that months_window=1 returns the current month against
                  the live backend. There is NO class-level gate for "windows anchored to
                  data rather than to now" — see below.
Siblings checked: One other trailing-N site examined: `sortedTxns.slice(-5)` in
                  get_recurring_transactions (src/tools/tools.ts). NOT an instance — it
                  samples the most recent N of what exists rather than claiming a window
                  relative to now, so there is no anchor to get wrong. No other
                  data-anchored windows found in src/.
Ledger updated:   Yes — evidence note added to the Query.categories:response entry
                  (src/conformance/ledger.ts), recording that month ordering/anchoring is
                  now guarded by the budgets tests and live smoke.
```

### Maintainer note

Sections filled in by @ignaciohermosillacornejo a week after the code was ready, so a correct live fix wasn't held up by paperwork the contributor hadn't been asked for when they opened it. The class name above is my proposal, not the contributor's — if `data-relative-anchoring` earns a second instance it should get an entry in [`docs/bugs/`](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/tree/main/docs/bugs); one instance is an accident.

Worth noting the fix's own detector is the interesting gap: the unit tests only became meaningful once the clock was mocked. Before that they'd have passed or failed depending on when in the year CI ran — a test whose correctness is seasonal is a close cousin of a vacuous one.

### DIFF
--- a/scripts/smoke/budgets.ts
+++ b/scripts/smoke/budgets.ts
@@ -8,6 +8,7 @@
  *   2. Warm call (within TTL) returns _cache_hit=true with sub-50ms latency
  *   3. refresh_cache --scope budgets invalidates categoriesCache (the alias)
  *      and triggers a refetch on the next get_budgets_live call
+ *   4. months_window=1 anchors to the current month, not future projection rows
  *
  * Exits non-zero on any assertion failure. Output is intended to be pasted
  * into the PR description.
@@ -33,6 +34,19 @@ async function main(): Promise<void> {
   log('cold', { rows: cold.count, total: cold.total_budgeted, hit: cold._cache_hit, ms: coldMs });
   if (cold._cache_hit !== false) throw new Error(`expected cold _cache_hit=false`);
   if (cold.count === 0) throw new Error('expected at least one budget');
+
+  const now = new Date();
+  const currentMonth = `${now.getFullYear()}-${String(now.getMonth() + 1).padStart(2, '0')}`;
+  const currentWindow = await tools.getBudgets({ months_window: 1 });
+  const offAnchor = currentWindow.budgets
+    .flatMap((budget) => Object.keys(budget.amounts ?? {}))
+    .filter((month) => month !== currentMonth);
+  log('months_window=1', { currentMonth, offAnchor: offAnchor.length });
+  if (offAnchor.length > 0) {
+    throw new Error(
+      `months_window=1 returned non-current budget months: ${[...new Set(offAnchor)].join(', ')}`
+    );
+  }
 
   // 2. Warm
   const t1 = Date.now();
--- a/src/conformance/ledger.ts
+++ b/src/conformance/ledger.ts
@@ -287,13 +287,14 @@ function queryOperation(name: string): LedgerEntry {
  * passed here MUST have a matching QUERY_RESPONSE_SCHEMAS entry — enforced
  * bidirectionally by tests/conformance/ledger.test.ts.
  */
-function gatedQueryResponseShape(name: string): LedgerEntry {
+function gatedQueryResponseShape(name: string, overrides?: Partial<LedgerEntry>): LedgerEntry {
   return {
     surface: `Query.${name}:response`,
     kind: 'response-shape',
     oracle: `runtime:${READ_RESPONSE_SHAPE_RUNTIME_CHECK}`,
     class: 'gated',
     evidence: READ_RESPONSE_SHAPE_GATED,
+    ...overrides,
   };
 }
 
@@ -766,7 +767,13 @@ export const CONFORMANCE_LEDGER: readonly LedgerEntry[] = [
       'surfaced via _dropped_invalid_rows + a deduped stderr warning.',
   },
   queryOperation('categories'),
-  gatedQueryResponseShape('categories'),
+  gatedQueryResponseShape('categories', {
+    evidence:
+      READ_RESPONSE_SHAPE_GATED +
+      '; budget.month ordering/anchoring is guarded by get_budgets_live synthetic tests ' +
+      '(tests/tools/live/budgets.test.ts, issue #598) and the budgets live smoke ' +
+      '(scripts/smoke/budgets.ts).',
+  }),
   queryOperation('tags'),
   gatedQueryResponseShape('tags'),
   queryOperation('recurrings'),
--- a/src/tools/live/budgets.ts
+++ b/src/tools/live/budgets.ts
@@ -43,23 +43,36 @@ export interface GetBudgetsLiveResult {
   _cache_hit: boolean;
 }
 
+function currentMonthKey(now = new Date()): string {
+  return `${now.getFullYear()}-${String(now.getMonth() + 1).padStart(2, '0')}`;
+}
+
+function monthIndex(month: string): number {
+  const parts = month.split('-');
+  const year = Number(parts[0]);
+  const monthNumber = Number(parts[1]);
+  return year * 12 + monthNumber - 1;
+}
+
 /**
- * Return a trimmed amounts map containing only the most recent N months
- * by lexicographic key order. `YYYY-MM` keys sort chronologically, so this
- * is equivalent to "trailing N months" for normal data. The function does
- * NOT consult the current date; if the input contains future-dated keys
- * (e.g., projected budgets), they win the trailing-N selection. If the
- * input has fewer than N months, returns the input unchanged.
+ * Return a trimmed amounts map containing only the current month plus the
+ * preceding N-1 calendar months. Copilot can return future budget months
+ * (scheduled/rollover projections), so anchoring to the last payload key would
+ * return future amounts for callers asking for the current budget window.
  */
 function trimAmountsToWindow(
   amounts: Record<string, number>,
   windowMonths: number
 ): Record<string, number> {
+  const maxMonth = monthIndex(currentMonthKey());
+  const minMonth = maxMonth - windowMonths + 1;
   const keys = Object.keys(amounts).sort();
-  if (keys.length <= windowMonths) return amounts;
-  const kept = keys.slice(-windowMonths);
   const trimmed: Record<string, number> = {};
-  for (const k of kept) trimmed[k] = amounts[k]!;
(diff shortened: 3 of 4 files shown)


ignaciohermosillacornejo:
Welcome, and thank you — this is a genuinely strong first contribution. Before anything else: **we verified your fix against the live endpoint, and it's correct.** Pre-fix code with `months_window: 1` returns a month ~12 months in the future on real data; your code returns the current month. Your regression test is the real thing too — we reverted the fix and exactly that test fails (its fixture spans past + current + future months, which is precisely right). The root-cause diagnosis (tail-slice of a series Copilot extends with future projections) matches the mechanism exactly, and your fix actually *restores* the contract the docstring always promised. Nice work.

The code would merge on its merits. What's left is our PR-process layer — the failing "Check required PR sections" gate is the repo's core discipline for bug fixes, not boilerplate. To make it painless, here's a skeleton you can paste into the PR body and tweak (we did the sibling audit for you — it came back clean):

```markdown
## External assumptions
Month keys in the budgets payload parse as `YYYY-MM` (evidence: live probe in this PR's
review — `monthIndex` NaNs and silently drops any other format). Declared in the ledger.

## Bug Response Ritual
Root cause: `trimAmountsToWindow` took `keys.sort().slice(-N)` — a tail slice of a series
Copilot extends ~12 months into the future with projections — instead of anchoring to the
current month.
Bug class: count/window slice applied to a server-returned series whose extent differs from
the caller's mental model.
Detector added: regression test with past+current+future fixture months; smoke assertion
that the window contains the current month (see below).
Siblings checked: none found — `time_frame` params (networth/balance history/investment
prices) are server-side windows; holdings/transactions slices are explicit caller pagination;
`sortedTxns.slice(-5)` in tools.ts is an unanchored sample by design.
Ledger updated: yes — `YYYY-MM` month-key format assumption.
```

Three small code asks alongside that:

1. In `scripts/smoke/budgets.ts`, also assert the current month is **present** in the window (right now a regression returning empty `amounts` everywhere would pass the no-off-anchor check), and bump `passed` to 4.
2. Your anchored window means stale-only budgets can now return a literal empty `amounts: {}` — either omit the key when empty (matching `projectCategory`'s convention) or document the new shape; your call, just make it deliberate.
3. Add a CHANGELOG entry under `### Fixed` — every user-visible fix gets one.

Two purely-optional notes: anchoring to the payload's own `budget.current.month` instead of client-local time would be drift-proof across month-boundary hours (fine to skip); and don't worry about the skipped `review` workflow — fork PRs need maintainer approval for it to run, that's on us. A rebase onto main will also pick up the new `check:tool-counts` gate; we've verified your branch passes the full current gate after a clean merge, so it should be uneventful.

Happy to answer anything — fast turnaround promised. Looking forward to landing this.

ignaciohermosillacornejo:
Thanks for this — the code side is in good shape. Quality Checks and Tests are both green, and I appreciate that you added regression, smoke, and conformance coverage for the month-anchoring assumption rather than just the one failing case.

The only thing blocking merge is the **Check required PR sections** gate, which is about the PR description rather than the code. It needs two things added to the body:

**1. `## External assumptions`** — a section listing any NEW assumption this PR makes about Copilot's API or data, each with an evidence class (`probe transcript` / `live round-trip` / `unverified` + a ledger entry). For this change I'd expect the honest answer to be close to:

```markdown
## External assumptions

None new — the fix corrects how we window data we already fetch, and adds no
new assumption about the server's response.
```

`None` is a perfectly valid answer; the gate just wants it stated explicitly rather than omitted, so "we didn't think about it" and "we thought about it and there are none" don't look identical.

**2. The Bug Response Ritual block**, since the title is a `fix:`. From [CONTRIBUTING.md](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/blob/main/CONTRIBUTING.md#bug-response-ritual):

```text
Root cause:       <the mechanism, not the symptom>
Bug class:        <the class, not the instance>
Detector added:   <the class-level gate that catches the whole class>
Siblings checked: <other instances audited, or "none found">
Ledger updated:   <ledger entries touched, or "n/a">
```

For `Bug class`, we've just published [`docs/bugs/`](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/tree/main/docs/bugs) — a post-mortem knowledge base with a class list. Worth a look: if this bug fits an existing class, that recurrence is a more useful finding than the bug itself. A window anchored to fetched data rather than to "now" may well be its own class we haven't named yet, and proposing one is very welcome.

`Detector added: none — instance-only regression test` is an acceptable answer if that's the truth. We'd rather have an honest "no detector" than an invented one.

No rush — I'll leave this open for a week or so. If you'd rather not deal with the boilerplate, say the word and I'll fill it in and merge; I only held off because the external-assumptions section is meant to be the author's attestation about their own change, and it seemed wrong to write yours for you without asking.

ignaciohermosillacornejo (approved):
Approving and merging. Verified against current `main` rather than trusting the 3 August CI run — the branch was nine days behind and `main` has moved a lot since (decoder rework, new cache smoke). Rebased locally, clean, `bun run check` green at 2649 pass / 0 fail.

I filled in the two required sections myself rather than let a correct live fix sit waiting on paperwork you weren't asked for when you opened this. The substance is yours.

Two things I want to call out, because they're the parts that were easy to get wrong and you got right:

**Anchoring to the clock rather than to the payload.** The old code took the trailing N keys by sort order, which silently assumes the data ends where the caller thinks "now" is. Copilot returns future-dated projection months, so that assumption fails in exactly the case the parameter exists to serve. The fix reads as a small diff but it's a change of reference frame, and that's the right one.

**Mocking the clock in the tests.** This is the detail I'd have flagged if it were missing. Without it those assertions would pass or fail depending on when in the year CI happened to run — a test whose correctness is seasonal is a close cousin of one that can't fail at all. Pinning the date is what makes them mean something.

Also appreciated: you extended the live smoke and added a ledger evidence note rather than only fixing the code path. That's the repo's convention and it's not obvious from the outside.

One note on the class name in the ritual block — `data-relative-anchoring` is my proposal, not yours. If it earns a second instance it'll get a proper entry in [`docs/bugs/`](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/tree/main/docs/bugs); one instance is an accident, not a class.

Thanks for this, and sorry it sat as long as it did.

Issue #597, opened 2026-08-02
Context diet: reduce response token cost across all tool surfaces (v3)


ignaciohermosillacornejo:
## Summary

A full-surface context audit (2026-08-02) measured the serialized response cost of all 15 cache-mode read tools (local synthetic-comparable run with per-field breakdowns) and all 16 live read tools (one bounded call each through the real MCP path). Result: 11 of 31 read surfaces carry significant bloat for an LLM caller, falling into four patterns — flat field width, embedded time-series/nested objects, GraphQL plumbing leakage, and serialization overhead. This is the tracking issue for the cleanup; breaking items ship together as **v3.0.0** alongside #592/#593. The get_transactions terse-by-default reshape is tracked in #604.

Headline measurements (chars are raw response text; tokens ≈ chars/4):

| Tool | Typical call | Dominant bloat |
|---|---|---|
| `get_transactions` (cache) | ~120K chars / 100 rows; ~768K for a 6-month pull | 31 flat fields, no single dominator → field selection (#592/#593) |
| `get_transactions_live` | ~80K chars/month at default limit 100 | same, plus id triplet ≈28% of row bytes |
| `get_investment_prices` (cache) | ~204K chars bare | nested `prices` series = 98.3% |
| `get_recurring_live` | ~56K chars bare | `rule` (server matcher config) 23% + embedded `payments` 22% |
| `get_categories_live` | ~34K chars bare | embedded `budget` object 62% (duplicates `get_budgets_live`) |
| `get_top_movers_live` | ~15K chars / 3 rows | intraday `price_points` series = 95% |
| `get_investment_balance_live` | ~15K chars | `history` 87% — 214 daily rows, uncapped |
| `get_recurring_transactions` (cache) | ~38K chars bare | embedded `transactions` 29% + `confidence_reason` prose 15% |
| `get_accounts` / `_live` | ~17K / ~11K chars | cache: embedded `holdings` ~20% + `official_name`/`original_*`/`user_id` dupes; live: sync-machinery flags ~23% |
| `get_upcoming_recurrings_live` | ~8K chars | same `rule`+`payments` shape as recurring |

Cross-cutting: `__typename` leaks into 6+ live tools (22% of every tag row, 36% of every networth row); every response is pretty-printed (`JSON.stringify(result, null, 2)` in `src/server.ts`), a 10–20% whitespace tax; the 33 base tool schemas cost ~48K chars (~12K tokens) of every session before any call.

[...]

PR #607, opened 2026-08-03
feat(tools): shared field-selection engine (#597 v3 base)

ignaciohermosillacornejo:
# Summary

Extract #593's inline `projectTransactionFields` into a shared, generic field-selection engine (`src/tools/field-selection.ts`) so the upcoming v3.0.0 PRs (cache reshape, live parity, accounts preset) all build on one implementation instead of forking copies. Non-breaking in 2.x: opt-in `fields`/`compact` behavior on `get_transactions` is otherwise preserved (no selection → full documents, `compact` keeps its 7-field set, `fields` beats `compact`), with two deliberate additive changes: (1) unknown requested field names are reported in a top-level `_field_warning` on the response (previously silently ignored); (2) `"default"`, `"all"` and `"*"` are now interpreted as selection tokens — on main they were ordinary (nonexistent) field names, so e.g. `fields: ["all"]` projected every row down to `{}` and now returns full documents. These three names are permanently unselectable as literal fields. Part of #597.

## What's in the engine

- `DEFAULT_TRANSACTION_FIELDS` — the 10-field v3 baseline preset (`transaction_id`, `date`, `amount`, `name`, `category_name`, `account_id`, `item_id`, `pending`, `excluded`, `internal_transfer`). The existing 7-field `compact` preset is untouched and still powers `compact: true`.
- `expandFieldSelection(fields, preset)` — `"default"` expands to the preset (order-preserving, deduped, e.g. `["default", "plaid_category_id"]`); `"all"`/`"*"` anywhere disables projection; empty/undefined → `undefined` (caller decides the default; in 2.x that means no projection). `"default"` with an empty/absent preset throws instead of silently expanding to nothing (all-empty rows).
- `projectRows(rows, fields, opts)` → `{ rows, warning? }` — generic allowlist projection over any row shape, plus unknown-name detection: against `opts.knownFields` when provided (the authority set — works even with zero rows, and overrides row keys in both directions), else falling back to own-key presence on the rows (`Object.hasOwn`, so inherited prototype names like `toString` cannot mask the warning). Non-array `fields` (host skipped schema validation) throws a clear Error instead of silently iterating. Projection skips a hostile own `__proto__` key so JSON.parse'd rows cannot pollute projected objects.

## Wiring

- `projectTransactionFields` in `src/tools/tools.ts` now delegates to the engine. Known-name set is derived programmatically from the Transaction zod schema's keys plus the `category_name`/`normalized_merchant` enrichment keys — model drift cannot desync it.
- Both `get_transactions` paths (filtered list + single-ID lookup) surface the engine's warning as `_field_warning` (same pattern as `_cache_warning`).
- `fields` param description documents the `"default"`/`"all"`/`"*"` tokens and `_field_warning`; schema is 4,799 chars against the 5,100 ratchet budget (no bump needed).
- `bun run sync-manifest` ran clean (manifest embeds only tool-level descriptions; no drift).

## Tests

- `tests/tools/field-selection.test.ts` (33 tests): token expansion (default/all/*/dedupe/ordering), generic projection, warning generation with and without `knownFields` — including three tests that specifically discriminate the knownFields authority branch from the row-key fallback (mutation-verified: disabling the branch fails them) — empty/undefined/non-array handling, empty-preset throw, `__proto__` hardening, input-mutation guard. Synthetic data with Firestore-shaped opaque IDs.
- `tests/tools/tools.test.ts`: `_field_warning` asserted on both the filtered-list and single-ID lookup paths (the latter mutation-verified), `fields: []`-beats-`compact` pin, and a drift ratchet asserting every full-width row key is in the derived known-field set (the schema is `.passthrough()`, so undeclared decoder keys would otherwise make `_field_warning` self-contradicting).
- `bun run check` fully green; new test files run twice for determinism.

## External assumptions

None — internal refactor + additive warning over our own cached data.


ignaciohermosillacornejo:
Review findings addressed in ebc25b7:

- **Medium (tokens undocumented):** the `fields` description now documents `"default"` (10-field baseline) and `"all"`/`"*"` (full document). Schema is 4,799 chars against the 5,100 budget — no bump needed.
- **Low (`out as T`):** annotated the deliberate widening at the cast site — `Partial<T>` would force every caller (who already knows what it selected) to null-check each field.
- **Nit 1 (double guard):** reordered `projectRows` so `expandFieldSelection` runs first; its non-array guard now also covers `detectUnknownFields`, leaving a single load-bearing check. (The outer call wasn't redundant as written — `detectUnknownFields` ran before `expandFieldSelection` — the reorder makes one guard suffice.)
- **Nit 2 (`fields: []` + `compact: true`):** behavior kept (explicit empty wins over compact → no projection, unchanged from #593), now documented at the fallback site and pinned by a dedicated test.

ignaciohermosillacornejo:
Adversarial re-review findings addressed in 2041d8b:

1. **HIGH — knownFields branch mutation-invisible:** three discriminating tests added (a knownFields name absent from every row stays silent; zero rows + knownFields still warns on a bogus name; a name present on a row but not in knownFields warns). Mutation-verified: replacing `if (opts.knownFields)` with `if (false)` fails all three (214 pass / 3 fail), then reverted.
2. **MEDIUM — single-ID lookup `_field_warning` spread untested:** dedicated test added; deleting the spread fails exactly that test (183 pass / 1 fail), then reverted.
3. **MEDIUM — claim accuracy:** CHANGELOG and PR body now name both deliberate behavior changes — `_field_warning` and the token interpretation (`"default"`/`"all"`/`"*"` were ordinary nonexistent names on main; `fields: ["all"]` used to project rows to `{}`) — plus a sentence that these three names are permanently unselectable as literal fields.
4. **LOW-MED — `"default"` with empty preset:** `expandFieldSelection` now throws a clear Error naming the missing preset instead of returning `[]` (all-empty rows). Tested at both the engine and `projectRows` levels.
5. **LOW — prototype-chain leaks:** fallback detection switched to `Object.hasOwn` (test: `['toString']` now warns), and projection skips a hostile own `__proto__` key (test: JSON.parse'd row cannot pollute projected objects).
6. **LOW — drift ratchet:** new test asserts every full-width `get_transactions` row key is in the derived known-field set, pinning today's verified-clean state against `.passthrough()` schema drift.
7. **LOW — stale figure:** PR body updated to the measured 4,799 chars (vs 5,100 budget).

ignaciohermosillacornejo:
3rd-pass nits addressed in c3385e7: (1) `assertFieldsArray` now asserts `readonly string[] | undefined`, so the `!fields` guard is no longer dead code to TypeScript; (2) the token-plus-typo warning behavior is kept as the deliberate contract (every unknown requested name is surfaced, whether or not projection made it consequential), now documented at the call site and pinned by a test (`["all", "not_a_real_field"]` returns full rows AND warns).

Issue #596, opened 2026-08-02
Mutation-guard registry: make safety invariants prove they have detectors


ignaciohermosillacornejo:
## Problem

A safety guard can be fully covered by tests and still have **no detector**. We hit this twice in one review of #587:

**1. `update_transactions` stop-on-first-failure.** Deleting *both* `stopOnError` guards in `runBoundedPool` left all 2435 tests passing. The only default-mode test asserted:

```
/update_transactions failed at transaction_id=txn-05 \(\d+\/8 succeeded\)/
```

`\d+` matches whether the batch stopped after 4 writes or ran all 8. The test pinned the error *message*; the property — "entries queued behind the failure are never written" — was never asserted. A refactor dropping that guard would let a 200-edit batch that fails at row 3 write the remaining 197 against real financial data, with an error string indistinguishable from correct behaviour.

**2. A steering ratchet in the same PR.** `/name.*note.*date.*amount/` was meant to pin a specific sentence in a tool description, but matched the description's unrelated field *enumeration*. Deleting the sentence left it green. (Worse: the commit message claimed the test was mutation-tested. Part of it had been; that assertion had not.)

Same class both times: **a predicate satisfied by both the correct and the broken behaviour.**

### Why existing gates miss it

- **Coverage can't see it.** Codecov reported 98.7% patch coverage on #587 and both guards were *executed*. Line coverage answers "did this run", not "would anything notice if it were wrong."
- **Review catches it only by luck.** It was found because the reviewer was explicitly told to mutation-test rather than eyeball. That is not a repeatable property of review.
- **We already claim to do this.** PR bodies in this repo routinely assert "mutation-tested", and CONTRIBUTING's bug-response ritual demands a class-level detector. Nothing verifies either claim — as demonstrated above, one of those claims was wrong.

## Proposal: a mutation-guard registry

Not full mutation testing (Stryker) — too slow for `check`, and most surviving mutants in this codebase would be noise. Instead, a small explicit ledger of **designated safety invariants**, each paired with the exact mutation that must break the suite:

```ts
// scripts/mutation-guards.ts
export const MUTATION_GUARDS = [
  {
    name: 'update_transactions stops on first failure',
    file: 'src/tools/tools.ts',
    remove: 'if (stopOnError && failures.length > 0) return;',
    expectFails: 'tests/tools/update-transactions-batching.test.ts',
  },
  {
    name: 'bulk write refuses an unbounded row set',
[...]

PR #616, opened 2026-08-03
test: cover soft-deleted transaction reads

### LINKED ISSUE #609: Test fixture gap: createCombinedDb has no user_deleted transaction
Noted informationally by the AI review on #599: the synthetic test DB built by `tests/helpers/test-db.ts` (`createCombinedDb`) never includes a transaction with `user_deleted: true` — Copilot's soft-delete mechanism for transactions (other collections use real tombstones; transactions don't).

That means the read-path invariant "deleted transactions never appear in get_transactions results" is untested against the fixture. If the filter regressed, no test would notice.

**Work:** add one `user_deleted: true` transaction to the combined fixture and assert the transaction read paths (list + by-ID lookup) exclude it. Small, test-only, no GraphQL surface, no live credentials needed. Mutation-check your test per repo convention: temporarily break the filter and confirm the new test fails.

JasonColapietro:
## Summary

Add a soft-deleted transaction to the shared combined LevelDB fixture and verify that both transaction lists and direct ID lookups exclude it.

Closes #609

## Why

The decoder already filters `user_deleted: true`, but the shared combined fixture dropped that field and could not protect the public read paths from regression.

## Testing

- `bun test tests/context-budget.test.ts`
- `bun run check`
- Mutation check: temporarily bypassing the decoder guard made the new regression test fail; restoring it returned the suite to green.

## Scope

Test fixture fidelity and regression coverage only. No runtime behavior changes.

## Risks

Low. The new row is synthetic and is filtered before response-budget measurement.

## External assumptions

None

### DIFF
--- a/tests/context-budget.test.ts
+++ b/tests/context-budget.test.ts
@@ -132,6 +132,7 @@ const ACC_BROKERAGE = 'acc_6dRz1KvMpN4WxT7yQc3U';
 const ITEM_BANK = 'item_5cJn3WfHbD8XqS1kMt6Z';
 const ITEM_BROKER = 'item_8gLp4XdJcF2VzB9sWk5A';
 const GOAL_FUND = 'goal_1sVc8BnMdQ5XzK3wRj9T';
+const DELETED_TRANSACTION_ID = 'txn_4nYc8LbQwR2VzK7mJd5P';
 
 async function seedFixture(): Promise<void> {
   await createCombinedDb(DB_PATH, {
@@ -182,6 +183,15 @@ async function seedFixture(): Promise<void> {
         name: 'Synthetic Streaming',
         category_id: CAT_TRANSPORT,
         is_transfer: false,
+      },
+      {
+        transaction_id: DELETED_TRANSACTION_ID,
+        account_id: ACC_CHECKING,
+        amount: 27.5,
+        date: '2024-03-06',
+        name: 'Synthetic Deleted Merchant',
+        category_id: CAT_GROCERIES,
+        user_deleted: true,
       },
     ],
     accounts: [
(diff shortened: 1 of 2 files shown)


ignaciohermosillacornejo (approved):
Approving — this is exactly the right shape for the gap in #609, and it closes it completely.

I ran an independent review that went past "does it pass" to **"would it fail?"** — the standard this repo learned the hard way in #596 (a test can execute a guard and still not detect its deletion; coverage doesn't imply detection). Your test clears that bar, and both halves clear it *independently*:

- Commenting out the `user_deleted === true` early-return in `processTransaction` (`src/core/decoder.ts`) turns the list assertion red.
- Suppressing the list assertion while keeping the mutation, the by-ID assertion (`count === 0`) **also** goes red on its own.

So neither assertion is riding on the other. I also checked that list mode applies no default date period, so the 2024-03-06 row isn't passing by accidentally falling outside a window — which is the failure mode that would have made this vacuous.

A nice side effect you may not have intended: seeding the deleted row gives the response-size ratchet in the same file incidental detection power too — under the mutation, the `get_transactions` budget assertion also goes red, because the leaked row inflates the response.

Also appreciated: the fixture uses a properly opaque Firestore-shaped ID rather than a trivial `id === name` value. That's a real convention here ([memory of a resolution bug that trivial fixtures masked](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/issues/609)) and it's easy to get wrong. Bringing `createCombinedDb` to parity with `createTransactionDb`, which already plumbed `user_deleted`, was precisely the gap.

One **entirely optional** follow-up, not a change request and not a reason to hold this: the `not.toContain` assertion could go vacuous under future fixture drift — if the ID constant and the seeded row ever diverged, it would still pass while testing nothing. A one-line companion assertion that a known-*active* sibling ID **is** present would pin that the query surface actually covers the fixture. Today it's provably non-vacuous, so this is hardening for later, not a gap now. Happy to open it as a separate good-first-issue if you'd rather not extend this PR.

Merging. Thanks for picking this one up — it's a small PR that leaves the invariant guarded at both the decoder and tool-surface layers.

PR #619, opened 2026-08-03
fix: make privacy comment stripping syntax-aware

### LINKED ISSUE #614: check-privacy-endpoints: block-comment stripper can be blinded by /* and */ inside separate string literals
Known edge case deliberately deferred from #594: `stripComments()` in `scripts/check-privacy-endpoints.ts` uses a non-greedy block-comment regex across the whole file, so a `'x/*'` in one string literal and a `'*/y'` in a later one swallow everything between them — hiding any URL literal in that span from the scan. Glob patterns like `'src/*'` are a plausible accidental trigger. Latent today (no such strings exist in `src/`).

**Work:** make the stripper string-aware (or tokenize minimally), add a regression case to the existing ten-case synthetic-tree test suite (`CHECK_PRIVACY_ENDPOINTS_ROOT` override pattern), and mutation-check it: reintroduce the naive regex and confirm the new test fails. Small, script + test only, no live credentials.

JiataiWang:
# Summary

- Replace whole-file regex comment stripping with TypeScript parser-owned comment ranges, preserving delimiter text inside strings, templates, and regex literals.
- Add synthetic-tree coverage for cross-literal delimiters and the template/regex siblings that a hand-written quote state can misclassify.

Closes #614

## External assumptions

None

## Bug fix?

Root cause: A whole-file block-comment regex treated delimiter-like characters inside literals as comment syntax.
Bug class: Source scanners inferring TypeScript comment boundaries without parser context.
Detector added: End-to-end synthetic-tree cases cover cross-literal delimiters across quote styles, escapes, templates, and regexes, plus actual comments inside template expressions and regex parser boundaries.
Siblings checked: Real block and line comments, single/double/template literals, escaped quotes, template expressions, regex quote/slash/block-delimiter forms, and the real repository endpoint scan.
Ledger updated: n/a — not an external-assumption bug.

## Testing

- `bun test tests/scripts/check-privacy-endpoints.test.ts` (14 passed)
- `bun run check` (2,598 passed; 20 integration tests requiring a real local database skipped)
- Mutation check: restoring the old regex makes the new cross-literal regression fail.

### DIFF
--- a/scripts/check-privacy-endpoints.ts
+++ b/scripts/check-privacy-endpoints.ts
@@ -45,6 +45,7 @@
 import { readFileSync, readdirSync, statSync } from 'fs';
 import { dirname, join } from 'path';
 import { fileURLToPath } from 'url';
+import ts from 'typescript';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 // CHECK_PRIVACY_ENDPOINTS_ROOT lets tests point the checker at a synthetic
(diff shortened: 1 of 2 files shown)


ignaciohermosillacornejo (approved):
Approving — and this is a stronger fix than #614 asked for.

The issue described one blinding input, and the tempting fix is a hand-rolled tokenizer that handles that case. You did the structurally right thing instead: delegate lexing to the TypeScript compiler (`ts.createSourceFile` + `getLeading`/`getTrailingCommentRanges` over an AST walk), so comment-ness is decided by the same grammar the runtime uses. That eliminates the "stripper disagrees with the language" bug *class* rather than patching one instance — which matters a lot more here than usual, because this is a privacy detector and its failure mode is silent.

Since a bug in this script can hide a real network destination from the scan, I reviewed it adversarially rather than just reading it: 17 hand-built inputs run through the actual checker against synthetic trees, each classified by **direction** of failure — a stripper bug that makes the scanner report *more* candidates is fail-safe (noisy), one that makes it report *fewer* is fail-open and is the thing that actually matters.

**Result: zero fail-open misses on any parseable file.** Correctly caught real traffic despite adversarial literals in nested-template `/*`…`*/` spanning, escaped-slash regexes (`/a\/*b/`), mid-line `//` inside strings, `String.raw` glob pairs, and URLs in template cooked text. Correctly ignored genuine comments in leading-JSDoc, trailing-same-line, inline-call-argument, and at-EOF-without-newline positions.

The residuals are all in the safe direction — shebangs, Annex B `<!--`/`-->`, and conflict markers aren't treated as comments, so URLs there survive into the scan and get reported. The one input that *hides* a URL is an unterminated `/*` running to EOF, and that's not fail-open in practice: such a file is a syntax error, can't clear the typecheck gate, can't execute, and per the lexical grammar its tail genuinely *is* comment-to-EOF. Worth noting the old regex would have reported it only because it required a closing `*/` — the new behavior is the semantically correct one.

Mutation-checked the tests too, per #596's standard: reverting `stripComments` to the naive two-regex version turns 2 of the 4 new tests red, and deleting the `getTrailingCommentRanges` collect line turns another red. No vacuous assertions.

Two things worth calling out because they were clearly deliberate: blanking comment ranges with **spaces** rather than deleting them preserves offsets and line structure, which keeps any future line-based diagnostics honest; and the dedupe map plus overlap guard make double-collection harmless by construction.

One informational note, no action needed: two of the four new tests ("a regex containing quotes…" and "a regex containing `//`…") pass even under the naive stripper, since the old line-comment regex was `^`-anchored and the old block regex stripped unconditionally. They're still worth keeping as regression pins on the new regex handling — just noting the mutation coverage comes from the other two.

`bun run check` is green on the branch, and `typescript` was already a devDependency, so no production dependency was added.

Merging. Thanks — this is careful work on exactly the kind of code where careful work is hard to see.

Issue #350, opened 2026-05-04
leveldb-reader: harden copy + fingerprint paths against compaction TOCTOU (ENOENT)


ignaciohermosillacornejo:
## Summary

Both `copyDatabaseToTemp`'s copy loop and the new `sourceFingerprint` helper (added in #349) iterate the source directory via `fs.readdirSync` and then call `fs.statSync` / `fs.copyFileSync` per file. There's a small TOCTOU window between listing and per-file access where Copilot's app could run a LevelDB compaction and delete a `.ldb` file that was just listed, causing `ENOENT`.

Flagged as a non-blocking observation by the AI reviewer on PR #349. Pre-existing in the copy loop; the new fingerprint path inherits the same race shape.

## Mechanism

```ts
// src/core/leveldb-reader.ts (sourceFingerprint, simplified)
for (const file of fs.readdirSync(srcPath)) {
  if (!isLevelDBFile(file)) continue;
  const m = fs.statSync(path.join(srcPath, file)).mtimeMs;  // can throw ENOENT
  if (m > max) max = m;
}
```

```ts
// src/core/leveldb-reader.ts (copy loop, simplified)
for (const file of fs.readdirSync(srcPath)) {
  if (isLevelDBFile(file)) {
    fs.copyFileSync(path.join(srcPath, file), path.join(tempDir, file));  // can throw ENOENT
  }
}
```

If the Copilot app deletes an old `.ldb` (compaction promoting/merging SSTables) between `readdirSync` and the per-file call, the per-file call throws `ENOENT` and the operation fails.

## Fix sketch

Wrap the per-file `statSync` / `copyFileSync` in a try/catch that silently skips entries that disappear:

```ts
try {
  const m = fs.statSync(filePath).mtimeMs;
  if (m > max) max = m;
} catch (err) {
  if ((err as NodeJS.ErrnoException).code !== 'ENOENT') throw err;
  // file vanished between readdir and stat — fine, skip it
}
```

Same shape for the copy loop.

For the fingerprint path, missing a vanished file is correct behavior (it's no longer part of the source's state). For the copy path, missing a deleted file is also correct (the resulting temp copy will simply lack a file LevelDB had already removed from its MANIFEST anyway).

## Production impact

**Low.** Compactions are infrequent (minutes to hours apart on a typical Copilot DB), and the TOCTOU window is microseconds. Has not been observed in practice — this issue tracks defensive hardening, not a known incident.

## Acceptance

[...]

Issue #345, opened 2026-05-03
leveldb-reader: tempDbCache returns stale snapshots for up to 5 minutes after source changes


ignaciohermosillacornejo:
## Summary

`copyDatabaseToTemp(srcPath)` in `src/core/leveldb-reader.ts` caches its temp-dir copy keyed only by source path with a 5-minute TTL. It does **not** check whether the source LevelDB has changed since the cache entry was created, so any caller that hits the cache within the TTL window reads from a stale snapshot.

Discovered while investigating #326 (PR #344) on 2026-05-03.

## Mechanism

```ts
// src/core/leveldb-reader.ts
const tempDbCache = new Map<string, TempDbCacheEntry>();

function copyDatabaseToTemp(srcPath: string): string {
  const cached = tempDbCache.get(srcPath);
  if (cached && fs.existsSync(cached.tempPath)) {
    cached.refCount++;
    cached.lastAccess = Date.now();
    return cached.tempPath;     // stale snapshot reused
  }
  // ... else mkdtempSync + copy + add to cache
}
```

After iteration ends, `releaseTempDatabase` decrements refCount; if 0, schedules cleanup in `TEMP_DB_CACHE_TTL = 5 * 60 * 1000` ms. Until the timer fires, the entry stays. There is no source-mtime comparison anywhere in the read path.

## Reproduction (synthetic, no live Copilot data needed)

```ts
import { createTestDatabase, LevelDBReader, iterateDocuments, cleanupAllTempDatabases } from '../src/core/leveldb-reader.js';

const dbPath = '/tmp/repro-leveldb-cache';
await createTestDatabase(dbPath, [
  { collection: 'transactions', id: 'txn-1', fields: { amount: 1.11, date: '2026-04-22' } },
]);

// First read populates the temp cache
let count = 0;
for await (const _ of iterateDocuments(dbPath)) count++;
console.log(count);  // 1 — correct

// Mutate the source LevelDB
const r = new LevelDBReader(dbPath);
await r.open();
await r.deleteDocument('transactions', 'txn-1');
await r.close();

// Second read returns the stale snapshot
count = 0;
for await (const _ of iterateDocuments(dbPath)) count++;
console.log(count);  // 1 — STALE (expected 0)

// Force-cleanup proves the bug source
cleanupAllTempDatabases();
count = 0;
for await (const _ of iterateDocuments(dbPath)) count++;
console.log(count);  // 0 — correct
```

## Production impact

[...]

PR #632, opened 2026-08-09
fix(leveldb-reader): never let temp DB copies outlive the process (#631)

### LINKED ISSUE #631: leveldb-reader: temp DB copies (`copilot-leveldb-*`) leak — deferred setTimeout cleanup never fires in short-lived server/worker (33GB observed)
### Summary

`copyDatabaseToTemp` copies the Copilot LevelDB (~120 MB) into `os.tmpdir()` as `copilot-leveldb-XXXXXX` on every read, but the cleanup is **deferred** via `setTimeout(…, TEMP_DB_CACHE_TTL)` (5 min). Because the MCP server typically runs as a **short-lived, per-request process** — and decoding happens in a **worker thread** (`dist/decode-worker.js`) that is terminated right after it posts its result — the process/worker almost always exits **before** the timer fires. Pending timers are dropped on exit, so the temp copy is never deleted.

### Impact

Every DB read leaks a ~120 MB temp directory. On a machine driving the server on a schedule (mine hits it ~6×/day), these accumulate without bound:

- Observed **366 orphaned `copilot-leveldb-*` dirs ≈ 33 GB** in `/private/var/folders/.../T`
- Filled the disk; nothing in the tool surface or docs hints that temp state is being left behind

### Root cause

`src/core/leveldb-reader.ts` → `releaseTempDatabase` (also bundled into the decode worker):

```js
function releaseTempDatabase(srcPath) {
  const cached = tempDbCache.get(srcPath);
  if (!cached) return;
  cached.refCount--;
  cached.lastAccess = Date.now();
  if (cached.refCount <= 0) {
    const scheduledTime = cached.lastAccess;
    setTimeout(() => scheduledCleanupCallback(srcPath, scheduledTime), TEMP_DB_CACHE_TTL); // never fires
  }
}
```

The reader's `finally` already does `await db.close(); releaseTempDatabase(dbPath);`, so by the time refCount hits 0 the DB is closed and the copy is safe to delete **now**. The 5-minute deferral only helps if the process stays alive that long — which the per-request server and the decode worker do not.

This is distinct from #345 (stale snapshots) and #350 (copy/fingerprint TOCTOU); both of those keep the deferred-cleanup design.

### Reproduce

```bash
# one short-lived read, then inspect tmp
BEFORE=$(ls -d "$(getconf DARWIN_USER_TEMP_DIR)"/copilot-leveldb-* 2>/dev/null | wc -l)
# ...run the MCP server, issue a single get_transactions, let the process exit...
AFTER=$(ls -d "$(getconf DARWIN_USER_TEMP_DIR)"/copilot-leveldb-* 2>/dev/null | wc -l)
echo "leaked: $((AFTER - BEFORE))"   # => 1 per invocation
```

### Suggested fix

Clean up **immediately** when the last reader releases (the DB is already closed), instead of scheduling:

```js
if (cached.refCount <= 0) {
  cleanupTempDatabase(cached.tempPath);
  tempDbCache.delete(srcPath);
}
```

A `process.on('exit', …)` backstop that purges `tempDbCache` handles the abandoned-generator case (refCount never reaches 0). If the 5-minute reuse window is worth keeping for long-lived sessions, gate the deferral on the server actually being long-lived, and always sweep on exit.

Verified locally against v2.3.0 with an isolated `TMPDIR` + a real LevelDB fixture: the immediate-cleanup change turns a reproducible 1-per-read leak into 0 residual, while reads still return correct data.

Happy to open a PR if useful.

papiluxo:
Fixes #631.

## The problem

`copyDatabaseToTemp` copies the LevelDB (~120 MB) into `os.tmpdir()` on every read, and `releaseTempDatabase` only *schedules* the cleanup:

```ts
setTimeout(() => scheduledCleanupCallback(srcPath, scheduledTime), TEMP_DB_CACHE_TTL); // 5 min
```

The server usually runs as a **short-lived, per-request process**, and decoding runs in a **worker thread** that exits as soon as it posts its result. Both are gone long before that timer can fire, and pending timers are dropped on exit — so the temp copy is never deleted.

Observed here: **366 orphaned `copilot-leveldb-*` directories ≈ 33 GB**, which filled the disk.

## Part 1 — stop new leaks

The obvious fix — delete immediately when `refCount` hits 0 — works, but it throws away the reuse cache: each subsequent read of the same DB re-copies ~120 MB, and it breaks the contract asserted in `tests/core/leveldb-reader.test.ts` (`cleanupAllTempDatabases` expects the cache to still be populated *after* iteration).

So this PR keeps the cache and its reuse semantics exactly as they are, and instead ties a copy's lifetime to the process:

1. **Sweep `tempDbCache` on `'exit'`** — covers worker exit, stdin close, and explicit `process.exit()`. `rmSync` is synchronous, which `'exit'` requires.
2. **Sweep on `SIGINT`/`SIGTERM`/`SIGHUP`, then re-raise** — a signal terminates the process *without* emitting `'exit'`. The listener is registered with `once`, so by the time we re-raise it has been removed, the default disposition applies, and the conventional `128+n` exit code is preserved.
3. **`unref()` the TTL timer** — a pending cleanup should never keep a finished server or worker alive for the full 5 minutes. (It also means the timer will usually never fire, which is precisely why the exit sweep, not the timer, is what guarantees deletion.)

No behavior change for callers: same reuse, same fingerprint-based invalidation (#345), same TOCTOU handling (#350). Existing tests are untouched.

## Part 2 — reclaim what already leaked

Part 1 stops the bleeding but does nothing for anyone already sitting on hundreds of orphans, so there's now a best-effort background sweep that removes `copilot-leveldb-*` directories untouched for over an hour.

Deleting directories owned by other processes deserves scrutiny, so the safety argument is explicit:

- **Copies this process still has cached are skipped** by path.
- **Deleting another process's idle copy is harmless** — `copyDatabaseToTemp` already guards every reuse with `fs.existsSync` and simply re-copies when the directory is gone. Worst case for another process is one extra copy, never a failed read.
- **`touchTempCopy` refreshes mtime on every reuse**, so "untouched for an hour" means genuinely idle rather than merely old. Without this, a long-lived process reusing one copy for hours could have had it swept mid-use; with it, a sweepable copy is one nobody has read from in an hour (12× `TEMP_DB_CACHE_TTL`).
- **Never blocks anything** — unref'd timer, async `rm`. It can't delay startup or stall a read; a per-request process may exit before the sweep starts, or mid-pass, and simply finishes the job next run.
- **Opt out** with `COPILOT_MCP_NO_TEMP_SWEEP=1`.

## Part 3 — close the class at the source (review follow-up)

@ignaciohermosillacornejo found one more member of the uninterceptable class while reviewing: `copyDatabaseToTemp` runs `mkdtempSync` and only writes to `tempDbCache` *after* the copy loop, so a real mid-copy failure (EACCES/EIO — not the ENOENT the loop deliberately skips) stranded a directory that no in-process cleanup path could name. Pre-existing on `main`, and it's why the suite stranded one small directory per run through the `copy loop propagates non-ENOENT copy errors` test.

The copy loop now removes its own directory before rethrowing, so the sweep is a backstop for that case rather than the only reclaim path. The failure still propagates — silently returning a half-copied database would be a far worse bug than the stranded directory.

## Review nits, all applied

- `runOrphanSweep` was exported as "@internal exported for tests" but nothing imported it. **Unexported** — the child-process tests cover it end to end.
- `COPILOT_MCP_NO_TEMP_SWEEP` now compares **`=== '1'`**, matching `COPILOT_DISABLE_PERSISTENT_INDEX`. Under truthiness, `=0` also disabled the sweep.
- Both **overclaimed doc comments** corrected: the `128+n` preservation is now stated as holding when nothing else listens for the signal (true for every caller today, with the reason recorded), and the sweep's "never keeps the process alive" is now scoped to the unref'd timer — once a pass has started, its awaited fs promises do hold the event loop until it finishes, which the comment now says explicitly, including that the first post-upgrade run against a large backlog can briefly delay exit.
- `bun run fix` applied; `bun run format:check` and `bun run typecheck` are clean.

## Tests

`tests/core/leveldb-reader-temp-cleanup.test.ts` covers the property a same-process unit test structurally cannot observe: **once the reading process is gone, no temp copy remains.** Each child runs with its own `TMPDIR`, so assertions are exact and can't collide with a real server's copies.

- exits normally → no `copilot-leveldb-*` left (asserts exit code 0 first, so "no copy" can't pass vacuously)
- killed with `SIGTERM` → asserts the copy **exists** while running, then that it's gone after exit
- sweep → two 2-hour-old orphans removed, while a recently-touched copy, an unrelated directory, and the child's own in-use copy all survive
- opt-out → child is deliberately kept alive past the point where the enabled sweep demonstrably finishes, so a surviving orphan can only mean the env var was honoured
- **new:** a mid-copy EACCES leaves nothing behind. This one runs in-process by pointing `TMPDIR` at an empty directory — `os.tmpdir()` re-reads the variable on every call — so "nothing was left behind" is an exact assertion rather than a search of the real temp dir.

**Mutation-checked, as requested.** Against `main`'s `leveldb-reader.ts`, **4 of the 5 fail**; the fifth is the opt-out test, which is vacuous against a build with no sweep to disable (inherent to opt-out tests, and it was already rewritten to be non-vacuous against the fixed build). The Part 3 test was checked the same way against this branch: delete the `cleanupTempDatabase(tempDir)` line and it goes red naming the stranded directory.

I verified the mechanisms directly rather than assuming them: that an `unref()`'d timer doesn't hold a worker open, that `process.on('exit')` still fires and can synchronously remove the directory on natural worker exit, that the signal path sweeps, and that the sweep's predicate discriminates correctly across seven cases (2 h and 25 h orphans swept; 59-minute, fresh, in-cache, foreign-prefix, and not-a-directory entries all left alone).

`SIGKILL` and `worker.terminate()` (the decode-timeout path) remain uninterceptable by definition — the orphan sweep is what eventually reclaims those.

## Post-mortem

`docs/bugs/631-temp-copy-deferred-cleanup.md`, filed under a proposed new class **`deferred-cleanup-never-runs`** (added to `docs/bugs/README.md`): *releasing a resource is deferred to a timer or callback that only fires if the process outlives it, in a process that routinely exits first.* Nothing in the existing list fit — it isn't `stale-cache-semantics` or `silent-failure-masking` — and per the README the third instance of a class is the finding, so it seemed better to name it now than to file this under something adjacent.

The `Detector` field says **none — instance-only**, honestly: the class is defined by a deployment property (does this process outlive its own timers?) that no static check in this repo can evaluate, and an invented gate would be worse than the admitted gap. The entry also records *why the existing tests were blind* — they called `_runScheduledCleanup` directly, doing for the code the one thing production never did.

## Two things found while doing this, neither fixed here

1. **`tests/context-budget.test.ts` fails on any machine that has run the scheduled drift check.** `get_connection_status` embeds `readScheduledSmokeStatus()`, which reads `~/.claude/copilot-money/scheduled-smoke.json` — including the full `report` string. Mine is 1330 chars, which pushes the response to 2664 against a 1900 budget. Move that file away and the suite is 68/68. It reproduces identically on `main` (so it is not from this PR), but it's arguably two real findings: a supposedly hermetic test reads real home-directory state, and the #597 budget a user actually experiences can be blown by a file the budget test normally never sees.
2. **One adjacent stranded-artifact sibling**, not of the timer class: `transaction-meta-store.ts`'s cap valve writes `${file}.${pid}.tmp` and renames it — if `writeFileSync` succeeds and `renameSync` throws, the `.tmp` stays. Bounded and tiny (one small file per pid), so I left it alone, but happy to close it if you want the shape gone everywhere.

Full suite on this branch: **2652 pass, 20 skip, 1 fail** — that one failure being (1) above.

## External assumptions

None — this PR touches only local temp-file lifecycle and process-exit handling. It makes no new assumption about Copilot's API, its data shapes, or its on-disk format.

## Bug fix?

```text
Root cause:       temp-copy deletion was deferred to a 5-minute setTimeout that never fires — the MCP server is a short-lived per-request process and the decode worker exits milliseconds after posting its result, and pending timers are dropped on exit
Bug class:        deferred-cleanup-never-runs (new — proposed and defined in docs/bugs/README.md; releasing a resource is deferred to a timer/callback that only fires if the process outlives it, in a process that routinely exits first)
Detector added:   none — instance-only regression tests (tests/core/leveldb-reader-temp-cleanup.test.ts, child-process lifetime assertions, mutation-verified against main). The class is a deployment property no static check here can evaluate; an invented gate would be worse than the admitted gap
Siblings checked: every setTimeout/setInterval in src/ — decoder.ts:3369 (decode timeout, cleared in settle(), no resource attached) and graphql/client.ts:226 (retry backoff, not cleanup). Neither belongs to the class. Also audited every on-disk artifact creator in src/: found one adjacent stranded-artifact case in transaction-meta-store.ts (cap-valve .tmp if rename throws), reported above, not fixed here
Ledger updated:   n/a — not an external-assumption bug
```

### DIFF
--- a/docs/bugs/631-temp-copy-deferred-cleanup.md
+++ b/docs/bugs/631-temp-copy-deferred-cleanup.md
@@ -1,0 +1,121 @@
+---
+id: 631
+title: Every LevelDB read left its ~120 MB temp copy on disk — 366 stale directories (~33 GB) filled a user's disk
+class: deferred-cleanup-never-runs
+status: fixed
+detected: user-report  # a user's disk filled; they traced the `copilot-leveldb-*` pile back to this module and filed #631 with the mechanism
+fixed_in: https://github.com/ignaciohermosillacornejo/copilot-money-mcp/pull/632
+issue: https://github.com/ignaciohermosillacornejo/copilot-money-mcp/issues/631
+date: 2026-08-12
+---
+
+## Symptom
+
+`os.tmpdir()` accumulated one `copilot-leveldb-*` directory per read, forever. Each is a
+full copy of the Copilot LevelDB (~120 MB). On the reporting machine, automation that hit
+the MCP a handful of times a day for several weeks produced **366 directories, ~33 GB**,
+and the disk filled — at which point unrelated work started failing with `ENOSPC` while
+the server itself still appeared healthy.
+
+Nothing in the tool's own output was wrong. The failure surfaced somewhere else entirely
+(a full disk), which is why it went unnoticed for weeks despite being trivially visible in
+`ls "$TMPDIR"`.
+
+## How it was detected
+
+Not by anything this project built. A user ran out of disk space, went looking for what
+was eating it, found the pile of `copilot-leveldb-*` directories, and read the source to
+work out why they were never deleted. The issue arrived with the mechanism already
+diagnosed.
+
+Worth recording plainly: the module had a passing test suite covering exactly this
+lifecycle — refcounting, TTL expiry, the cleanup callback — and none of it could see the
+bug, because all of it ran in a process that never exited during the test.
+
+## Root cause
+
+`releaseTempDatabase` (`src/core/leveldb-reader.ts`) deferred deletion to a timer:
+
+```ts
+if (cached.refCount <= 0) {
+  setTimeout(() => scheduledCleanupCallback(srcPath, scheduledTime), TEMP_DB_CACHE_TTL); // 5 min
+}
+```
+
+The design assumed a long-running host. Neither real host is one: the MCP server is
+normally spawned per request and exits when stdin closes, and the actual decode runs in a
+worker thread (`dist/decode-worker.js`) that is done milliseconds after it posts its
+result. Node drops pending timers on exit, so the 5-minute callback essentially never ran
+— and it was the only thing that deleted anything.
+
+Two aggravating details, both fixed in the same PR:
+
+- The TTL timer was not `unref()`'d, so on the main-thread path it kept the *worker*
+  alive for the full 5 minutes after its work was done — holding the V8 isolate and
+  native LevelDB memory that worker isolation exists to release promptly.
+- `copyDatabaseToTemp` creates the directory with `mkdtempSync` and only records it in
+  `tempDbCache` after the copy loop succeeds. A real mid-copy failure (EACCES/EIO) left a
+  directory that no in-process cleanup path could name. Small, but it is the same class
+  reached from a different direction, and the repo's own test suite was stranding one
+  directory per run through it.
+
+## Why the tests didn't catch it
+
+`tests/unit/signal-timer-coverage.test.ts` and the reader's unit tests exercised the
+cleanup thoroughly — by calling `_runScheduledCleanup` directly, i.e. by doing for the
+code the one thing production never did. They asserted "when the callback runs, the
+directory is deleted", which was true and always had been. The false assumption was not in
+the assertion but in the harness: a same-process test cannot observe that its process
+would have exited before the timer fired.
+
+The property that was actually broken — *no temp copy outlives the process that made it* —
+is not expressible in a same-process test at all. It needs a child process, which is why
+the regression tests spawn one.
+
+## The fix
+
+[PR #632](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/pull/632), in
+three parts:
+
+- **Root cause:** a lazily-registered `process.on('exit')` sweep, plus `once` handlers for
+  SIGINT/SIGTERM/SIGHUP that sweep and re-raise, so deletion is tied to process exit
+  rather than to a timer. The TTL timer stays (it preserves the reuse cache that avoids
+  re-copying ~120 MB) but is now `unref()`'d and is no longer load-bearing.
+- **Reclaim:** a best-effort orphan sweep on startup removes `copilot-leveldb-*`
+  directories untouched for an hour, since the exit sweep does nothing for the users
+  already sitting on tens of GB — and it remains the only reclaim path for genuinely
+  uninterceptable deaths (SIGKILL, `worker.terminate()` on a decode timeout).
+- **Downstream cleanup:** the copy loop now removes its own directory before rethrowing.
+
+## Detector
+
+**None — instance-only regression tests**, and the honest reason is that the class is
+defined by a *deployment* property (does this process outlive its own timers?) that no
+static check in this repo can evaluate. `tests/core/leveldb-reader-temp-cleanup.test.ts`
+asserts the property for this module only: spawn a child, complete a real read, kill it
+normally and by signal, assert an isolated `TMPDIR` is empty afterwards.
+
+**Mutation-verified.** Against `main`'s reader, 4 of the 5 tests fail; the fifth is the
+`COPILOT_MCP_NO_TEMP_SWEEP` opt-out test, which is vacuous against a build that has no
+sweep to disable and was rewritten during review to be non-vacuous against the fixed
+build. The copy-loop test was checked the same way: delete the `cleanupTempDatabase(tempDir)`
+line and it goes red naming the stranded directory.
+
+The cheapest real gate for the class would be a review question rather than code: *does
+anything here schedule cleanup on a timer, and is this process guaranteed to be alive when
+it fires?* Both existing `setTimeout` siblings in `src/` were audited against it
+(`decoder.ts`'s decode timeout, cleared in `settle()` with no resource attached;
+`graphql/client.ts`'s retry backoff) and neither belongs to the class.
+
+## Lesson
+
+A resource whose release is scheduled rather than performed is only as reliable as the
+process's remaining lifetime — and a per-request server plus a fire-and-exit worker have
+approximately none. Tie cleanup to an event the runtime guarantees (`exit`, a `finally`,
+the release call itself), and treat any deferred cleanup as unreliable unless something
+else also guarantees it.
+
+The testing lesson is sharper and more general than the bug: **a lifecycle property that
+only manifests at process death cannot be tested in-process.** The suite here was not thin;
+it was measuring the right thing in a harness where the thing could not go wrong. Spawning
+a child costs a few seconds and was the only way to see it.
--- a/docs/bugs/README.md
+++ b/docs/bugs/README.md
@@ -94,11 +94,12 @@ a class yet.
 | `stale-kill-switch` | A temporary disablement outlives the condition it guarded and silently disables a restored feature. | none |
 | `packaging-environment-mismatch` | The shipped artifact works in dev but fails in the target runtime — missing bundled deps, hardened-runtime constraints, GUI PATH. | `tests/integration/mcpb-bundle.test.ts` (extract-and-boot outside the repo) |
 | `unsettled-promise` | An async operation has a completion path where neither resolve nor reject fires, hanging callers forever. | none |
+| `deferred-cleanup-never-runs` | Releasing a resource is deferred to a timer or callback that only fires if the process outlives it — in a process that routinely exits first. The resource is acquired eagerly and released never, while every same-process test still observes correct bookkeeping. | none — instance-only; the class is a deployment property (does this process outlive its own timers?) that no static check here can evaluate |
 
 ## How we find bugs
 
 Recorded per entry, using a fixed vocabulary so the corpus stays countable. Here is what
-this corpus actually says, across all 43 entries:
+this corpus actually says, across all 44 entries:
 
 | Found by | Count | |
 |---|---|---|
@@ -106,7 +107,7 @@ this corpus actually says, across all 43 entries:
 | `live-probe` — a probe or smoke against the real backend | 7 | ██████ |
 | `audit-sweep` — a deliberate cross-cutting audit | 7 | ██████ |
 | `incidental` — found while working on something else | 6 | █████ |
-| `user-report` | 5 | ████ |
+| `user-report` | 6 | █████ |
 | `adversarial-review` — a reviewer tried to refute a claim or mutation-tested a guard | 2 | █ |
 | `detector-first` — a detector was built, and then found bugs | 1 | ▌ |
 | `code-review` | 1 | ▌ |
@@ -287,3 +288,9 @@ record near-misses.
 | | Bug | Found by | Date |
 |---|---|---|---|
 | #129 | [Decode-worker promise hangs forever if the worker exits without sending a result](129-worker-exit-promise-hang.md) | `code-review` | 2026-03-12 |
+
+**`deferred-cleanup-never-runs`** — 1
+
+| | Bug | Found by | Date |
+|---|---|---|---|
+| #631 | [Every LevelDB read left its ~120 MB temp copy on disk — 366 stale directories (~33 GB) filled a user's disk](631-temp-copy-deferred-cleanup.md) | `user-report` | 2026-08-12 |
--- a/src/core/leveldb-reader.ts
+++ b/src/core/leveldb-reader.ts
@@ -280,8 +280,12 @@ function registerTempDbExitSweep(): void {
 
   // A signal terminates the process WITHOUT emitting 'exit'. Sweep, then
   // re-raise: the listener is registered with `once`, so by the time we
-  // re-raise it has already been removed and the default disposition applies,
-  // preserving the conventional 128+n exit code.
+  // re-raise it has already been removed and the default disposition applies.
+  // That restores the conventional 128+n exit code as long as this is the only
+  // listener for the signal — true for every caller today, since the reader
+  // runs in the decode worker and standalone scripts, never in the server
+  // process that installs its own SIGINT/SIGTERM handlers. A host that does
+  // install its own keeps its handling, not ours; the sweep still runs.
   for (const signal of SWEEP_SIGNALS) {
     process.once(signal, () => {
       cleanupAllTempDatabases();
@@ -316,14 +320,21 @@ let orphanSweepStarted = false;
  *     is one nobody has read from in an hour.
  *
  * Best effort by design: it runs on an unref'd timer with async removal, so it
- * never delays startup, blocks a read, or keeps a short-lived process alive. A
- * per-request process may exit mid-sweep and simply finish the job next run.
- * Set COPILOT_MCP_NO_TEMP_SWEEP=1 to disable.
+ * never delays startup and never blocks a read. The unref'd timer also means a
+ * process that finishes first simply exits without sweeping at all. Once the
+ * pass has started, though, its awaited fs promises do hold the event loop
+ * until it finishes — so the first run after upgrading, with a large backlog to
+ * delete, can briefly delay exit. That is the intended trade-off: the backlog
+ * is the whole reason this exists, and a per-request process interrupted
+ * mid-pass just finishes the job on the next run.
+ *
+ * Set COPILOT_MCP_NO_TEMP_SWEEP=1 to disable (`1` exactly, matching
+ * COPILOT_DISABLE_PERSISTENT_INDEX).
  */
 function startOrphanSweep(): void {
   if (orphanSweepStarted) return;
   orphanSweepStarted = true;
-  if (process.env.COPILOT_MCP_NO_TEMP_SWEEP) return;
+  if (process.env.COPILOT_MCP_NO_TEMP_SWEEP === '1') return;
 
   const timer = setTimeout(() => {
     void runOrphanSweep();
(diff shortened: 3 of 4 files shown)


ignaciohermosillacornejo (commented):
Thanks for this — it's an excellent PR. The diagnosis is right, the writeup made it easy to verify, and the tests are the kind that actually prove the property (process-lifetime assertions, not same-process unit tests). I reviewed it hands-on rather than from the diff alone; details below.

**Design verdict: agreed on both parts, as designed.** Part 1's cache-preserving approach (exit sweep + unref'd timer) is the right call over immediate cleanup — it keeps the reuse semantics and ties deletion to the thing that actually guarantees it. And to answer the offer at the end of your description directly: please keep Part 2. The safety argument holds up (prefix scoping, own-cache exclusion, idle threshold + touch-on-reuse, and every reuse path already self-heals via re-copy), and the sweep is more than a migration shim — it's the only reclaim path for the genuinely uninterceptable cases (SIGKILL, `worker.terminate()` on decode timeout, and one more found below). Default-on with the env opt-out is the right product call too: the users sitting on tens of GB are the least likely to find an opt-in flag.

## What I verified locally

- Your 4 new tests pass on this branch; swapping in main's `leveldb-reader.ts`, 3 of 4 fail (the opt-out test passes vacuously against main, which is inherent to opt-out tests). Full suite: 2672 pass, 0 fail.
- **The fix works on the real production path, which the tests don't exercise.** The tests spawn child *processes*; production runs the reader inside a Node *worker thread* (`dist/decode-worker.js`). I drove the built worker bundle directly under Node 25: on this branch the copy exists during decode, the worker exits ~3 ms after posting its result, and the exit sweep leaves 0 copies. On main's bundle, the worker was still alive 30+ s after the result (held by the non-unref'd TTL timer) and the copy orphaned when the parent exited.
- Which means this PR quietly fixes a **second bug worth crediting**: on main, every decode worker lingers ~5 minutes per decode, holding exactly the V8 isolate / native LevelDB memory that worker isolation exists to free promptly — and a lingering ref'd worker holds the parent process too. The `unref()` fixes both.
- The signal re-raise design is correct for every real caller today: in production the reader only runs inside the decode worker (only `decoder.ts` and standalone scripts import it), so the handlers never collide with `server.ts`'s own SIGINT/SIGTERM listeners.

The follow-up commit making the opt-out test non-vacuous deserves a special callout — "a test that can't fail proves nothing" is a discipline this repo has been burned on before, and it's nice to see it arrive pre-applied.

## Required before merge (both chores, not design issues)

1. **`bun run check` fails on Prettier** — one line in the new test file (the `child.on('exit', …)` promise around line 106 wants joining). `bun run fix` resolves it.
2. **The `required-sections` CI gate will reject the current PR body.** The title matches `fix(...):`, so the body must contain a `## External assumptions` section and the five Bug Response Ritual labels verbatim (see CONTRIBUTING.md). Paste-ready, feel free to edit:

```markdown
## External assumptions

None — this PR touches only local temp-file lifecycle; no new assumptions about Copilot's API or data.

## Bug fix?

Root cause:       temp-copy cleanup was deferred via a 5-minute setTimeout that never fires — the decode worker / per-request process exits first, and pending timers are dropped on exit
Bug class:        timer-deferred resource cleanup that assumes a long-lived process
Detector added:   tests/core/leveldb-reader-temp-cleanup.test.ts — child-process lifetime tests asserting no copilot-leveldb-* dir survives process exit (normal exit and SIGTERM), plus sweep-discrimination tests
Siblings checked: decoder.ts decode-timeout timer (cleared in settle(); no resource attached); graphql/client.ts setTimeout (retry backoff, not cleanup) — none found
Ledger updated:   n/a — not an external-assumption bug
```

## Non-blocking nits

- `runOrphanSweep` is exported with "@internal exported for tests", but nothing imports it — the child-process tests cover it end-to-end. Drop the export or the comment (or add a direct unit test, your pick).
- The repo's existing kill-switch convention is `=== '1'` (`COPILOT_DISABLE_PERSISTENT_INDEX`); `COPILOT_MCP_NO_TEMP_SWEEP` uses truthiness, so `=0` also disables the sweep. Worth aligning.
- Two doc-comment claims are very slightly overclaimed, wording tweaks optional: the 128+n preservation on re-raise only holds when no other signal listeners exist (true for all current callers), and once the orphan sweep's first fs op starts, in-flight awaited fs promises do keep the event loop alive until it finishes — so the first post-upgrade run deleting a large backlog will hold process exit for its duration (arguably desirable).

## One more member of the uninterceptable class, found while reviewing

`copyDatabaseToTemp` runs `mkdtempSync` and then the copy loop — if the loop throws (real EACCES/EIO mid-copy), the fresh dir exists on disk but never reached `tempDbCache`, so neither the exit sweep nor `cleanupAllTempDatabases` can see it. Pre-existing on main, not introduced here — and it's why the repo's own test suite strands one tiny dir per run (traced to the `copy loop propagates non-ENOENT copy errors` test). Your orphan sweep mops these up after an hour, which is a good argument for Part 2 all by itself, but closing it at the source (try/catch around the copy loop that removes the fresh dir before rethrowing) would complete the class.

Would you be up for taking that on — either in this PR or as a separate follow-up? Both are equally welcome. If it's not something you want to pick up, no problem at all, we'll take it.

Happy to merge once the two chores above are in.

ignaciohermosillacornejo:
Thanks for this — a 33GB leak is a serious find, and the diagnosis in #631 (deferred `setTimeout` cleanup never firing in a short-lived process) is exactly right. The scheduled-cleanup design assumed a long-running host, which the MCP server and the decode worker are not.

Two mechanical things are blocking CI, both trivial:

**1. Formatting.** `bun run format:check` is failing. `bun run fix` will sort it.

**2. Required PR sections.** The gate wants two things in the PR body:

```markdown
## External assumptions

None — this changes process-local temp-file lifecycle and makes no new
assumption about Copilot's API or data.
```

plus the Bug Response Ritual block, since the title is a `fix:` ([CONTRIBUTING.md](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/blob/main/CONTRIBUTING.md#bug-response-ritual)):

```text
Root cause:       <the mechanism — deferred cleanup that never runs>
Bug class:        <see docs/bugs/README.md; propose a new one if none fits>
Detector added:   <the class-level gate, or "none — instance-only regression test">
Siblings checked: <other deferred-cleanup paths audited, or "none found">
Ledger updated:   <n/a for non-API bugs>
```

On `Bug class` — we recently published [`docs/bugs/`](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/tree/main/docs/bugs), a post-mortem knowledge base organised by class. Worth a look before naming one. I don't think this fits an existing class cleanly; "cleanup scheduled on a timer that outlives nothing" may deserve its own. `Detector added: none — instance-only regression test` is a fine answer if that's the truth — we'd much rather have an honest gap than an invented gate.

For `Siblings checked`: the question worth asking is whether any *other* `setTimeout`-deferred work in the codebase has the same assumption. That's the part that turns this from a fix into a ratchet.

Say the word if you'd rather I just push the formatting fix and fill the sections in myself — happy to, I only held off because the sections are meant to be the author's account of their own change.

One substantive note for when CI is green: I'd like the regression test to be **mutation-checked** — revert the fix, confirm the test actually goes red. We got burned by a test that executed a guard without being able to fail on its deletion ([#596](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/issues/596)), so it's become the local standard for anything guarding a resource leak.

papiluxo:
Thanks — that's the most useful review I've had on an outside PR, and driving the built worker bundle directly rather than reading the diff caught something I'd missed entirely (the lingering-worker cost of the non-unref'd timer). Everything you asked for is now pushed, plus the follow-up you offered.

**Yes to the copy-loop case, and I took it in this PR** (Part 3 in the body). `mkdtempSync` creates the directory before the loop and the cache entry is only written after it, so a real EACCES/EIO mid-copy left a directory nothing in-process could name — the loop now removes it before rethrowing. The error still propagates: silently returning a half-copied database would be a much worse bug than the stranded directory. Nice side effect, as you predicted: the suite no longer strands one directory per run through the `copy loop propagates non-ENOENT copy errors` test.

**The two blockers:**

1. `bun run fix` applied. `bun run check` passes in full locally — 2653 pass, 20 skip, 0 fail.
2. PR body rewritten with `## External assumptions` and the ritual block. I ran it through `scripts/check-pr-sections.sh` with the real title before pushing, so the gate should be satisfied.

On `Bug class` — I read `docs/bugs/` and agree nothing there fits, so I proposed **`deferred-cleanup-never-runs`** with a definition sharp enough to sort future bugs: *releasing a resource is deferred to a timer or callback that only fires if the process outlives it, in a process that routinely exits first.* Post-mortem is in `docs/bugs/631-temp-copy-deferred-cleanup.md`. `Detector` says **none — instance-only**, honestly: the class is a deployment property (does this process outlive its own timers?) that no static check here can evaluate, and per your own README an invented detector is worse than an admitted gap. For `Siblings checked` I walked every `setTimeout`/`setInterval` in `src/` — `decoder.ts`'s decode timeout (cleared in `settle()`, no resource attached) and `graphql/client.ts`'s retry backoff — and neither belongs to the class. Rename or reclassify freely; you know the corpus better than I do.

**Mutation-checked, as you asked.** Against `main`'s `leveldb-reader.ts` 4 of the 5 tests fail — the fifth being the opt-out test, vacuous against a build with no sweep to disable. The new copy-loop test got the same treatment against this branch: delete the `cleanupTempDatabase(tempDir)` line and it goes red naming the stranded directory.

**Nits, all applied:** `runOrphanSweep` unexported; `COPILOT_MCP_NO_TEMP_SWEEP` now compares `=== '1'` to match `COPILOT_DISABLE_PERSISTENT_INDEX`; and both doc claims corrected — you were right on both, including that a started sweep does hold the event loop through its awaited fs promises, which the comment now states along with the first-run-after-upgrade consequence.

**Two things I found while working, neither fixed here:**

1. **`tests/context-budget.test.ts` fails on any machine that has run the scheduled drift check.** `get_connection_status` embeds `readScheduledSmokeStatus()`, which reads `~/.claude/copilot-money/scheduled-smoke.json` — including the whole `report` string. Mine is 1330 chars, which takes the response to 2664 against a 1900 budget. Move that file aside and it's 68/68. It reproduces identically on `main`, so it isn't from this PR, but it looks like two separate findings: a supposedly hermetic test reading real home-directory state, and a #597 budget that a real user's machine can blow in a way the budget test normally can't see. Happy to open an issue if it's news.
2. **One adjacent stranded-artifact sibling**, not of the timer class: `transaction-meta-store.ts`'s cap valve writes `${file}.${pid}.tmp` then renames — if the write succeeds and the rename throws, the `.tmp` stays. Bounded and tiny, so I left it, but say the word and I'll close the shape everywhere.

Last thing: the workflows on the new head are sitting at `action_required` — fork PRs need your click to run. Nothing has actually executed yet, so the red you may see is "not started", not "failed".

---

**Edit — filed as #638, with one correction to what I wrote above.** I said the budget "can be blown on a real user's machine"; that needs a caveat I hadn't checked yet when I wrote it. The oversized `scheduled-smoke.json` on my machine was written by *my own* launchd job, not by `scripts/scheduled-smoke.ts` — mine puts the full report text in `report` where yours puts a file path. So your writer keeps that response well inside 1900, and the overage is my writer's fault, which I'll fix on my side. The two findings in #638 hold either way: the budget test reads real `$HOME` state (and the read side has no path override, so no test can isolate it), and `report` is unbounded in a budgeted response by schema rather than by accident of my setup.

ignaciohermosillacornejo (approved):
Approving — everything from the review round is in and I re-verified it hands-on rather than from the diff:

- `bun run check` passes in full here (2653 pass / 20 skip / 0 fail, matching your numbers).
- Part 3 is correct (error still propagates; the directory doesn't), and I ran your mutation check myself: deleting the `cleanupTempDatabase(tempDir)` line turns the new test red.
- All four nits landed properly — including the two doc-comment rewrites, which now state the real semantics rather than softening them.
- The post-mortem and the `deferred-cleanup-never-runs` class are exactly what `docs/bugs/` is for, and **`Detector: none — instance-only` is accepted as the right answer**: the class is a deployment property, and your "review question" framing of the cheapest real gate is going to get used.

One correction to the record, found by measurement: "the suite no longer strands one directory per run" is true for the copy-loop path specifically, but a full `bun test` run still strands ~38 tiny dirs on this branch — bun's test runner never fires `process.on('exit')` listeners (verified with a marker probe), so the exit sweep can't run at suite end. Pre-existing on `main`, production unaffected (Node and plain `bun script.ts` both fire `'exit'`, as your own child-process tests demonstrate), and your orphan sweep reclaims them an hour later — Part 2 quietly ends up being the fix for dev machines too. Noting it here so the claim in the thread doesn't outlive its truth; no action needed in this PR.

On your two findings: #638 is a good issue — both points verified here, and the disclosure section is a model for how to file a finding your own setup influenced. And **yes please** to the `transaction-meta-store.ts` `.tmp` sibling as a follow-up PR whenever you like; it's yours.

Workflows are approved and running. Thanks for a genuinely excellent contribution arc — issue with the mechanism diagnosed, fix with the class closed, post-mortem filed.

PR #636, opened 2026-08-12
docs: correct contributing setup and test guidance

### LINKED ISSUE #634: docs: CONTRIBUTING.md drift — wrong hook name, dead bun test --filter example, Node prerequisite overstated
Two claims in `CONTRIBUTING.md` are factually wrong (both verified against the repo), plus two smaller wording improvements. All are self-contained doc edits — a good first contribution; no knowledge of the decoder or GraphQL surface needed.

### 1. Wrong hook name in "Code Style"

`CONTRIBUTING.md` says:

> ESLint + Prettier enforced via pre-commit hooks

The only hook in `.husky/` is **`pre-push`**, which runs `bun run check`. This is deliberate policy (gate what leaves the machine; keep local commits fast for TDD red-state work), so the doc contradicts the design, not just the implementation.

**Fix:** change to something like "ESLint + Prettier enforced via a pre-push hook (`bun run check`)."

### 2. Dead command example in "Testing"

`CONTRIBUTING.md` says:

```bash
bun test --filter "getBalanceHistory"        # Pattern match
```

`--filter` is not a test-name flag — running this searches 560 files and executes **zero tests**. The working flag is `-t` / `--test-name-pattern`:

```bash
bun test -t "getBalanceHistory"   # runs 38 matching tests today
```

**Fix:** replace `--filter` with `-t` in the example.

### 3. "Bun (latest) or Node.js 18+" overstates Node

Development and tests require **bun** specifically (`bun:test` runner, bun-based scripts); Node alone can only run the built artifact. Node 18 is also long EOL. **Fix:** reword Prerequisites to make bun required for development, and mention Node only as a runtime for the built server (bumping the stated floor is optional and can be left to a separate `engines` decision).

### 4. (Optional — maintainer call) `Verification:` line in the PR template

`.github/PULL_REQUEST_TEMPLATE.md` has Summary, External assumptions, and the Bug Response Ritual, but nothing asks what was actually run. A one-line `Verification:` field (e.g. "commands run and their results — `bun run check`, targeted tests, live smoke") would match the repo's evidence-class culture and surface "didn't run `bun run check`" before CI does. If this is adopted, note that `scripts/check-pr-sections.sh` intentionally only enforces the existing sections — deciding whether the new field should be gated there too is part of this item.

### Context

Found while reviewing PR #632, where a first-time contributor's push failed `bun run check` in CI — the pre-push hook only exists in a clone after `bun install` triggers husky's `prepare`, so docs are the first line of defense for fork contributors. Items 1–3 are pure text edits; item 4 touches the template and optionally the gate script + its tests (`tests/scripts/check-pr-sections.test.ts`).

dongyikuan919:
## Summary

- Clarify that Bun is required for development and tests, while Node.js is only needed to run the built server.
- Replace the ineffective `bun test --filter` example with Bun’s supported `-t` test-name option.
- Document the actual pre-push quality hook.

Closes #634

## External assumptions

None

## Verification

- `bun test --help` confirms `-t` / `--test-name-pattern` is the supported test-name filter.
- Verified `.husky/pre-push` runs `bun run check`.
- Reviewed the Markdown diff for formatting and scope.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,8 @@ Contributions welcome! This guide covers development setup, architecture, and ho
 
 ### Prerequisites
 
-- **Bun** (latest) or **Node.js 18+**
+- **Bun** (latest; required for development and tests)
+- **Node.js 18+** (optional; only needed to run the built server)
 - **Copilot Money** installed on macOS (for integration testing)
 
 ### Getting Started
@@ -148,7 +149,7 @@ Same as read tools, plus:
 bun test                                    # Run all tests
 bun test --watch                            # Watch mode
 bun test tests/tools/tools.test.ts          # Specific file
-bun test --filter "getBalanceHistory"        # Pattern match
+bun test -t "getBalanceHistory"               # Test name pattern
 ```
 
 Tests mirror the `src/` structure in `tests/`. The synthetic test DB is generated at runtime by `tests/helpers/test-db.ts` (no checked-in DB fixtures).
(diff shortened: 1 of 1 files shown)


ignaciohermosillacornejo (approved):
LGTM

PR #639, opened 2026-08-12
fix(scheduled-smoke-status): make the status path one knob and bound report on read

### LINKED ISSUE #638: context-budget test reads real $HOME state, and get_connection_status embeds an unbounded scheduled-smoke report
Found while working on #632. Two related things, one a plain test bug, the other a design gap in a budgeted surface. **Neither is a defect in your `scheduled-smoke.ts` writer** — see the disclosure at the end, which matters for how you scope the fix.

## 1. `tests/context-budget.test.ts` reads real `$HOME` state

`bun run check` fails on any machine that has a `~/.claude/copilot-money/scheduled-smoke.json`:

```
(fail) context-budget ratchet (#597) > response-size budgets (cache-mode read tools, synthetic DB)
       > get_connection_status response stays within budget
Expected: <= 1900
Received: 2664
```

Move that one file aside and the same suite is 68/68. Nothing else in the file is environment-dependent — the DB is synthetic, as intended.

The mechanism: `getConnectionStatus` (`src/tools/tools.ts:1345`) calls `readScheduledSmokeStatus()` with no argument, so it resolves `defaultScheduledSmokeStatusPath()` → `join(homedir(), '.claude', 'copilot-money', 'scheduled-smoke.json')`. The reader takes an optional `path` parameter, but the tool never passes one and there is no env override on the read side — `COPILOT_MCP_SMOKE_STATUS_PATH` is consulted only by the writer in `scripts/scheduled-smoke.ts`. So a test has no seam to point it at a fixture, which is presumably why nothing does.

The budget test is exactly the kind that should be hermetic: it asserts a byte count. Right now that count includes whatever a machine happens to have in its home directory, so the ratchet is loose in CI (where the file never exists) and can fire spuriously in dev.

Two ways out, whichever you prefer:

- **Give the read side the same env override the write side has**, then set it to a fixture in the test. That also makes the *populated* case testable, which today it isn't — CI only ever exercises `scheduled_smoke: null`.
- **Inject the status into `CopilotMoneyTools`** rather than reading a well-known path from inside the tool, if you'd rather not add another env var.

## 2. `report` is unbounded inside a budgeted response

`ScheduledSmokeStatusSchema` accepts `report` as any string, and `get_connection_status` embeds its value verbatim. In your writer this is a *path* (`…/smoke-reports/<ts>-smoke-failure.txt`, ~80 chars, `null` on pass), so upstream-only machines stay well inside 1900. But the schema doesn't encode that intent — the doc comment says "Dated report file for failures; null for pass/auth-missing" while the type says `string`. Anything that writes to that path can inflate a budgeted response without tripping any gate, because the budget test can't see the field at all (point 1).

A `z.string().max(...)`, or truncating on read, would make the contract structural rather than conventional. Cheap, and it turns a silent budget blowout into a bounded one.

Worth noting this is the same shape as the finding in #632: a value the code never bounds because the only writer it knows about happens to be well-behaved.

## Disclosure: the oversized file on my machine is mine, not yours

My `scheduled-smoke.json` was written by my own launchd job, not by `scripts/scheduled-smoke.ts`. That script's docstring claims it writes "the exact schema the MCP reads", and it does structurally — `last_run` / `result` / `summary` / `report` all validate — but it puts the **full report text** in `report` where yours puts a path. Mine is 1330 chars, which is the entire 764-char overage.

So: my writer is the non-conforming one, and I'll fix it on my side. I'm still filing this because both points above hold independently of who wrote the file — the test reads real home state either way, and `report` is unbounded in a budgeted response by schema, not by accident of my setup. If you'd rather treat #2 as "don't do that then" and only fix the hermeticity, that's a completely reasonable call and I won't argue it.

## Reproduce

```bash
mkdir -p ~/.claude/copilot-money
python3 - <<'PY'
import json, os
p = os.path.expanduser('~/.claude/copilot-money/scheduled-smoke.json')
json.dump({
    "last_run": "2026-08-10T12:00:07Z",
    "result": "pass",
    "summary": "All 6/6 MCP read tools decoded clean; decode_health ok.",
    "report": "x" * 1200,          # any writer that puts contents here, not a path
}, open(p, 'w'))
PY
bun test tests/context-budget.test.ts   # get_connection_status now exceeds its budget
```

Environment: macOS 15.6 (arm64), bun 1.3.9, `main` at eea599f — reproduces on `main`, so it is not from #632.

papiluxo:
# Summary

Makes the context-budget ratchet hermetic and bounds `scheduled_smoke.report` on read. Closes #638.

The budget test asserted a byte count that included bytes it did not own: `get_connection_status` reads the drift-check status from a `$HOME` path, and the reader had no seam to point at a fixture because `COPILOT_MCP_SMOKE_STATUS_PATH` was honoured by the writer only. `bun run check` therefore failed on any machine with a status file and passed everywhere else, while CI only ever measured `scheduled_smoke: null`.

That blind spot hid the second defect: `report` was `z.string()` with the intent stated only in prose, so any writer to that path could inflate a budgeted response without tripping a gate.

## External assumptions

None

## Bug fix?

Root cause:       `COPILOT_MCP_SMOKE_STATUS_PATH` was writer-only, so the reader had no fixture seam and the budget test measured real `$HOME` state; separately, `report` was unbounded inside a budgeted response.
Bug class:        `unbounded-trusted-payload` — new class proposed in `docs/bugs/README.md`. A budgeted or validated surface embeds a value whose size or shape is guaranteed only by convention, because the only writer it has met is well-behaved. Same shape as #631 one layer up.
Detector added:   `tests/context-budget.test.ts` pins the status path for the whole suite (hermeticity) and adds a case measuring the populated branch against a maximal legitimate status, so 1900 is enforced against a true upper bound. Both mutation-verified — see below.
Siblings checked: Audited all three external-file readers in `src/`. `browser-token.ts` feeds auth, not a tool response. `transaction-meta-store.ts` is reached only via `live-database.ts`, which the cache-mode ratchet does not cover — adjacent rather than an instance, and it is the follow-up already on my list. `scheduled_smoke` is the only external file embedded in a budgeted cache-mode response today.

## What changed

- `defaultScheduledSmokeStatusPath()` honours `COPILOT_MCP_SMOKE_STATUS_PATH`, so reader and writer resolve identically from one variable. `scripts/scheduled-smoke.ts` drops its own `??` and calls the shared helper, per your preference for env parity over a new knob or constructor injection.
- `report` is truncated on read at `SCHEDULED_SMOKE_REPORT_MAX_CHARS` (256) with an ellipsis marker, and the doc comment now says plainly that the intended payload is a path. Truncating rather than `z.string().max()` for the reason you gave: a schema max turns an oversized file into a parse failure, degrading the tool on exactly the machines whose state is most worth reporting.
- Budget suite pins the status path to a temp dir; new case exercises the populated branch, which had never been measured in the project's history.

## Mutation verification

Both detectors were verified by reintroducing the defect, not just by passing:

| Mutation | Result |
|---|---|
| Restore the oversized `~/.claude/copilot-money/scheduled-smoke.json`, run against upstream code | red — `Expected: <= 1900, Received: 4537` |
| Same file, patched code | green — 69 pass |
| Delete the `truncateReport` call, keep everything else | red — populated case, `Expected: <= 256, Received: 382` |
| Restore | green |

The populated case also asserts `scheduled_smoke` is non-null before measuring, so a broken seam cannot make it pass vacuously.

`bun run check`: 2673 pass, 20 skip, 0 fail.

## Note on the post-mortem

`docs/bugs/638-*.md` records this as `detected: incidental` and explicitly does **not** claim a point for CI. The gate fired locally, on a condition CI is structurally blind to, and what it found was its own non-hermeticity. The corpus tally in `README.md` is updated to 45 entries / `incidental: 7`, which I verified matches the frontmatter across all entries rather than incrementing by hand.

The `Detector` section is honest that coverage is class-level for this surface only: there is exactly one budgeted response embedding an external file today, so a general sweep would have one member and would be a regression test wearing a costume. The entry argues for building it when a second appears.

### DIFF
--- a/docs/bugs/638-unbounded-trusted-payload-in-budgeted-response.md
+++ b/docs/bugs/638-unbounded-trusted-payload-in-budgeted-response.md
@@ -1,0 +1,130 @@
+---
+id: 638
+title: A context-budget assertion counted bytes it did not own, and the field that could blow the budget was the one it could not see
+class: unbounded-trusted-payload
+status: fixed
+detected: incidental  # surfaced while working #631; `bun run check` failed on a dev machine and passed in CI
+fixed_in: not yet
+issue: https://github.com/ignaciohermosillacornejo/copilot-money-mcp/issues/638
+date: 2026-08-12
+---
+
+## Symptom
+
+Two behaviours from one cause.
+
+**Developer-visible.** `bun run check` failed on any machine that had a
+`~/.claude/copilot-money/scheduled-smoke.json`, and passed everywhere else:
+
+```
+(fail) context-budget ratchet (#597) > response-size budgets (cache-mode read tools, synthetic DB)
+       > get_connection_status response stays within budget
+Expected: <= 1900
+Received: 2664
+```
+
+Moving that one file aside made the same suite 68/68. Nothing else in the file is
+environment-dependent — the DB is synthetic, as intended.
+
+**Silent, and the more expensive of the two.** `get_connection_status` embedded
+`scheduled_smoke.report` verbatim, and the schema accepted any string. With the repo's own
+writer that field is a path (~80 chars, `null` on pass), so upstream-only machines stayed
+well inside 1900 and nothing ever looked wrong. Any other writer to that path could inflate
+a budgeted response without tripping a gate, because the budget test could not see the
+field at all — see below.
+
+## How it was detected
+
+Incidentally, and by a gate firing for the *wrong* reason. While working #631 the local
+`bun run check` went red on the budget assertion. The proximate cause was a stale
+oversized status file left on the machine by earlier debugging, not by
+`scripts/scheduled-smoke.ts` — so the honest reading is that a machine-specific artifact
+made a latent design gap briefly visible. CI never saw it and never could: the file does
+not exist there, so the populated branch had never been exercised in the project's history.
+
+Worth recording against this corpus's `ci-gate: 0` tally: this is not a point for CI. The
+gate fired locally, on a condition CI is structurally blind to, and the finding was the
+gate's *own* non-hermeticity.
+
+## Root cause
+
+One asymmetry and one missing bound, both in `src/utils/scheduled-smoke-status.ts`.
+
+**1. The env override was writer-only.** `scripts/scheduled-smoke.ts:63` read
+`COPILOT_MCP_SMOKE_STATUS_PATH`; the reader did not. `getConnectionStatus`
+(`src/tools/tools.ts:1345`) calls `readScheduledSmokeStatus()` with no argument, so it
+resolved `defaultScheduledSmokeStatusPath()` → `join(homedir(), …)`. The reader took an
+optional `path`, but the tool never passed one and no env seam existed on the read side, so
+a test had no way to point it at a fixture. That is why nothing did, and why an assertion
+about byte counts ended up counting whatever a developer's home directory contained.
+
+**2. `report` was unbounded inside a budgeted response.** `ScheduledSmokeStatusSchema`
+declared `report: z.string()`, while the doc comment said "dated report file". The intent
+was a path; the type permitted a novel. The contract lived in prose, and prose is not
+enforced.
+
+The second defect was undetectable while the first existed — the only test that measures
+this response could not construct a populated status. Hermeticity was the load-bearing fix.
+
+## Why the tests didn't catch it
+
+- **The context-budget ratchet (`tests/context-budget.test.ts`) had no seam.** It ran the
+  real tool, which read a real home-directory path. In CI that file never exists, so the
+  budget was only ever measured against `scheduled_smoke: null` — the cheapest possible
+  branch. The ratchet held not because the response was bounded but because every writer it
+  had met was polite.
+- **`tests/tools/scheduled-smoke-status.test.ts` covered the reader in isolation** and
+  asserted `report` round-trips, which it did. It never asked how large `report` may be,
+  because nothing in the schema suggested a limit.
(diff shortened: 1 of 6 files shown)


ignaciohermosillacornejo (approved):
Approving — reviewed hands-on, and every claim in the mutation table reproduced:

- `bun run check` passes in full here (2673 pass / 20 skip / 0 fail, matching your numbers).
- **Hermeticity, verified in both directions with a real hostile file at the default path**: upstream code goes red (2709 > 1900 with my file's sizes), this branch stays green with the same file in place — the suite now owns every byte it counts.
- **Truncation mutation**: deleting the `truncateReport` call turns the populated case red at exactly your numbers (382 > 256); restored, green. The non-null vacuity guard does its job.
- The env-parity implementation is the right shape — the writer delegating to the shared helper means there is structurally one knob, not two that agree.
- The post-mortem's refusal to count this as a CI catch keeps the corpus honest, and the "build the sweep when a second surface appears" argument in the Detector section is the correct answer for a class with one member. `unbounded-trusted-payload` is accepted as a class, and the #631 rhyme in the Lesson section — a guarantee that depends on the good behaviour of something the code doesn't control is not a guarantee — is the sentence I expect this corpus to quote later.

One cosmetic nit, no revision needed: the new class's per-bug index entry at the bottom of `docs/bugs/README.md` is a bullet where every other class uses a table, and is missing a blank line after the preceding table row. Fine to sweep into a later nit batch — or into your `transaction-meta-store.ts` follow-up if it's convenient.

Two-for-two on contribution arcs done properly. Merging.

PR #644, opened 2026-08-15
fix(server): scope the db-availability gate to tools that read the local cache (#640)

### LINKED ISSUE #640: get_accounts_live (and all --live-reads/--write tools) blocked by db.isAvailable() check even when native app isn't installed
## Summary

`handleCallTool()` in `dist/cli.js` gates **every** tool call behind `this.db.isAvailable()`, which only checks for the native Copilot Money macOS app's local LevelDB cache (`~/Library/Containers/com.copilot.production/...`). This blocks pure live/GraphQL tools like `get_accounts_live`, `get_connection_status`, `get_cache_info`, etc. — which never touch `this.db` at all (their handlers call `ctx.live.*` directly, confirmed in `TOOL_REGISTRY`) — on any machine where the native app isn't installed, even with a fully valid, logged-in browser session for `--live-reads`/`--write` mode.

## Environment

- `copilot-money-mcp@2.3.0` (npm, latest as of 2026-08-13)
- macOS, invoked via config with `args: ["--write"]`
- Native Copilot Money macOS app **not installed** on this machine (a separate machine syncs it); browser session at `app.copilot.money` is logged in and open

## Steps to reproduce

1. Install `copilot-money-mcp` on a machine that has never had the native Copilot Money app installed
2. Run with `--write` (implies `--live-reads`)
3. Ensure a logged-in browser session exists at `app.copilot.money`
4. Call any tool, e.g. `get_accounts_live`

## Expected

`get_accounts_live` succeeds via the live GraphQL path, since it doesn't need the local cache at all.

## Actual

Every tool call returns:
```
"Database not available. Please ensure Copilot Money is installed and has created local data, or provide a custom database path."
```
This happens **before** the tool's own handler ever runs — the check is unconditional in `handleCallTool()`:

```js
if (!this.db.isAvailable()) {
  return {
    content: [{ type: "text", text: "Database not available. ..." }]
  };
}
```

`isAvailable()` (class `CopilotDatabase`) only checks for `.ldb`/`MANIFEST-*` files at `this.dbPath`, which defaults to the native app's container path. It has no awareness of whether the requested tool actually needs the cache.

## Suggested fix

Scope the `isAvailable()` gate to cache-mode tools only (tools without `requiresLiveReads`), or check it lazily inside each cache-backed tool handler instead of unconditionally in `handleCallTool()`. Live-only tools (`get_accounts_live`, `get_connection_status`, `get_cache_info`, and the rest of the `_live` family) should be callable purely from a valid browser session with zero dependency on local db presence.

## Workaround (confirmed working)

Passing a `--db-path` pointing at any directory containing a dummy file matching `*.ldb` or `MANIFEST-*` satisfies the `isAvailable()` check (which only checks file existence, not content) without providing real cache data. Since live tools never read from `this.db`, this unblocks them entirely:

```bash
mkdir -p /tmp/dummy-copilot-db
touch /tmp/dummy-copilot-db/MANIFEST-000001
copilot-money-mcp --write --db-path /tmp/dummy-copilot-db
```

This works but is a fragile hack that shouldn't be necessary — `get_cache_info`, `refresh_database`, and any genuinely cache-dependent tool call against this dummy path will presumably fail or behave oddly since there's no real data there; only the live-only tool family should be relied on when using this workaround.

ignaciohermosillacornejo:
# Summary

`handleCallTool()` gated **every** tool call on `db.isAvailable()`, which only checks for the native app's local LevelDB cache — so on a machine without the native Copilot Money app, pure-GraphQL live tools (and live-mode writes) returned "Database not available" even with a valid browser session. The gate is now derived from the registry classification and fires only for tools whose dispatch actually reads the cache. Closes #640

Reproduced exactly as reported (live+write server on a nonexistent db path, working GraphQL client): `get_accounts_live` hit the gate before its handler ran; after the fix it completes end-to-end. Credit to @mykemetzger for the precise diagnosis in the issue — the root cause was exactly where they pointed.

## What changed

- `src/server.ts`: the db gate moved after tool resolution and is scoped to `!requiresLiveReads && (readOnly || no live layer)`:
  - **live tools** run entirely on GraphQL (`LiveCopilotDatabase.getCache()` has zero consumers) — never gated;
  - **write tools in live mode** (all production writes, since `--write` implies `--live-reads`) resolve routing/validation live-first and touch the cache only via null-guarded `patchCached*` write-through no-ops — never gated;
  - **cache-mode reads** and **degraded-mode writes** (write without live layer — test-only config) still need the cache — gated as before, same message.
  - Side effect of the reorder: an unknown tool name now reports `Unknown tool` instead of the database message when the cache is absent.
- The db-unavailable response intentionally still omits `isError` (pinned by an existing e2e test) — unchanged.
- `tests/integration/live-reads.test.ts`: instance regression + class-level registry sweeps + pins on unchanged gating.
- `docs/bugs/`: post-mortem entry and a new `overbroad-precondition-gate` class.

## External assumptions

None — this PR touches only dispatch gating; no new assumptions about Copilot's API/data. (The claim that live handlers never read the local cache is internal and verified against the registry: `getCache()` has no callers.)

## Bug fix?

Root cause:       `handleCallTool()` applied the local-cache availability check to every dispatch, blind to the per-tool registry classification that already records which tools need the cache
Bug class:        `overbroad-precondition-gate` (new — a one-resource precondition checked at a shared chokepoint for all requests, locking out configurations where the resource is legitimately absent)
Detector added:   registry-walk sweep in `tests/integration/live-reads.test.ts` — every `LIVE_TOOL_DEFS` and `WRITE_TOOL_DEFS` entry must dispatch past the gate on a cache-less live server; new tools join via registry membership; mutation-verified (reverting the gate to unconditional sends 3 tests red)
Siblings checked: `handleListTools` (no resource gate), CLI startup (no db check — nothing to moot the fix), `preflightLiveAuth` (scoped to live mode, the resource its users need), per-tool handlers (`get_cache_info`, `refresh_database`, `get_connection_status` legitimately read the cache and stay gated) — none overbroad
Ledger updated:   n/a — not an external-assumption bug

https://claude.ai/code/session_01P7FqwcaijCVcJe5vhtVb72

### DIFF
--- a/docs/bugs/640-live-tools-blocked-by-cache-gate.md
+++ b/docs/bugs/640-live-tools-blocked-by-cache-gate.md
@@ -1,0 +1,76 @@
+---
+id: 640
+title: Every tool call — including pure-GraphQL live tools — blocked on machines without the native app's local cache
+class: overbroad-precondition-gate
+status: fixed
+detected: user-report # issue #640 — a --write user on a machine that never had the native app installed
+fixed_in: https://github.com/ignaciohermosillacornejo/copilot-money-mcp/pull/644
+issue: https://github.com/ignaciohermosillacornejo/copilot-money-mcp/issues/640
+date: 2026-08-15
+---
+
+## Symptom
+
+On a machine where the native Copilot Money macOS app was never installed (no LevelDB
+cache on disk), every tool call returned:
+
+```
+Database not available. Please ensure Copilot Money is installed and has created
+local data, or provide a custom database path.
+```
+
+This included `get_accounts_live` and the rest of the `_live` family under
+`--write`/`--live-reads` with a fully valid, logged-in browser session — tools whose
+handlers never touch the local cache at all. The reporter worked around it by pointing
+`--db-path` at a directory containing a dummy `MANIFEST-000001` file.
+
+## How it was detected
+
+User report (issue #640), with the diagnosis already done: the reporter read
+`dist/cli.js`, found the unconditional `db.isAvailable()` check in `handleCallTool()`,
+confirmed from `TOOL_REGISTRY` that live handlers call `ctx.live.*` and never read
+`this.db`, and included a working dummy-file workaround.
+
+## Root cause
+
+`handleCallTool()` (`src/server.ts:210` at the time) applied the `db.isAvailable()`
+gate to every dispatch, before consulting the tool's registry classification. The gate
+predates live mode — when it was written, every tool did read the cache — and was never
+rescoped when `requiresLiveReads` tools arrived. A precondition for one resource
+(the local LevelDB cache) sat at a shared chokepoint and silently asserted "all
+requests need this."
+
+## Why the tests didn't catch it
+
+Live-mode tests built servers either with mock databases whose `isAvailable()` returns
+true, or with paths on dev machines where the real cache exists. The db-unavailable
+tests (`server-protocol.test.ts`, `e2e/server.test.ts`) only ever called cache-mode
+tools, where the gate is correct. The failing configuration — cache absent *and* live
+mode on, i.e. exactly a live-only user's machine — was never constructed, because every
+contributor machine has the native app installed.
+
+## The fix
+
+PR #644. Root-cause fix: the gate is now scoped by the same registry classification
+that drives listing and dispatch — it fires only for tools whose dispatch actually
+reads the cache (`!requiresLiveReads && (readOnly || no live layer)`). Live tools run
+entirely on GraphQL; live-mode writes resolve live-first and touch the cache only via
+null-guarded `patchCached*` write-through, which no-ops when the cache never loaded.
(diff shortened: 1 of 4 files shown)


ignaciohermosillacornejo:
Re the `this.live` vs `this.liveReadsEnabled` nit: kept `!this.live`, addressed the reader-chase with a comment at the gate (80d20a1) instead. Rationale: both gates are justified by the live resolution layer *existing* — the dispatch guard protects handlers' `ctx.live!` dereference, and the cache-gate skip for writes is earned by the live-first write path, which lives on the same object. The constructor makes object and flag equivalent today, but if that ever drifts (flag set, object skipped), the object check degrades to a clean error while the flag check would dispatch handlers into a crash on `ctx.live!`. The gate now states that invariant inline so it no longer requires chasing the constructor.

PR #653, opened 2026-08-16
feat(security): check:concealment — fail on content hidden from review

### LINKED ISSUE #648: security: check:concealment — detect hidden-payload PRs as a class
## Problem

The better-auth attack hid its payload two ways at once:

1. **Horizontal concealment** — appended after ~800 tabs so the rendered diff line ends at `};` and the code is off-screen right
2. **Cross-commit concealment** — added in one commit, removed in a later one, so the combined "Files changed / All commits" diff (the view we actually review from) showed only a trailing-newline change

Neither is visible in normal review. Our audit found **195 "ghost lines"** in this repo's history — content added by a PR commit that is absent from that PR's final tree, i.e. content the combined diff never showed. All benign review churn, but the blind spot is real here.

## Proposal

`scripts/check-concealment.ts` + `bun run check:concealment`, wired into `bun run check` and `test.yml`, in the same class-level-detector style as `check:privacy-endpoints` and `check:tool-counts`.

Flags, on non-Markdown files:

- a run of >=20 whitespace chars mid-line
- invisible / bidi Unicode outside emoji ZWJ sequences
- lines over ~400 columns in `src/`, `scripts/` or any config file
- `eval` / `new Function` / `['constructor']` / long hex-escape runs
- any `preinstall` / `postinstall` / `prepare` / `prepublish` script in `package.json`

Plus a separate PR-scoped mode for the one check no off-the-shelf scanner runs:

- **ghost lines** — content added in a PR commit that is absent from the final tree. Needs the full PR branch, not the merged result.

## Requirements

- Must fail against a planted payload, not just pass on clean input — mutation-test the guards (see #596 lesson). A vacuous detector is worse than none.
- Known-benign exceptions today: the ZWSP in `scripts/smoke/cache.ts` (JSDoc, stops `*/` closing the comment), `createRequire(import.meta.url)` in `src/server.ts`, long generated GraphQL operation strings, emoji ZWJ in captured fixtures.

---
_From the 2026-08-15 fork-PR supply-chain audit, prompted by the [better-auth PR #6003 attack](https://casco.com/blog/the-blueprint-of-a-north-korean-attack-on-open-source). The audit itself came back clean — all 31 external PRs, every commit, every reachable blob. These are gaps that would give a future attempt room to work._

ignaciohermosillacornejo:
# Summary

Adds `check:concealment`, a class-level gate against changes engineered to be invisible in code review. Prompted by the better-auth PR #6003 attack and the audit of this repo's 31 external fork PRs that followed it — that audit came back clean, so this is prophylactic, not remedial.

Closes #648

## External assumptions

None

## What it catches

better-auth #6003 appended an obfuscated loader to a postcss config after ~800 tab characters, so the rendered diff line ended at `};` and the payload sat off-screen right. A later commit in the same PR deleted it, so the combined "Files changed / All commits" view — the one maintainers review from — showed only a trailing-newline change.

Two properties of that shape decide where this gate sits:

- **The payload was never merged.** The PR was closed unmerged and the attack still worked, because a build config executes during install/build. It only had to survive on the branch long enough to run once, in CI or on a maintainer's machine. So this runs on every PR, forks included, not at merge.
- **Nothing about it is unusual to a compiler.** It parses, lints and type checks. Only its *shape on screen* is anomalous — so this checks shape, not semantics.

Rules, each verified to produce zero hits across every blob reachable from `main` and from all 31 external fork PR heads:

| Rule | Threshold |
|---|---|
| whitespace run on an over-wide line | ≥20 space/tab chars **and** line >120 cols |
| over-long line | >400 cols, generated files excepted |
| invisible / bidi formatting characters | U+200D allowed between two symbols so emoji survive |
| dynamic execution | `eval`, `new Function`, `['constructor']`, hex-escape runs |
| install-time scripts | `preinstall`/`install`/`postinstall` refused; `prepare`/`prepublish*` pinned to exact values |

The whitespace rule needs **both** conditions. This repo aligns JSDoc continuations with gaps well over 20 chars — 19 such lines, longest 94 cols — and those are reviewable precisely because they end where you can see them. Gap alone would have made the gate noisy enough to get switched off, which is the usual way a check like this dies.

## Verification

- `bun run check` exits 0; 2690 tests pass
- 23 new tests, driven end-to-end against synthetic trees via `CHECK_CONCEALMENT_ROOT`, matching the sibling gates
- **All 7 guards mutation-tested** — disabling any one of them turns the suite red. Per #596, a suite that only feeds clean input executes every branch and still passes with the whole checker commented out, so coverage alone would not have shown this.
- One test asserts the gate stays green on the real repository tree, so it fails loudly rather than getting disabled the first time it blocks a merge.

## Incidental fixes needed to land green

None change behaviour:

- removed a zero-width space from a JSDoc comment in `scripts/smoke/cache.ts` that existed to stop `*/` closing the block early
- rewrapped two over-long tool description strings into the concatenation style already used in `src/tools/registry/transactions.ts`. **Tool schemas verified byte-identical** before and after, rather than assumed — the reformat only changes how the literal is written
- added `!scripts/check-concealment.ts` to `.gitignore`; `scripts/*` is allowlist-ignored, so without it the script would never have reached CI

## Not in this PR

The cross-commit half of #6003 — content added by one commit and removed by another, so it never appears in the combined diff — cannot be seen from a working tree. It needs the PR's commit range. That check found 195 real instances ("ghost lines") in this repo's history, all benign review churn, and lands in a follow-up so this one stays reviewable. #648 stays open until it does.


ignaciohermosillacornejo:
All four review items addressed in c98139c, plus the missing tests.

**1. `withTree` dropping async assertions** — real trap, fixed. Widened to `void | Promise<void>` and awaited. No current callback is async, so nothing was broken; the harness just would not have told us if one became async.

**2. `prepublish`** — you were right that it was accidental correctness. It reached the right answer through an undefined lookup. Moved it to `FORBIDDEN_LIFECYCLE` where it belongs (deprecated npm hook that also ran on plain `npm install`, so it is consumer-facing), and `PINNED_NAMES` is now `Object.keys(PINNED_LIFECYCLE)` — a hook can no longer be listed as pinned without a value to pin it to.

**3. Triple read of package.json** — threaded the already-read contents through.

**4. `0x2000` floor** — documented. It is not the emoji block; it is the point above which every character that legitimately neighbours a ZWJ lives, the lowest real ones being the gender signs U+2640/U+2642.

**Missing tests** — added all four, plus one you did not ask for: *the whitespace rule still applies inside generated files*. The long-line exemption is the obvious place to hide a payload once it exists, so it needed a test proving the exemption is narrow rather than a hole.

Mutation coverage went from 7 guards to **12** — the exemptions are now mutation-tested too (generated-file exemption, binary skip, vendored-dir skip, lockfile skip, markdown scope), on the same reasoning: an exemption that cannot be removed detectably is the easiest place to hide something. All 12 killed.

`bun run check` exits 0, 2708 tests pass.

PR #654, opened 2026-08-16
ci(security): pin every action by SHA, deny gh api writes in the audit agent

### LINKED ISSUE #651: ci: pin third-party actions by SHA
## Problem

`audit-reviews.yml` correctly pins its actions by full commit SHA. Others do not:

- `claude-review.yml` — `actions/checkout@v7`, `1password/load-secrets-action@v4`, `anthropics/claude-code-action@v1`, `actions/github-script@v9`
- `test.yml` — `actions/checkout@v7`, `oven-sh/setup-bun@v2`, `codecov/codecov-action@v7`

Repo setting `sha_pinning_required` is **off**.

A git tag can be repointed by whoever controls the action repository, and `claude-review.yml` has `OP_SERVICE_ACCOUNT_TOKEN` in scope.

## Fix

- Pin every third-party action by full commit SHA with a trailing `# vX` comment, matching the style already used in `audit-reviews.yml`
- Enable the repository's SHA-pinning requirement so new workflows inherit the rule
- Dependabot already updates SHA-pinned actions, so this does not add manual upkeep

---
_From the 2026-08-15 fork-PR supply-chain audit, prompted by the [better-auth PR #6003 attack](https://casco.com/blog/the-blueprint-of-a-north-korean-attack-on-open-source). The audit itself came back clean — all 31 external PRs, every commit, every reachable blob. These are gaps that would give a future attempt room to work._

### LINKED ISSUE #652: ci: narrow the audit-reviews agent allowlist to read-only gh calls
## Problem

`audit-reviews.yml` uses `pull_request_target`, which is the dangerous trigger — but it is used **correctly**: it checks out the base branch rather than the PR head, is gated on `merged == true`, and never installs or builds. No unmerged fork code runs with secrets in scope.

The residual risk is prompt injection. The agent reads PR diffs and review comments — contributor-controlled text — while holding `issues: write` and `pull-requests: write`, and its allowlist contains `Bash(gh api*)`. The deny list blocks `gh pr review`, `gh pr merge`, `gh pr close`, `gh issue` and `gh pr comment`, but **not** `gh api -X POST`, which reaches all the same endpoints.

Blast radius is capped at issue/comment spam today because `contents` is read-only. Still free to close.

## Fix

Either narrow the allowlist to the specific read endpoints the prompt needs, or add `Bash(gh api -X*)` and `Bash(gh api --method*)` to `disallowedTools`.

While here: add a comment to the workflow recording *why* the checkout has no `ref:` — it is one line away from becoming the critical variant of this misconfiguration.

---
_From the 2026-08-15 fork-PR supply-chain audit, prompted by the [better-auth PR #6003 attack](https://casco.com/blog/the-blueprint-of-a-north-korean-attack-on-open-source). The audit itself came back clean — all 31 external PRs, every commit, every reachable blob. These are gaps that would give a future attempt room to work._

ignaciohermosillacornejo:
# Summary

Two CI hardening items from the fork-PR supply-chain audit: pin every third-party action to an immutable SHA, and stop the audit agent from reaching write endpoints through `gh api`.

Closes #651
Closes #652

## External assumptions

None

## Action pinning

A git tag is mutable by whoever controls the action repository, so `@v7` is a standing grant of code execution to that repo's maintainers. `audit-reviews.yml` was already pinned; nothing else was. The scope turned out to be wider than #651 described — the **release pipeline** was tag-referenced too:

| Workflow | What it holds |
|---|---|
| `claude-review.yml` | `OP_SERVICE_ACCOUNT_TOKEN` |
| `npm-publish.yml` | `id-token: write` — npm OIDC trusted publishing |
| `build-mcpb.yml` | builds the released bundle |
| `auto-release.yml`, `test.yml` | — |

All now pin the full commit SHA with a trailing version comment, matching the style `audit-reviews.yml` already used. `actions/upload-artifact` resolved to the same SHA that workflow was already on, so the pins agree rather than diverge.

Dependabot updates SHA-pinned actions, so this adds no manual upkeep.

## gh api write denial

`audit-reviews.yml` runs an agent over contributor-controlled text — the PR diff and review comments — while holding `issues: write` and `pull-requests: write`. That is a prompt-injection surface.

Its deny list blocked `gh pr review`, `gh pr merge`, `gh pr close`, `gh pr comment` and `gh issue`. But `Bash(gh api*)` sat in the *allow* list, and `gh api -X POST` reaches every one of those endpoints. The deny list now also covers `gh api -X`, `--method`, `-f`, `--field` and `--input`.

Blast radius was capped at issue and comment spam either way — `contents` is read-only on that job — so this is closing a gap, not a live hole.

## Also

Added a comment on that workflow's checkout explaining why it has no `ref:`. Under `pull_request_target`, the base-branch checkout is the property that makes the trigger safe; adding `ref: github.event.pull_request.head.sha` would turn it into the critical variant of that misconfiguration, and nothing on the line said so.

## Verification

- all nine workflow files parse
- no `uses:` reference outside `./.github/...` remains unpinned:
  ```
  grep -rn "uses:" .github/workflows/ | grep -v "\./\.github" | grep -vE "@[0-9a-f]{40}"
  ```
  returns nothing

## Follow-up

The repository-level "require actions pinned to SHA" setting should be enabled **after** this merges — turning it on first would break the workflows this PR is fixing.

### DIFF
--- a/.github/workflows/audit-reviews.yml
+++ b/.github/workflows/audit-reviews.yml
@@ -75,6 +75,24 @@ jobs:
           OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
           CLAUDE_CODE_OAUTH_TOKEN: op://Dev/CLAUDE_CODE_OAUTH_TOKEN/credential
 
+      # Pre-fetch every API payload the audit needs, so the agent never needs
+      # `gh api` at all. This is the whole reason the allow list below can be
+      # read-only: --disallowedTools matches command PREFIXES, so denying
+      # `gh api -X*` does nothing against `gh api repos/... -X POST`, which is
+      # ordinary gh usage and still matches the `gh api*` allow pattern. There
+      # is no prefix pattern that reliably separates reads from writes, so the
+      # capability is removed instead of filtered.
+      - name: Fetch PR review data
+        if: steps.check.outputs.skip != 'true'
+        run: |
+          mkdir -p "$RUNNER_TEMP/audit"
+          gh api "repos/${{ github.repository }}/issues/$PR_NUMBER/comments" \
+            --paginate --jq '.[]' | jq -s '.' > "$RUNNER_TEMP/audit/issue-comments.json"
+          gh api "repos/${{ github.repository }}/pulls/$PR_NUMBER/comments" \
+            --paginate --jq '.[]' | jq -s '.' > "$RUNNER_TEMP/audit/pr-comments.json"
+          gh api "repos/${{ github.repository }}/pulls/$PR_NUMBER/reviews" \
+            --paginate --jq '.[]' | jq -s '.' > "$RUNNER_TEMP/audit/pr-reviews.json"
+
       - name: Audit PR reviews with Claude
         if: steps.check.outputs.skip != 'true'
         uses: anthropics/claude-code-action@e30d45253b1abcc29a121ab35b8bf156061582bd # v1
@@ -88,10 +106,10 @@ jobs:
 
             ## Steps
 
-            1. Fetch every Claude review body for this PR:
-               - `gh api repos/${{ github.repository }}/issues/${{ env.PR_NUMBER }}/comments --paginate`
-               - `gh api repos/${{ github.repository }}/pulls/${{ env.PR_NUMBER }}/comments --paginate`
-               - `gh api repos/${{ github.repository }}/pulls/${{ env.PR_NUMBER }}/reviews --paginate`
+            1. Read every Claude review body for this PR. They are already fetched:
+               - `${{ runner.temp }}/audit/issue-comments.json`
+               - `${{ runner.temp }}/audit/pr-comments.json`
+               - `${{ runner.temp }}/audit/pr-reviews.json`
                Keep only entries by `github-actions[bot]` whose body contains both "PR Review"
                or "Claude finished" AND "actions/runs/". Ignore everything else.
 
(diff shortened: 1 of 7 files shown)


ignaciohermosillacornejo:
Good catch, and the finding is correct — my deny list did not work.

`--disallowedTools` matches command **prefixes**. `Bash(gh api -X*)` only fires when the command literally starts with `gh api -X`. The ordinary form:

```
gh api repos/owner/repo/issues -X POST -f body=...
```

starts with `gh api repos/...`, matches the `Bash(gh api*)` **allow** pattern, and matches no deny pattern. Every write flag was reachable just by putting it after the endpoint, which is how people actually write `gh` commands. Same for `--method`, `-f`, `--field`, `--input` — and you are right that `-F` was never listed at all.

I took the structural fix rather than adding `-F` to the list, because the list was the wrong shape: there is no prefix pattern that reliably separates reads from writes on the same subcommand.

`ca739a1`: a plain `run:` step pre-fetches the three payloads, `gh api` moves to the **deny** list, and the prompt reads the files instead. What remains in the allow list is read-only by construction — `gh pr view` and `gh pr diff` have no write form.

Two things I kept from your suggestion but changed slightly: the files go in `$RUNNER_TEMP` rather than `/tmp` so they are cleaned up with the job, and each fetch is `--paginate --jq '.[]' | jq -s '.'` rather than bare `--paginate`, which emits concatenated arrays rather than one array for multi-page results.

Side benefit: pagination now happens once in shell instead of as three agent tool calls whose completeness the agent had to be trusted on.

On the `claude-code-action` SHA divergence — agreed it is expected, and Dependabot will converge them. Leaving as is.

Branch also rebased onto main to pick up #653.

ignaciohermosillacornejo:
Bundled the `auto-merge.yml` fix here (`ffea399`), since this PR already touches the workflows.

`gh pr merge --auto --rebase --delete-branch` is not "merge, then tidy up". When GitHub refuses the merge, `gh` falls back to **closing the PR and deleting the branch** — so a merge blocked by branch protection silently becomes a closed, branch-deleted PR, recoverable only from the reflog.

It was also redundant: the repo already has `delete_branch_on_merge` enabled, which deletes the branch on an actual merge and does nothing on a refusal. The flag bought nothing and cost the PR.

Newly urgent because `enforce_admins` is now on for `main` — merges are refused in more situations, including for the owner, which is exactly when the fallback fires.

Issue #624, opened 2026-08-05
get_accounts (cache): include_hidden filters against an empty collection; the real user_hidden flag is decoded then ignored


ignaciohermosillacornejo:
`get_accounts` (cache mode) hides nothing. Its `include_hidden=false` path filters against a Firestore collection that no longer has any documents, while the flag Copilot actually writes is decoded and then ignored.

Found while auditing siblings of #622 — same class (code reads a data shape reality no longer has), different collection.

## Symptom

An account hidden in the Copilot app is returned by a default `get_accounts()` call. Verified on a real cache: the cache contains an account with `user_hidden: true`, and a default call — `include_hidden` defaulting to `false` — returns it along with everything else. Nothing is filtered.

## Cause

`src/tools/tools.ts:1218-1227`:

```ts
if (!include_hidden) {
  accounts = accounts.filter((acc) => acc.user_deleted !== true);

  // Also filter by hidden flag from user account customizations
  const userAccounts = await this.db.getUserAccounts();
  const hiddenIds = new Set(userAccounts.filter((ua) => ua.hidden).map((ua) => ua.account_id));
  accounts = accounts.filter((acc) => !hiddenIds.has(acc.account_id));
}
```

Two problems:

1. **`getUserAccounts()` reads `users/{uid}/accounts`, which is empty.** A full scan of a real cache finds **zero** documents under that path. Copilot appears to have migrated account customizations onto the account documents themselves — `items/*/accounts` now carries `nickname`, `user_hidden`, and `dashboard_active` directly. So `hiddenIds` is always the empty set and the second filter is a no-op.
2. **The flag that *is* present is never consulted.** `Account.user_hidden` exists in the Zod model (`src/models/account.ts:74`) and is populated by the decoder, but `user_hidden` appears **nowhere** in `src/tools/tools.ts` or `src/core/database.ts` — it is decoded and dropped.

The surviving `user_deleted` filter still works (that field is on the account document), which is why this isn't total — merged/removed accounts are still excluded. Only user-*hidden* accounts leak.

## Scope

- **Cache mode only.** `get_accounts_live` is GraphQL-backed and unaffected; `--live-reads` / `--write` sessions get correct results.
- The code comment says "same pattern as getNetWorth", but there is no longer a `getNetWorth` in `tools.ts` (net worth moved to `get_networth_live`), so the blast radius is `get_accounts` alone. Worth confirming during the fix rather than trusting the stale comment.

## Fix

[...]

PR #660, opened 2026-08-19
fix(accounts): use Copilot nicknames in account responses

davidperjac:
## What changed

`get_accounts` now returns an account's Copilot `nickname` as its `name` when a nickname is present.

## Why

The local cache keeps the provider label in `name` and the user-visible Copilot label in `nickname`, so renamed accounts were returned with generic provider labels.

## Validation

- Added a regression test for nickname precedence.
- `bun test` (2709 passed, 20 real-cache integration tests skipped by the suite)
- `bun run build`
- `bun run typecheck`
- `bun run lint` (existing warnings only; no errors)
- `bun run format:check`

## External assumptions

None. `nickname` is already a decoded field on the cached account document
(`src/models/account.ts:40`, `src/core/decoder.ts:956`) and this change only
selects between two fields we already read. No new assumption about Copilot's
API or data shape.

## Bug fix?

Root cause:       `getAccounts` returned the institution-supplied `name` and never consulted the user-set `nickname`, so renamed accounts surfaced under provider labels.
Bug class:        `path-divergence` — the nickname→display-name rule already existed in `CopilotDatabase.getAccountNameMap()` (`src/core/database.ts:1357`) and fed `get_balance_history`, but `get_accounts` never applied it, so which name a caller saw depended on which tool they called.
Detector added:   None at class level — this PR adds an instance regression test only (`tests/tools/tools.test.ts`). The class-level fix is to resolve the display name in ONE place and have both call sites use it; that unification is tracked as part of the v3 accounts work (#597 Tier 2) rather than asked of the contributor.
Siblings checked: `getAccountNameMap()` (`src/core/database.ts:1357`) already applies `nickname ?? name` — the two paths now agree in behaviour, though they remain two implementations that differ on an empty-string nickname (`??` vs truthiness). `get_accounts_live` cannot apply the rule at all: Copilot's GraphQL `AccountNode` (`src/core/graphql/queries/accounts.ts:17-45`) exposes no `nickname` field, so cache mode and live mode now disagree on `name` — a parity gap to close alongside the shared accounts preset.
Ledger updated:   n/a — local-cache read path, not an external-assumption bug.

### DIFF
--- a/src/tools/tools.ts
+++ b/src/tools/tools.ts
@@ -1245,6 +1245,10 @@ export class CopilotMoneyTools {
       accounts = accounts.filter((acc) => acc.user_deleted !== true && acc.user_hidden !== true);
     }
 
+    accounts = accounts.map((account) =>
+      account.nickname ? { ...account, name: account.nickname } : account
+    );
+
     // Calculate totals by asset/liability classification
     let totalAssets = 0;
     let totalLiabilities = 0;
--- a/tests/tools/tools.test.ts
+++ b/tests/tools/tools.test.ts
@@ -822,6 +822,22 @@ describe('CopilotMoneyTools', () => {
       expect(result.accounts).toHaveLength(2);
     });
 
+    test('uses the Copilot nickname as the account display name', async () => {
+      db._injectDataForTesting({
+        accounts: [
+          {
+            ...mockAccounts[0],
+            name: 'Provider Checking',
+            nickname: 'Household Checking',
+          },
+        ],
+      });
+
+      const result = await tools.getAccounts();
+
+      expect(result.accounts[0].name).toBe('Household Checking');
+    });
+
     test('filters by account type', async () => {
       const result = await tools.getAccounts({ account_type: 'checking' });
       expect(result.count).toBe(1);


ignaciohermosillacornejo:
Thanks for this — it's a real bug and a clean fix. 🙏

I verified it with a live round-trip rather than just reading the diff, and the result is a bit better than the PR claims — cache mode and live mode now agree *exactly*:

| | before | after |
|---|---|---|
| cache-mode `name` matching live GraphQL `name` | 7 / 17 accounts | **17 / 17** |

On a real cache, 10 of 18 accounts carry a nickname that differs from `name`, and several collapse onto an identical generic provider label — so `get_accounts` was handing clients rows they couldn't tell apart. Your fix is also the right *direction*, not just a patch: Copilot's own GraphQL `Accounts.name` already returns the user-set label, so this brings the cache path into parity with live mode and with the app UI.

**One thing needed before it can merge.** The Required PR Sections check will fail as-is — it hasn't reported yet because fork workflows need my approval, which I'll do. Two additions to the PR body, no code changes:

1. An `## External assumptions` section. Yours isn't "None" — it assumes Copilot stores the user-visible label in `nickname` and the institution label in `name`. Evidence class is **live round-trip**; feel free to cite the 17/17 parity above. `Ledger updated:` is n/a — the ledger tracks GraphQL operation assumptions, not cache field semantics.
2. Because the title is `fix(accounts):`, the gate also requires all five Bug Response Ritual labels verbatim: `Root cause:`, `Bug class:`, `Detector added:`, `Siblings checked:`, `Ledger updated:` (see CONTRIBUTING.md). For `Bug class:` — this is a third instance of `fixture-reality-drift`, documented in `docs/bugs/624-hidden-accounts-not-filtered.md`, which names exactly the fields Copilot migrated onto account documents: `nickname`, `user_hidden`, `dashboard_active`. #624 fixed `user_hidden`; you're fixing `nickname`.

**Follow-ups this PR surfaced.** All pre-existing, none of them blockers:

1. **The decoder dedups accounts by provider name, ignoring nicknames.** `decodeAccounts` (`src/core/decoder.ts:325`, same key again at `:2977`) keys on `` `${name ?? official_name}|${mask ?? ''}` ``. On a real cache it silently drops accounts — the decoder demonstrably processes documents that never appear in its output, and cache mode returns fewer accounts than live mode. The dropped ones carry no balance today so nothing looks wrong, but the key drops any same-name/same-mask pair — the #72/#82 "account silently missing" class. It also partly undercuts this PR's premise: renaming two identically-named accounts to tell them apart doesn't help while dedup runs on the provider name first.

2. **Two sibling surfaces still report the institution label** — `resolveAccountName()` (`src/tools/tools.ts:637`, feeding `get_recurring_transactions`) and `getHoldings()` (`:2553`). Neither reproduces on the cache I checked, so they're latent rather than broken — but same class.

3. **Consolidating "account display name."** There are already three definitions — `getAccountDisplayName()` (`models/account.ts:115`), `getAccountNameMap()` (`database.ts:1357`, already `nickname ?? name`), and the decoder's dedup key — and this adds a fourth inline. Worth flagging the trap: folding nickname precedence into `getAccountDisplayName()` would *also* change the dedup key in (1), and therefore how many accounts survive decode. Deliberate, test-covered change — not a one-liner.

4. **Smaller ones:** the `get_accounts` description (`src/tools/registry/accounts-system.ts`) and `docs/tools-by-mode.md` don't mention that `name` is now the user's nickname; there's no cache↔live *naming* parity test (`tests/tools/live/accounts.test.ts` pins type/liability parity for the same reason); and the new test only asserts the positive branch — deleting the guard leaves it green, so it'd want the no-nickname fallback case too, per the #596 mutation discipline.

Would you like to take any of these on? Genuinely no pressure — if you'd rather stop at this fix, that's completely fine and I'll pick them up after merging.

Thanks again for digging into this one.

ignaciohermosillacornejo (approved):
Thanks for this — it's a real user-visible fix and the regression test is well placed.

Worth recording why it matters beyond the symptom: the nickname→display-name rule already existed in `CopilotDatabase.getAccountNameMap()` and fed `get_balance_history`, but `get_accounts` never applied it. So which name a caller saw depended on which tool they called — a `path-divergence` bug, not just a missing feature.

I filled in the PR template sections rather than sending you another round trip for them. Two follow-ups I'm taking on in the v3 accounts work, so nothing needed from you:

1. The rule now has two implementations that disagree on an empty-string nickname (`nickname ?? name` vs truthiness). That collapses into one shared helper.
2. `get_accounts_live` can't apply the rule at all — Copilot's GraphQL `AccountNode` exposes no `nickname` — so cache and live modes now disagree on `name`. Closing that alongside the shared accounts field preset.

Merging.

Issue #645, opened 2026-08-15
db-unavailable response omits isError: true — clients see a successful call carrying an error message


ignaciohermosillacornejo:
## Summary

The `Database not available...` response in `handleCallTool()` (`src/server.ts`, cache-gate branch) does not set `isError: true`, unlike the sibling guards (write-mode block, live-reads block, unknown tool), so MCP clients observe a *successful* tool call whose payload is an error message.

## Context

This is **intentional today** and pinned by a test: `tests/e2e/server.test.ts` ('database unavailable returns informative message') asserts `isError` is undefined, with a comment saying the message is meant to reach the LLM as guidance rather than a hard error. PR #644 deliberately preserved this while rescoping the gate (see its review thread, where the reviewer flagged it as worth a separate decision).

## Decision needed

Either:
- flip the response to `isError: true` and update the pinning e2e assertion (consistent with every other dispatch guard), or
- keep the guidance-not-error behavior and move the rationale comment from the test into `src/server.ts` so the asymmetry is explained at the site that produces it.

Refs: #640, #644.

PR #670, opened 2026-08-22
fix(persistence): cap valve cleans up its tmp file when the rename throws

### LINKED ISSUE #641: transaction-meta-store: cap valve can strand its .pid.tmp file if the rename throws (claimed follow-up from #632/#639)
Tracking issue for a follow-up @papiluxo claimed during the #632 and #639 review arcs, so it survives even if priorities shift. Santiago: this is yours as discussed — no timeline pressure; unassigned means unclaimed, not unwelcome.

## The finding (his, from the #632 thread)

`transaction-meta-store.ts`'s cap valve writes `${file}.${pid}.tmp` and then renames. If the write succeeds and the rename throws, the `.tmp` stays behind. Bounded and tiny — but it's the same *stranded-artifact* shape as the temp-copy leak fixed in #632 (an artifact created before the operation that names it durably completes), reached through the write-then-rename idiom rather than a timer.

From #639's sibling audit: this store is reached only via `live-database.ts`, which the cache-mode context-budget ratchet does not cover — adjacent to `unbounded-trusted-payload` rather than an instance of it, which is why it was deferred rather than folded into that PR.

Scope for the fix, per the "close the shape everywhere" offer in the #632 thread: audit every write-then-rename in `src/` for the same gap, not just this one call site.

## Rider: docs/bugs/README.md formatting nit (from the #639 approval)

The `unbounded-trusted-payload` per-bug index entry at the bottom of `docs/bugs/README.md` is a bullet list where every other class uses a `| | Bug | Found by | Date |` table, and the bold class header is glued to the preceding table row without a blank line. One-minute fix; folding it in here avoids a standalone one-line PR.

papiluxo:
# Summary

The meta-index cap valve stranded its `${file}.${pid}.tmp` when the write landed and the rename threw. The write-then-rename idiom now lives in `writeFileAtomic` (`src/utils/atomic-write.ts`), which removes the tmp on any failure before propagating; a sweep test confines the idiom to that helper. Includes the docs/bugs/README.md formatting rider from the #639 approval. Closes #641

## External assumptions

None

## Bug fix?

Root cause:       The valve's catch handled the error (warn once, degrade) but not the artifact — a tmp written before the rename that makes it durable has no owner on the failure path.
Bug class:        `stranded-artifact` (new class row in docs/bugs/README.md) — sibling of `deferred-cleanup-never-runs`: same litter, reached through a failure branch rather than a timer that never fires.
Detector added:   `tests/core/write-then-rename-sweep.test.ts` — `renameSync` is structurally confined to `writeFileAtomic`, whose failure-path cleanup is unit-tested through an ops seam (rename-throws, write-throws, and cleanup-throws paths). A future direct `renameSync` in `src/` fails CI.
Siblings checked: Swept every write-then-rename in `src/` per the #632 offer — the cap valve was the only site. The fs.promises `rename` spelling is also covered by the sweep regex; no hits.

## Notes

- **Why an ops seam instead of a filesystem fixture or a mock:** a rename-only failure has no portable setup — anything that breaks the rename in the same directory (dir permissions) breaks the tmp write first. A `mock.module('fs')` approach was tried and abandoned: the mock is process-global in bun and leaked into the sibling store test file, and a spread-copy factory (`{...realFs, ...}`) hangs the whole run on bun 1.3.9.
- Store behavior on valve failure is unchanged: degrade with one warning, never fail the caller — now without litter. The existing store-level failure test (dir chmod'd read-only) still passes unchanged.
- Rider: the `unbounded-trusted-payload` per-bug index entry now uses the same table form as every other class, with the blank line its header was missing.
- A `docs/bugs/MINOR.md` row will follow in a second commit on this branch (it cites this PR's number).

https://claude.ai/code/session_017PbxVtLNP3SBPegdVqm7mS


tests/core/write-then-rename-sweep.test.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+describe('write-then-rename sweep (#641)', () => {
+  test('renameSync appears in src/ only inside writeFileAtomic', () => {
+    const offenders: string[] = [];
+    for (const file of walk(SRC_ROOT)) {
+      const rel = file.slice(SRC_ROOT.length + 1);
+      if (ALLOWED.has(rel)) continue;
+      const text = readFileSync(file, 'utf8');
+      if (/\brenameSync\b|\brename\s*\(/.test(text)) offenders.push(rel);
```
ignaciohermosillacornejo: The `\brename\s*\(` alternative buys less than it costs. I mutation-tested the sweep against five variants:

| mutation in `src/core/database.ts` | sweep |
|---|---|
| `renameSync("a","b")` | ✅ fails |
| `import { renameSync as mv }` … `mv("a","b")` | ✅ fails (the import line trips `\brenameSync\b`) |
| `await fsp.rename("a","b")` | ✅ fails |
| `// we rename (later) the field` | ⚠️ fails — false positive |
| `export function rename(a: string) {…}` | ⚠️ fails — false positive |

The aliased-import case is already caught by `\brenameSync\b` on the import line, so the broad alternative is only earning the `fs.promises` spelling — and it pays for that with two false-positive shapes. A prose comment containing "rename (" or an unrelated domain function named `rename(` (plausible here — rename a tag/category) fails CI, and the pressure will be to widen `ALLOWED`, which weakens the real check. #619 was itself a detector-vs-comments bug, so this is on-theme for the repo.

Suggest narrowing to keep all three real spellings and drop both false positives:

```ts
if (/\brenameSync\b|\bpromises\s*\.\s*rename\b|\bfsp?\s*\.\s*rename\s*\(/.test(text)) offenders.push(rel);
```

Separately: `expect(offenders).toEqual([])` gives a future contributor a bare array diff with no hint of the remedy. Worth a message pointing at the helper, e.g.

```ts
expect(offenders, `write-then-rename must go through writeFileAtomic (src/utils/atomic-write.ts) — see #641`).toEqual([]);
```

---

docs/bugs/README.md
```diff
@@ -95,6 +95,7 @@ a class yet.
...
 | `unsettled-promise` | An async operation has a completion path where neither resolve nor reject fires, hanging callers forever. | none |
 | `deferred-cleanup-never-runs` | Releasing a resource is deferred to a timer or callback that only fires if the process outlives it — in a process that routinely exits first. The resource is acquired eagerly and released never, while every same-process test still observes correct bookkeeping. | none — instance-only; the class is a deployment property (does this process outlive its own timers?) that no static check here can evaluate |
+| `stranded-artifact` | An artifact created as a step toward a durable operation — a temp copy, a `.tmp` awaiting rename — outlives a failure that lands between its creation and the operation completing, because the failure path handles the error but not the artifact. Sibling of `deferred-cleanup-never-runs`: same litter, reached through a failure branch rather than a timer that never fires. | write-then-rename sweep (`tests/core/write-then-rename-sweep.test.ts`): the idiom is structurally confined to `writeFileAtomic`, whose failure-path cleanup is unit-tested through an ops seam |
```
ignaciohermosillacornejo: The detector column overclaims relative to the class description on the same row.

The description covers two shapes — *"a temp copy, a `.tmp` awaiting rename"* — but the sweep only structurally confines write-then-rename. The temp-copy shape has no structural detector: `src/core/leveldb-reader.ts:160` (`mkdtempSync`, i.e. #631) is correct today only because that call site hand-rolls its own `catch → cleanupTempDatabase(tempDir)`. Nothing fails if a future temp-dir site forgets that catch.

Every other row in this table is scrupulous about exactly this — `unbounded-trusted-payload` says `partial — … for this surface only`, and `deferred-cleanup-never-runs` says `none — instance-only`. Suggest matching that honesty:

> `partial` — write-then-rename sweep (`tests/core/write-then-rename-sweep.test.ts`): the idiom is structurally confined to `writeFileAtomic`, whose failure-path cleanup is unit-tested through an ops seam. The temp-directory shape (#631) remains instance-only — no static check confines `mkdtemp`-then-populate.

Everything else in the docs changes checks out: I confirmed minor-ledger bugs (#588, #619, #645) correctly stay out of the per-bug index, so #641 living only in `MINOR.md` is right, and the `unbounded-trusted-payload` rider matches the table form exactly.

---

src/utils/atomic-write.ts
```diff
@@ -0,0 +1,43 @@
...
+  rmSync: (file: string, opts: { force: boolean }) => void;
+}
+
+const REAL_OPS: AtomicWriteOps = { writeFileSync, renameSync, rmSync };
+
+/**
+ * Write `data` to `${file}.${pid}.tmp`, then rename over `file`. If either
+ * step throws, the tmp is best-effort removed before the error propagates —
+ * the caller owns the failure policy, this helper owns not littering.
+ */
```
ignaciohermosillacornejo, line 34: Docstring nit, forward-looking only — no change in behavior for the current caller.

Two things the name promises that the implementation doesn't deliver:

1. **Mode is not preserved.** The tmp is created at default umask and then renamed over the target, so the target's permissions are replaced by the tmp's. I checked — there are no mode-sensitive writes anywhere in `src/` today, so nothing is affected. But this is now advertised as *the* home of the idiom, and the repo does handle auth tokens (`src/core/auth/browser-token.ts`), so a future caller could quietly widen a `0o600` file.
2. **Not crash-durable.** No `fsync` on the tmp before the rename, and none on the directory after. It's atomic against a concurrent *reader* (which is all the cap valve needs), not against power loss — "atomic" reads like the stronger guarantee.

Neither is a regression: the inlined version at the call site had exactly these properties. Just worth one line in the docstring so the next caller picks the helper with eyes open, e.g. *"Does not preserve the target's mode, and does not fsync — atomic against concurrent readers, not durable across a crash."*

---

ignaciohermosillacornejo (commented):
Reviewed by checking out the branch and verifying the claims rather than reading the diff alone. **Nothing blocking** — the fix is correct and the ritual is properly filled out. Three nits inline, all non-blocking; the docs one (`stranded-artifact` detector column) is the one I'd most like addressed before merge.

## Verified locally on `pr-670`

| Claim | Result |
|---|---|
| The cap valve was the only write-then-rename in `src/` | ✅ `renameSync` / `rename(` now appears only in `atomic-write.ts` |
| A new direct `renameSync` in `src/` fails CI | ✅ mutation-confirmed |
| An **aliased** import (`renameSync as mv`) is caught | ✅ the import line trips the regex |
| The `fs.promises` `rename` spelling is caught | ✅ mutation-confirmed (the PR body listed this as reasoned, not tested) |
| Store behavior on valve failure unchanged | ✅ read-only-dir path still warns once and degrades; a failing cleanup cannot mask the original error |
| Docs conventions | ✅ minor-ledger bugs stay out of the per-bug index, so `MINOR.md`-only for #641 is right; the `unbounded-trusted-payload` rider matches the table form |

One local test fails for me (`cap-valve and load-failure warnings suppress independently`) — my container runs as uid 0, where `chmod 0o555` doesn't stop writes. It fails identically on `main`. Not this PR.

## What's good

The ops-seam decision is the right call and is justified in both the code and the PR body — the `mock.module` failure mode (process-global, hangs on bun 1.3.9) is real, and a rename-only failure genuinely has no portable filesystem setup, so the alternative was leaving the failure path untested. The `cleanup itself throws → the ORIGINAL error still propagates` test is the one most people skip, and it's exactly the case the read-only-directory path hits in production. And the sweep is a real class-level ratchet rather than an instance regression test, which is what the bug ritual is asking for.

## Nits (inline)

1. `tests/core/write-then-rename-sweep.test.ts` — the `\brename\s*\(` alternative false-positives on prose comments and on any unrelated function named `rename(`; both mutation-confirmed to fail CI. Narrower regex suggested.
2. `docs/bugs/README.md` — the new class row's detector column claims coverage the sweep doesn't have for the temp-copy shape (#631). Suggest a `partial` qualifier, matching the honesty of the neighbouring rows.
3. `src/utils/atomic-write.ts` — docstring should note that mode isn't preserved and there's no `fsync`, now that this is the canonical helper. No current caller is affected.

Minor coverage note, no action needed: the write-throws test injects a `writeFileSync` that never creates a tmp, so cleanup of a *partially written* tmp isn't directly exercised — the rename test covers cleanup of an existing tmp, so the aggregate is fine.

---

tests/core/write-then-rename-sweep.test.ts
papiluxo: Adopted in dd2f40e — your regex verbatim, plus the failure-message suggestion (bun's two-arg `expect` works on 1.3.9 and prints the remedy). Re-ran your mutation table locally: `fsp.rename(` still trips the sweep, and both false-positive shapes (`// we rename (later)` and an unrelated `rename()` function) now pass. A comment in the test records why the bare `rename(` alternative is deliberately absent.

docs/bugs/README.md
papiluxo: Done in dd2f40e with your wording — `partial`, sweep confines write-then-rename only, temp-directory shape (#631) called out as instance-only. Agreed the neighboring rows set the standard here.

src/utils/atomic-write.ts
papiluxo: Added in dd2f40e — the docstring now scopes the guarantee (atomic against concurrent readers, no fsync, mode not preserved) and flags the 0o600-token-file case explicitly so a future auth-adjacent caller picks the helper with eyes open.

tests/core/write-then-rename-sweep.test.ts
ignaciohermosillacornejo: My suggested regex was wrong — it opened a blind spot the broad version covered, and I should have caught that before proposing it. Sorry for the round trip.

Dropping `\brename\s*\(` also drops a **destructured import from `fs/promises`**, which has no `renameSync` token anywhere to trip the first alternative. Re-ran the table on `dd2f40e`:

| mutation in `src/core/database.ts` | sweep |
|---|---|
| `renameSync("a","b")` | ✅ fails |
| `import { renameSync as mv }` … `mv("a","b")` | ✅ fails |
| `await fsp.rename("a","b")` | ✅ fails |
| `// we rename (later) the field` | ✅ passes (fixed) |
| `export function rename(a: string) {…}` | ✅ passes (fixed) |
| `import { rename } from "fs/promises"` … `await rename("a","b")` | ❌ **passes — undetected** |
| `import { rename } from "node:fs/promises"` … | ❌ **passes — undetected** |

The last two are the most natural way to write it in new async code, so it's the gap most likely to be hit. Adding an import-shape alternative closes it without bringing the false positives back:

```ts
if (
  /\brenameSync\b|\bpromises\s*\.\s*rename\b|\bfsp?\s*\.\s*rename\s*\(|import\s*\{[^}]*\brename\b[^}]*\}\s*from\s*['"](?:node:)?fs(?:\/promises)?['"]/.test(
    text
  )
) {
```

`\brename\b` inside the brace list won't match `renameSync` (the `S` blocks the trailing boundary), and requiring the full `import {…} from 'fs…'` shape means prose can't trip it. Verified against all seven mutations plus a multi-line import form:

| | narrowed (current) | + import alternative |
|---|---|---|
| MUT1–3 real spellings | ✅ caught | ✅ caught |
| MUT4–5 false positives | ✅ clean | ✅ clean |
| MUT6–7 destructured `fs/promises` | ❌ missed | ✅ caught |
| multi-line `import {\n rename,\n}` | ❌ missed | ✅ caught |
| real `src/` tree | ✅ green | ✅ green |

Confirmed separately that bun 1.3.9's two-arg `expect` does surface the message — a forced failure prints `error: write-then-rename must go through writeFileAtomic (src/utils/atomic-write.ts) — see #641` above the diff. That part is working as intended.

The other two threads are fully addressed; this is the only thing left from my side.

---

papiluxo: Confirmed and fixed in 5436893 — your import-shape alternative, adopted verbatim. No apology needed; the round trip is the process working.

I re-derived the gap before applying rather than taking it on faith, and it reproduces exactly as you describe: `import { rename } from 'fs/promises'` carries no `renameSync` token anywhere, so nothing else in the pattern could see it. Mutation-tested against real files in `src/` (not just the regex in isolation), all 8 shapes:

| mutation | sweep |
|---|---|
| `renameSync("a","b")` | ✅ fails |
| `import { renameSync as mv }` … `mv("a","b")` | ✅ fails |
| `await fsp.rename("a","b")` | ✅ fails |
| `// we rename (later) the field` | ✅ passes |
| `export function rename(a: string) {…}` | ✅ passes |
| `import { rename } from "fs/promises"` | ✅ fails |
| `import { rename } from "node:fs/promises"` | ✅ fails |
| multi-line `import {\n rename,\n}` | ✅ fails |

Clean tree green, full `bun run check` green (2806 pass). The test comment now records why the bare `rename(` alternative is absent *and* why the import shape can't reintroduce those false positives, so the next person doesn't re-litigate either half.

One residual I'll name rather than leave implicit: a commented-out real import — `// import { rename } from 'fs/promises'` — does trip the sweep. That seemed like the right side to err on (a commented-out rename import is worth a look), and unlike the earlier false positives it can't be produced by ordinary prose.

All three threads addressed from my side.

ignaciohermosillacornejo: Verified `5436893` independently — re-ran the mutation table against real files in `src/` rather than trusting the summary, and it reproduces exactly as you reported: all three original spellings plus both destructured forms and the multi-line import fail CI; the prose comment and the unrelated `rename()` function stay green; clean tree green.

Two extra probes on the residual you named:

- `// import { rename } from 'fs/promises';` — confirmed it trips the sweep. Agreed this is the right side to err on, and thanks for surfacing it rather than letting me find it.
- `// we should rename the tag from fs to something else` — stays green. So the residual really is confined to commented-out *import syntax*; ordinary prose that happens to contain both "rename" and "fs" can't reach it. That's the distinction that makes erring this way cheap.

Also diffed the full failing-test set between this branch and `main` (my container has a broken install, so both fail en masse — the comparison is the only meaningful signal): **identical, 72 to 72**. No test regresses.

All three threads addressed; resolving them. No further comments from me — the remaining call on the PR is yours.

---

ignaciohermosillacornejo (approved):
LGTM 🚀

All three review threads addressed and verified on `5436893`. The fix is right, the sweep is an honest class-level ratchet with its limits documented rather than papered over, and the failure-path cleanup is tested through every branch including the one where cleanup itself throws.

Nice work on the second round in particular — you re-derived the blind spot before applying the fix rather than taking my word for it, and you surfaced the commented-out-import residual yourself instead of leaving it to be found.

---

Issue #691, opened 2026-08-30
stripComments in exported-constants discovery is not string-aware: a URL literal drops its constant


ignaciohermosillacornejo:
`discoverStringConstants` strips comments with a helper that is not string-aware, so a
string literal containing `//` or `/*` silently removes its constant from discovery — the
same class as the inline-comment hole just fixed, one step narrower.

## Where

`tests/exported-constants.test.ts` — the local `stripComments` helper added alongside the
`collectStringConstants` extraction.

## The gap

The helper replaces comment syntax without tracking whether it is inside a string literal.
Probed during review:

- `export const URLS = ['https://example.com', 'b'] as const` discovers **nothing** — the
  `//` in the URL is treated as a line comment and eats the rest of the declaration.
- An unbalanced `/*` inside a literal swallows forward to the next `*/`, dropping whatever
  constant it lands in.

## Why it is not urgent

No such literal exists in `src/` today, and the limitation is stated in the helper's doc
comment rather than left implied.

Note the direction, though — it is asymmetric:

- An **already-pinned** constant vanishing from discovery is caught: the backward test
  asserts every pinned expectation still names a live constant.
- A **new** constant whose literal contains `//` is silently missed. Nothing fails. That is
  the same shape as the bug this guard was just repaired for: discovery under-reports, and
  an under-reporting pin is indistinguishable from a passing one.

## Suggested fix

Consume string literals in the same alternation as the comment branches, and only replace
on the comment branches — so a literal is matched and passed through intact rather than
scanned for comment syntax.

## Provenance

Raised in task review of the PR closing #677; deferred there deliberately to keep that PR
scoped to the live hole.

PR #693, opened 2026-08-30
fix(test): make exported-constant discovery model the grammar

### LINKED ISSUE #677: [Audit] Unaddressed suggestions from PR #676
## Unaddressed review suggestions from PR #676

**PR title:** test(field-selection): discover presets instead of listing them

### HIGH: discoverStringConstants does not strip inline comments before parsing, so KNOWN_FREQUENCIES and IGNORED_ITEM_FIELDS escape discovery and are absent from PINNED.

- **File:** `tests/exported-constants.test.ts`:89
- **Reviewer said:** > Discovery is blind to any array containing an inline comment — 2 of 25 constants escape today: KNOWN_FREQUENCIES in src/models/recurring.ts:14, IGNORED_ITEM_FIELDS in src/models/item.ts:129
- **Evidence not applied:** Line 89 still uses rawBody.replace(/'[^']*'|,|\s/g, '') without stripping // comments first; KNOWN_FREQUENCIES (recurring.ts:14, six inline comments) and IGNORED_ITEM_FIELDS (item.ts:129, three inline comments) are absent from PINNED.

### MEDIUM: EXPORTED_ARRAY regex uses lazy (.*?) dotall so a non-as-const array can swallow every as-const array after it; this fires today at src/conformance/ledger.ts:305.

- **File:** `tests/exported-constants.test.ts`:67
- **Reviewer said:** > The lazy `.*?` lets a non-`as const` array swallow the ones after it — and it is firing today. Live example: src/conformance/ledger.ts:305 scans forward to the next `] as const` at line 493.
- **Evidence not applied:** Line 67 still reads /^export const ([A-Z][A-Z0-9_]*)\s*(?::[^=]+)?=\s*\[(.*?)\]\s*as const/gms; src/conformance/ledger.ts:305 declares CONFORMANCE_LEDGER: readonly LedgerEntry[] = [ with no as const, and line 493 has the next ] as const.

---

ignaciohermosillacornejo:
# Summary

`tests/exported-constants.test.ts` pins every exported string-literal array in `src/` so a
member cannot be added or removed unnoticed. Its discovery step could not see past a `//`,
so two constants were never in the pinned set at all — the guard reported success over a
set it had silently under-collected.

Closes #677

## What was wrong

**Live:** discovery ran on raw source, so any array containing an inline comment failed the
"is this purely string literals?" residue check and was dropped. Two real constants were
outside the pin: `KNOWN_FREQUENCIES` (`src/models/recurring.ts`, six inline comments) and
`IGNORED_ITEM_FIELDS` (`src/models/item.ts`, three). Discovery now covers 25 constants, up
from 23.

**Latent:** the array-body matcher used a lazy dotall `(.*?)`, so an array without `as const`
could scan forward to the next `] as const` and swallow the declarations between.

The audit finding was **correct about the mechanism**, and an earlier draft of this PR said
otherwise — that draft is wrong and has been corrected. Tracing the old matcher from
`src/conformance/ledger.ts:305`: the `as const` arrays at `:59` and `:78` are behind the sweep
cursor by then, so the first `] as const` ahead is the *inline*
`(['id', 'accountId', 'itemId'] as const)` at `:493`. The match fires, spanning 305 to 493.

Nothing is lost from the pin only because that swallowed span contains no exported
string-literal `as const` **declaration** — verified with
`awk 'NR>=305 && NR<=493 && /^export const/'`, which returns line 305 itself and nothing else.
That is an accident of today's tree, not a property of it. So the fix is hardening rather than
a live-hole repair, but the reason is "the swallowed span happens to be empty of declarations",
not "the match never happens".

## Bug fix?

Root cause:       Discovery parsed raw source, so a `//` inside an array literal made the residue check reject the whole declaration and drop the constant.
Bug class:        A discovery regex that under-reports silently — the pinned set is derived from a scan that cannot see everything it claims to cover, so a passing guard and an incomplete guard are indistinguishable.
Detector added:   `collectStringConstants(source)` split out of the tree walk, with tests that run it against synthetic snippets: one pinning that inline comments are seen, one pinning the declaration boundary. Both fail if their fix is reverted.
Siblings checked: `tests/core/decoder-field-completeness.test.ts` already strips comments before its census scan — but only for that one scan; its other scanners still run on raw source. That is the same class and is being fixed separately in this batch (#685). No other discovery-style scanner in `tests/` parses source without stripping.

## External assumptions

None

## Verification

- `bun run check` — 3024 pass / 20 skip / 0 fail
- Both new `PINNED` entries compared member-for-member against their source arrays; exact match, same order. No pre-existing entry changed.
- 25 confirmed as the correct total by independent enumeration: 33 exported uppercase array constants in `src/`, 8 legitimately outside the grammar (not pure string-literal `as const`).
- **Mutation-checked, both directions.** Comment-stripping neutered → 3 tests red. Boundary matcher reverted to `(.*?)` → 1 test red. That second mutation is the reason for the `collectStringConstants` split: neither of the two tests originally specified for this fix detects it, so without the split this PR would have shipped a change that survives its own deletion — in a PR about a guard that survived its own deletion.

## Known limitation

The comment stripper is not string-aware, so a literal containing `//` (a URL) would drop
its constant from discovery. No such literal exists in `src/` today and the assumption is
stated in the helper's doc comment. Tracked as #691 rather than expanded here.

### DIFF
--- a/tests/exported-constants.test.ts
+++ b/tests/exported-constants.test.ts
@@ -486,7 +486,9 @@ describe('exported string constants are pinned (#635 class detector)', () => {
   test('a non-as-const array is not itself discovered', () => {
     // CONFORMANCE_LEDGER (src/conformance/ledger.ts) is `readonly LedgerEntry[]`
     // with no `as const`; it must not consume the declarations that follow it.
-    // A shape pin only — the detector for the rule is the synthetic test below.
+    // A shape pin only. The detectors are the synthetic snippets below: the
+    // boundary one for the body bound, and the semicolon-free one for the
+    // annotation bound.
     expect(discovered.has('CONFORMANCE_LEDGER')).toBe(false);
   });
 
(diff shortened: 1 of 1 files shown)


ignaciohermosillacornejo:
Confirming for the reviewer, which has flagged across six rounds that it cannot reach the API from its job: **all four tracking issues referenced in this file exist and are open.**

| ref | issue | hazard |
|---|---|---|
| `#691` | [stripComments is not string-aware](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/issues/691) | a `//` inside a string literal drops the constant |
| `#694` | [discovery keys by name alone](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/issues/694) | cross-file name collision, last-write-wins |
| `#695` | [Set-declared allowlists are outside the grammar](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/issues/695) | `TRANSFER_CATEGORIES` / `INCOME_CATEGORIES` unpinned |
| `#696` | [the bracket bound and the quote-style drop](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/issues/696) | both modes, one root cause, one fix |

All four are instances of one class — discovery that under-collects while reporting success — which is what makes this PR's subject the instance that happened to get filed rather than the whole problem. #696 covers two modes because a single change closes both: consume string literals as literals in the scanning alternation, instead of bounding on bracket and quote characters.

PR #724, opened 2026-09-14
fix(guards): a security gate that vouched for files it never opened, and a fields fragment with no typo guard

### LINKED ISSUE #701: [Audit] Unaddressed suggestions from PR #698
## Unaddressed review suggestions from PR #698

**PR title:** fix(security): close five live bypasses of the concealment gate

### LOW: The new `gitDecline` declaration was inserted between the `gitFiles` JSDoc and `gitFiles` itself, re-stranding the two docstrings the way commit 81307f3 had already fixed once in this stack.

- **File:** `scripts/check-concealment.ts`:762
- **Reviewer said:** > JSDoc attaches to the next declaration, so the ~50-line docstring about SKIP_DIRS, `.gitignore` scope and the walk fallback now documents
- **Evidence not applied:** Reading scripts/check-concealment.ts:733-764 shows the gitFiles JSDoc (733-753) immediately followed by the gitDecline JSDoc+declaration (754-762) and then `function gitFiles(...)` at 764 — each docstring still sits on the wrong declaration.

### LOW: The scan loop's `catch { continue }` on an unreadable file silently drops it from the scan while `files.length` (computed before the loop) still counts it as scanned in the summary output.

- **File:** `scripts/check-concealment.ts`:1148
- **Reviewer said:** > A file that cannot be read is dropped from the scan with no output at all.
- **Evidence not applied:** scripts/check-concealment.ts:1144-1150 still has `try { contents = readFileSync(file, 'utf-8'); } catch { continue; }` with no logging or skip-count, and files.length at lines 1208/1214 is unchanged, so an unreadable file is still counted as scanned.

---

### LINKED ISSUE #709: Guard the accounts fields fragment against typo'd names, the way #707 guarded the transactions one
`TRANSACTION_FIELDS_PARAM_SCHEMA` and `ACCOUNT_FIELDS_PARAM_SCHEMA` are each shared
verbatim between a cache tool and its `_live` counterpart, so anything their
description text names ships into **both** schemas. PR #707 found that the
transactions fragment had been rewritten to enumerate the cache document: 16 of the
27 field names it listed have no live equivalent, and its worked example was
`user_note`, whose live spelling is `user_notes` — so a live caller copying the
parameter docs got a `_field_warning` instead of a field.

That fragment is now guarded by "the SHARED fragment names only fields BOTH modes
have" in `tests/tools/live/transactions.test.ts`, which asserts two things about
every identifier in the shared text:

1. **mode parity** — a name only one surface has is a failure;
2. **the typo half** — an underscored name that is a field on *neither* surface is a
   failure.

`ACCOUNT_FIELDS_PARAM_SCHEMA` (`src/tools/registry/accounts-system.ts`,
`src/tools/live/accounts.ts`) has neither.

**Half of the rule transfers and half does not.** The mode-parity half **cannot** be
lifted: that fragment names `current_balance` and `balance` deliberately, labeled per
mode, because the two account presets share no field names — its own docstring says
so. The **typo half transfers unchanged**: an underscored name in that description
that is in neither `ACCOUNT_KNOWN_FIELDS` nor `ACCOUNT_LIVE_KNOWN_FIELDS` ships into
both schemas today with nothing watching.

Worth also deciding whether a per-mode-labeled name should be *checked against the
mode it is labeled for*, which is the stronger version and the one that would catch a
`current_balance`/`balance` swap.

Raised in review on #707 and deliberately left out of that PR — it was at round seven
and the finding is latent, not broken.

ignaciohermosillacornejo:
# Summary

Two guards that did less than they appeared to.

`scripts/check-concealment.ts` dropped an unreadable file with a bare `catch { continue }` while the summary printed a count computed *before* the loop — so a **security** gate reported a file it never opened as scanned. Same file, second finding: PR #698 re-stranded a JSDoc block that commit `81307f3` had already un-stranded once.

`ACCOUNT_FIELDS_PARAM_SCHEMA` ships verbatim into `get_accounts` and `get_accounts_live` and was guarded only by schema *equality* — which pins that both modes say the same thing, not that the thing is true. A typo'd field name was identical in both schemas and wrong in both.

Closes #701
Closes #709

## External assumptions

None. Nothing here touches Copilot's API, data shapes, or enum values — one gate script, one parameter-description guard, and four docblock moves.

## Bug fix?

Root cause:       The scan loop's `catch { continue }` had no counter and no output, while the summary printed `files.length` — captured before the loop — so a file that could not be read was reported as one that was scanned; separately, a JSDoc block documents whatever declaration follows it, and inserting a docblock + declaration into the gap silently re-points it.
Bug class:        A number reported about work that was computed before the work, so it counts what the work skipped; and a docblock separated from its subject by an insertion, which compiles, lints and formats clean.
Detector added:   `tests/docs/jsdoc-stranding.test.ts` — sweeps every `.ts` in `src/`, `scripts/` and `tests/` for a comment that closes on one line and a docblock that opens on the next, with non-vacuity floors on files and delimiters seen. For the count: the printed number is now incremented at the point of inspection, so it cannot overstate by construction, and `tests/scripts/check-concealment.test.ts` pins both halves (the file is named with its measured cause, and the two counts are printed side by side).
Siblings checked: The stranding sweep found three more instances the moment it was written — a second site in `scripts/check-concealment.ts` (the `answersToName` docstring), `src/tools/tools.ts` (`filterSplitParents`), and `src/models/account.ts` (`AccountSchema`). All four fixed here; the tree is now clean, so the detector lands with no exemption list. For the fail-open class: the only other pre-loop count in this script is the listing itself, which the strategy check already refuses on.

---

## #701 — the fail-open, and why symlinks had to move first

Making the refusal honest required one behaviour change. git tracks a symlink as mode `120000` whose blob is the **target path**, so the target path is the whole of what a reviewer is shown for one. This repo tracks two symlinks that point at directories (`.agents/skills`, `.claude/skills`); `readFileSync` followed both into an `EISDIR` and the silent `continue` swallowed them. **Those two files were being dropped from every single run** — which is exactly why the bug stayed invisible. They are now scanned as their target strings, which is also what the diff contains, so `readFileSync` is still used for every link that resolves to a readable file (the `package.json` -> `manifest.json` routing test depends on that).

With that in place, an unreadable file is a **refusal** rather than a note, matching the posture the walk fallback in the same file already takes: *"refusing rather than reporting a green run over a shrunken scan."* The message names the file, the measured cause (not a guess at it), and both counts.

Findings are printed *before* the unreadable block and never instead of it — a run that hit both has to show both.

## #709 — the typo guard, and the tool-name exemption

The typo half of #707's rule transfers; the mode-parity half cannot, because this fragment names `current_balance` and `balance` deliberately, one per mode, since the two account presets share no field names at all.

**A verbatim port fails immediately.** The fragment names four underscored identifiers: `account_id` and `current_balance` (cache fields) and `get_accounts` / `get_accounts_live`, which are **tool names** — fields on neither surface. The transactions fragment happens to name no tool, which is why that guard never needed an exemption.

The exemption is derived from `TOOL_REGISTRY` — the registry's own name→definition map, the same one the server dispatches from — rather than a pair of string literals. Mutation-verified: renaming the live tool makes the now-stale doc reference **fail** the guard, which a hardcoded pair would have stayed green through.

**Explicitly deferred:** the stronger per-mode-labeled variant (parsing `for get_accounts` / `for get_accounts_live` out of the prose and checking each labeled name against the mode it is labeled for, which is what would catch a `current_balance`/`balance` swap). Not attempted here; noted in the test.

## Mutation testing

Every guard touched, run in both directions. Precedent: `docs/bugs/596-vacuous-assertions-bulk-edit.md` — a test can execute a guard and still not detect its deletion.

| # | Mutation | Expected | Result |
|---|---|---|---|
| M1 | restore `catch { continue; }` | all 4 new concealment tests fail | 4 fail ✅ |
| M2 | print the listed count instead of the read count | "count is what was read" fails | 1 fail ✅ |
| M3 | delete the symlink fallback | symlink tests **and the pre-existing real-repo test** fail | 3 fail ✅ |
| M4 | typo a field name in the accounts fragment | typo guard fails, naming the misspelling | fail ✅ |
| M6 | drop the tool-name exemption | guard fails on the two tool names | fail ✅ |
| M8 | make the identifier regex match nothing | non-vacuity floors fail | fail ✅ |
| M9 | rename `get_accounts_live` in its schema | guard fails on the stale doc reference | fail ✅ |
| M10 | re-strand the `AccountSchema` docblock | stranding detector fails | fail ✅ |
| M11 | reach `linkTarget` from the `catch` only (round-1 state) | resolvable-symlink test fails | fail ✅ |
| M12 | typo `balance` in the **live** worked example | typo guard fails, naming `balnce` | fail ✅ |
| M14 | typo `id` in the live worked example | typo guard fails, naming `idd` | fail ✅ |
| M15 | re-strand a **one-line** docblock | stranding detector fails | fail ✅ |
| M16 | narrow the detector back to bare delimiters, with M15 still applied | *stays green* — the honest mutation, and exactly the gap finding 3 named | pass (documents the gap) |
| M17 | revert the `contents !== link` conjunct on `prose` | dangling-`.md` test fails | fail ✅ |
| M18 | delete the one-line arm of `closingKind` | the one-line floor fails (it did NOT before round 3) | fail ✅ |
| M19 | put a dangling `*.ts` symlink under a scanned tree | named, informative failure instead of a module-scope crash | fail ✅ |
| M20 | break `OPENS_DOCBLOCK` so it never matches | the opens floor fails (it did NOT before round 4) | fail ✅ |
| M21 | drop `scripts` from `SCANNED_TREES` | the pinned tree list fails — **the first fix for this passed it**, see below | fail ✅ |
| M22 | delete the one-line arm of `closingKind` | floor **and** positive control fail | fail ✅ |
| M23 | delete the line that accumulates the rule's findings | end-to-end control fails — **floors and unit controls did not see this**, see below | fail ✅ |
| M24 | truncate the pair loop to one iteration | floors and positive control fail (the controls' hits now sit off index 0) | fail ✅ |
| M25 | replace the sweep's unreadable collector with `catch { continue }` | the unreadable control fails — **every other assertion stayed green**, see below | fail ✅ |

M4 initially failed on a *non-vacuity floor* rather than on the guard itself, reporting `expected >= 2, received 1` instead of the misspelled name — the guard detecting the right thing and reporting the wrong one. The floors now run **after** the main assertion; re-verified in both orders.

## Review round 1 — all three findings addressed

**1. The symlink principle was applied only in the failure path.** Correct, and the same class as the bug being fixed pointing the other way: content inspected ≠ content listed. `linkTarget` is now called for **every** symlink and its target path is checked through `checkLine` whether or not the link resolves — ahead of the NUL early-return (a link to a `.png` still has an attacker-chosen target) and with `prose: false`, because a target path is a path, not a paragraph. Following the link is still how a resolvable one is read, so the `package.json` -> `manifest.json` routing test is untouched. New test: a file whose **name** is the payload, symlinked to, so the bytes are clean and the finding can only come from checking the link string (M11).

**2. The accounts typo guard could not see the live example.** Correct, and it was the cheapest false green available — cheaper than anything my own analysis listed. The underscore filter went blind to `["id", "name", "balance"]` entirely, including `balance`, one of the two per-mode names the test's own docstring singles out. Quoted names inside bracketed examples are now extracted directly (`/\[[^\]]*\]/` then `/"([a-z][a-z0-9_]*)"/g`): six references, zero English words, no reliance on a naming convention only cache mode follows. The underscore sweep stays for prose-level names. Mutation-verified on both previously-blind names (M12, M14).

**3. The stranding detector missed the one-line idiom.** Correct — and not hypothetical. Widening it found a **fifth** instance immediately, in `tests/core/decoder-field-completeness.test.ts:221`, where a one-line summary sat above an `ASSUMPTION` block added later. Both described the same function, so that one is merged rather than moved. `statSync` -> `lstatSync` as suggested. The non-doc-block-comment false-positive shape is now stated in the docblock with the remedy (a blank line, which is the idiom anyway).

**Nits:** the "files were actually read" wording overstated what happens to a link and now says *inspected*. `withGitTree`'s positional params left as-is per your own note that it isn't worth the churn. The local-friction trade is deliberate and now has a test of its own — a tracked file deleted without staging the deletion hard-fails `check:concealment` locally, with the remedy named in the message; CI is a clean checkout and is unaffected.

**On your closing point:** you were right that finding 2 was cheaper than any route I listed. The updated analysis below reflects it.

## Review round 2 — finding addressed

**`prose: false` applied only to links that resolve.** Correct, and it is the same "stated but half-applied" shape as the thing the commit before it fixed. The dedicated link-target `checkLine` is skipped when the target *is* the contents — the dangling case — so that path fell through to the generic loop's `isProse(rel)`, and for a link named `*.md` / `*.txt` / `*.rst` that dropped MAX_LINE and DYNAMIC_EXEC from a string the attacker wrote. The **dangling** case, whose target need not correspond to anything in the tree, was being checked *less* strictly than the resolving one.

Fixed as suggested — `isProse(rel) && contents !== link`, not `false` for every symlink, because `AGENTS.md` resolves to `CLAUDE.md`, whose prose lines run past MAX_LINE and would manufacture findings on this repo's own tree. Prose is a property of the bytes being scanned, not of the path. New test: the existing fallback fixture with the link named `link.md` — same payload, same assertion, one extension apart. M17 pins it.

**Nits, all taken:** the two `toContain('link')` assertions now pin `link:1`; the anticipatory `default`/`all` filter says it is anticipatory, since every other filter in that test is load-bearing; the lstat-per-listed-file cost (~549 on this repo) and the tiny stat→lstat scan shrink in the stranding walker are both stated in the code rather than left to be discovered. The three-times-scanned `CLAUDE.md` observation is pre-existing and left alone.

## Review round 3 — both findings addressed

**1. The round-2 widening had no floor of its own.** Correct, and it is this PR's own subject turned on its own guard. `closesAComment` was a disjunction floored only on the sum: deleting the one-line arm took closings from 1297 to 999, still far over the 500 floor, so the widening that closed round-1 finding 3 could be removed with the suite green and the detector quietly back to the round-1 rule. It returns the **kind** now, the two spellings are counted separately, and each has its own floor naming how its input can go to zero (999 bare / 298 one-line as this lands). M18 pins it — and M18 fails now where it would have passed before.

**2. The `lstat` fix was half-applied.** Also correct, and the docblock asserted the robustness property rather than the code having it — `doc-reality-drift` inside the file written to detect the in-code form of it. `lstat` stops the *walker* throwing, but a dangling link named `*.ts` is not a directory, lands in the file list, and throws at the read — uncaught, at module scope, since `scan()` runs there. Unreadable files are now collected and asserted on, which is the posture `check-concealment.ts` takes about its own scan in this same PR. M19 pins it.

**Nits:** the `MINOR.md` row said three instances across `src/` (it is four, across `src/`, `scripts/` and `tests/`) — a `doc-reality-drift` row drifting, which is the joke writing itself. And the prose conjunct now states the asymmetry it leaves: for a link that **resolves**, `prose` comes from the *link's* extension while the bytes are the *target's*, safe only because the git-visible content (the target path) gets code rules unconditionally and an in-tree target is scanned strictly under its own path — two conditions that have to hold together, so they are written down together.

## Review round 4 — both findings addressed, and one of my own fixes failed its mutation

**1. The opening delimiter had no floor, and the detector had no positive control.** Correct. Both closing counters are incremented *before* the `OPENS_DOCBLOCK` test, so breaking that constant left every assertion green and the sweep reported zero for the same reason a clean tree does. Floors could not have fixed it on their own — they measure inputs, and a detector observed only returning `[]` has not been observed detecting. The pair loop is extracted as `strandedLines` and has a positive control (both closing spellings) and a negative one (the blank line the header docblock's adjacency argument turns on), every fixture assembled from the delimiter constants so the file still cannot match itself. The opening delimiter has its own floor too (M20).

**2. `SCANNED_TREES` floored only in aggregate.** Correct, and `scripts/` — where the originating bug lives — was one of the two trees that could leave silently.

**The part worth reporting: my first fix for finding 2 failed its own mutation.** I wrote `for (const tree of SCANNED_TREES) expect(result.perTree[tree]).toBeGreaterThan(30)`, ran M21, and it **passed** — a loop over `SCANNED_TREES` cannot see a tree removed *from* `SCANNED_TREES`. A guard whose domain is the thing it guards has no domain. The list is now pinned with a literal (`expect([...SCANNED_TREES].sort()).toEqual(['scripts', 'src', 'tests'])`), which is what makes the per-tree floors reachable at all, and M21 fails as it should. That is the whole argument for mutating in both directions rather than reasoning about it, so it is recorded rather than quietly corrected.

**Nits:** `files++` moved below the read, so the counter obeys the same "increment at the point of inspection" rule this PR's gate fix rests on instead of relying on a neighbouring test to contradict it; and `ONE_LINE_DOCBLOCK`'s greedy `.*` is documented as the mirror of the header's known false-positive shape.

## Review round 5 — finding addressed, and it needed one more step than proposed

**The floors and controls measured a loop that wasn't the rule.** Correct. Extracting `strandedLines` had left `scan()` with two traversals of the same lines: every floor read the counting loop while the rule lived in the other one, so the rule's *use* was unguarded even though the rule itself was controlled. One pass now, counts returned alongside the hits, so each floor is evidence the rule ran over the tree. The controls' hits also moved off index 0, so a loop truncated to one iteration fails them (M24).

**The proposed fix did not cover the narrowest mutation.** Deleting only the line that *accumulates* the rule's output — the rule runs, the counts are recorded, the findings are dropped — left every floor and both unit controls green, because over a clean tree "found nothing" and "threw the findings away" are the same observation. So the sweep is parameterised on its root and there is an **end-to-end control**: walk, read, rule and accumulate, run over a synthetic tree built to hold exactly one stranded file. M23 fails on it.

**Nits:** the "roughly 3x" margin claim was false — measured 1.6x to 2.3x — and now states the numbers instead of a comfortable rounding of them; the pinned tree list says it fails on an *added* tree too, which is the intended direction (a new tree arrives with its own floor rather than sliding under the aggregate).

**Rebased onto current main** (#716 and #721), resolving `docs/bugs/MINOR.md` by keeping both sides. #721's new `no-hand-rolled-comment-strippers` detector passes against this branch unchanged — the concealment gate matches comments with `git check-attr`-style path parsing, not with a comment-matching regex. The rebase moved the tree from 336 files to 340, so every stated count in the stranding test was re-measured.

## Review round 6 — finding addressed

**The `unreadable` collector had never been observed collecting.** Correct, and it is the round-4 argument about `stranded` applied to the list beside it. Both assertions on it were `toEqual([])` and nothing ever fed it a file it could not read — so replacing the wrapped read with the literal `catch { continue }` this PR exists to remove left every assertion in the file green, `unreadable` included, because an empty list is also what a clean tree produces. M19 had been covering it by hand since round 3, which is the relationship round 3 itself named about M15: a manual mutation is not a test.

The control is the case the `lstat` docblock already *claimed* was handled — a dangling symlink named `*.ts`, which lstat admits to the file list and which fails at the read — so the claim is now an assertion. It also pins that the failure is scoped to the file that caused it: the readable sibling in the same synthetic tree is still swept.

## Cheapest way to turn these guards green wrongly

- **#709 (round 1, now closed):** the cheapest route needed no deletion at all — a typo in `"id"`, `"name"` or `"balance"` was simply outside where the guard looked. Closed by extracting the bracketed examples; M12/M14 pin both.
- **#709:** add the typo'd name to `ACCOUNT_KNOWN_FIELDS` — but that set is `Object.keys(AccountSchema.shape)`, so it means adding the field to the Zod model, which `tests/tools/live/known-fields-wire-parity.test.ts` and the decoder's coverage gate both watch. Second cheapest: register a tool named after the typo, which the registry's duplicate-name check and `check:tool-counts` both see. Third: delete the guard, which the stranding-style sweeps do *not* catch — no gate in this repo detects a deleted test, and that is true of every test here.
- **#701:** widen the refusal back to a warning. Not gated beyond review — but the count itself now is: it is incremented at the point of inspection, so making it overstate again requires reintroducing a pre-loop total rather than merely deleting a branch. The `chmod 000` path is the one a person could quietly special-case; the test asserts on the cause string (`EACCES`/`EPERM`) rather than only on exit code, so a blanket "ignore permissions errors" exemption fails it.
- **The stranding detector:** add a blank line between the two docblocks. That is not a false green — a blank line is the file-header idiom, and the block below a blank line still documents the declaration after it. The rule is adjacency on purpose.

## Verification

- `bun run check` → exit 0 (3249 pass, 20 skip, 0 fail)
- `bun run scripts/check-concealment.ts` on this repo → `548 files scanned (git), nothing hidden`, stderr empty — same count as before, with the two symlinks now genuinely inspected rather than counted-and-skipped

Note: `scripts/` is outside `format:check`'s globs (`src/**`, `tests/**`), so `scripts/check-concealment.ts` is deliberately left in its existing hand-formatting — running Prettier over it added ~300 lines of unrelated array reflow.


ignaciohermosillacornejo:
All three findings addressed in `7963385` (rebased onto current main).

**1 — symlink principle half-applied.** Right, and it is the same class as the bug being fixed pointing the other way. `linkTarget` is now called for every symlink and the target path goes through `checkLine` whether or not the link resolves, ahead of the NUL early-return (a link to a `.png` still has an attacker-chosen target) and with `prose: false`, since a target path is a path rather than a paragraph. Following the link is still how a resolvable one is read, so the routing test is untouched. Test fixture: a file whose *name* is the payload, symlinked to — bytes clean, so the finding can only come from checking the link string. Mutation M11 (reach `linkTarget` from the catch only) fails it.

**2 — the typo guard could not see the live example.** Right, and cheaper than anything my own false-green analysis listed: no deletion needed, the guard simply did not look there. Quoted names inside bracketed examples are now extracted directly (`/\[[^\]]*\]/` then `/"([a-z][a-z0-9_]*)"/g`) — six references, zero English words, and no reliance on a naming convention only cache mode follows. The underscore sweep stays for prose-level names. Mutation-verified on both names that were blind: `balance`→`balnce` and `id`→`idd` each fail now.

**3 — one-line docblocks.** Right, and not hypothetical: widening found a **fifth** instance of the class immediately, `tests/core/decoder-field-completeness.test.ts:221`, where a one-line summary sat above an ASSUMPTION block added later. Both described the same function, so that one is merged rather than moved. `statSync`→`lstatSync` as suggested, and the non-doc-comment false-positive shape is stated in the docblock with its remedy.

Nits: "files were actually read" now says *inspected*. `withGitTree`'s positional params left alone per your own note. The local-friction trade is deliberate and now has its own test — a tracked file deleted without staging hard-fails locally, with the remedy named in the message.

On your closing point: agreed, finding 2 was the cheapest green of all of them, and the updated analysis in the description says so.

ignaciohermosillacornejo:
Round-2 finding addressed in `6e7f1e5`.

**The `prose: false` rule applied only to links that resolve.** Correct, and it is the same stated-but-half-applied shape as the thing the previous commit fixed — which is uncomfortable and worth saying out loud. The dedicated link-target `checkLine` is skipped when the target *is* the contents, so the dangling case fell through to the generic loop's `isProse(rel)`, and a link named `*.md` lost MAX_LINE and DYNAMIC_EXEC on a string the attacker wrote. The dangling case, whose target need not correspond to anything in the tree, was checked *less* strictly than the resolving one.

Fixed as you suggested: `isProse(rel) && contents !== link`, not `false` for every symlink — `AGENTS.md` resolves to `CLAUDE.md`, whose prose lines run past MAX_LINE, so the blanket version would manufacture findings on this repo's own tree. Prose is a property of the bytes being scanned, not of the path. New test is your fixture: the existing dangling-link case with the link named `link.md`, same payload, same assertion, one extension apart. Mutation M17 (revert the conjunct) fails it.

Nits all taken: `toContain('link')` → `toContain('link:1')` in both places; the `default`/`all` filter now says it is anticipatory rather than load-bearing, and notes it would exempt a real field with either name; the lstat-per-listed-file cost (~549 here) and the stat→lstat scan shrink in the stranding walker are both written down in the code. The triple-scan of `CLAUDE.md` via `AGENTS.md`/`GEMINI.md` is pre-existing and left alone — happy to file it if you'd rather it not stay tacit.

ignaciohermosillacornejo:
Round-3 findings addressed in `9216e95`.

**1 — the round-2 widening had no floor of its own.** Correct, and it is this PR's own subject turned on its own guard, which is not a comfortable thing to have pointed out. Your arithmetic was right: dropping the one-line arm took the count from 1297 to 999, still far over 500, so the widening that closed round-1 finding 3 could have been deleted with the suite green and the detector quietly back to the round-1 rule. `closesAComment` is now `closingKind`, returning `'bare' | 'oneLine' | undefined`, the two are counted separately, and each has its own floor (999 / 298 measured). M18 — delete the one-line arm — fails now where it passed before.

**2 — the lstat fix was half-applied.** Also correct, and worse than a gap: the docblock asserted the robustness property rather than the code having it, inside the file written to detect the in-code form of exactly that. A dangling `*.ts` link is not a directory, so lstat lets it into the file list and the read throws at module scope. `scan()` now collects unreadable files and a test asserts the list is empty — the same posture `check-concealment.ts` takes about its own scan in this PR. M19 (real dangling `*.ts` link under `tests/`) produces the named failure instead of a raw ENOENT crash.

**Nits:** `MINOR.md` said three instances across `src/`; it is four, across `src/`, `scripts/` and `tests/` — a `doc-reality-drift` row drifting. Fixed. And the prose conjunct now states the asymmetry it leaves, in the terms you put it: for a link that resolves, `prose` comes from the link's extension while the bytes are the target's, which is safe only because the target path gets code rules unconditionally *and* an in-tree target is scanned strictly under its own path — two conditions that have to hold together, so they are written down together, with the out-of-tree case named as where it would stop being safe.

Thanks for three rounds of this — findings 1 of round 2 and round 3 were both cheaper false-greens than anything my own analysis had listed.

ignaciohermosillacornejo:
Round-4 findings addressed in `e9748e5` — and one of them caught a bug in my own fix, which is worth reporting rather than quietly correcting.

**1 — the opening delimiter had no floor, and the detector had no positive control.** Correct. Both closing counters increment before the `OPENS_DOCBLOCK` test, so breaking that constant left everything green and the sweep reported zero for the same reason a clean tree does. Your point that floors alone cannot fix it is the important half: they measure inputs, and a detector observed only returning `[]` has not been observed detecting. The pair loop is extracted as `strandedLines`, with a positive control covering **both** closing spellings and a negative control for the blank-line case the header's adjacency argument turns on — fixtures assembled from the constants so the file still cannot match itself. The opening delimiter has its own floor as well. M20 (break `OPENS_DOCBLOCK`) and M22 (delete the one-line arm) both fail now; M22 fails twice, once on the floor and once on the control.

**2 — `SCANNED_TREES` floored only in aggregate.** Correct, and `scripts/` — where the originating bug lives — was one of the two that could leave silently.

**My first fix for 2 failed its own mutation.** I wrote `for (const tree of SCANNED_TREES) expect(result.perTree[tree]).toBeGreaterThan(30)`, ran M21 (drop `scripts`), and it **passed**: a loop over `SCANNED_TREES` cannot see a tree removed *from* `SCANNED_TREES`. A guard whose domain is the thing it guards has no domain. The list is pinned with a literal now — `expect([...SCANNED_TREES].sort()).toEqual(['scripts', 'src', 'tests'])` — which is what makes the per-tree floors reachable at all, and M21 fails as it should. The code comment records the measurement rather than the conclusion.

**Nits:** `files++` moved below the read, so the counter obeys the same increment-at-the-point-of-inspection rule this PR's gate fix rests on instead of depending on a neighbouring test to contradict it. `ONE_LINE_DOCBLOCK`'s greedy `.*` is documented as the mirror of the header's known false-positive shape. And thank you for re-measuring the stated counts against the tree — that check is the one I would most expect to rot.

ignaciohermosillacornejo:
Round-5 finding addressed in `ee5955e`, plus a rebase onto current main.

**The floors and controls measured a loop that wasn't the rule.** Correct — extracting `strandedLines` left `scan()` with two traversals, so every floor read the counting loop while the rule lived in the other one. One pass now, counts returned alongside the hits, so each floor is evidence the rule actually ran over the tree. Control fixtures carry two ordinary lines ahead of the hit, so M24 (truncate to one iteration) fails them.

**Your fix did not cover the narrowest mutation, which is worth reporting.** Deleting only the line that *accumulates* the rule's output — rule runs, counts recorded, findings dropped — left every floor and both unit controls green. Over a clean tree, "found nothing" and "threw the findings away" are the same observation, and no floor or unit control can separate them. So `scan` is parameterised on its root and there is an end-to-end control: walk, read, rule, accumulate, over a synthetic tree built to hold exactly one stranded file. That mutation fails on it now.

**Nits:** the "roughly 3x" claim was false (1.6x–2.3x measured) and now states the figures rather than a rounding of them — you were right that it was the summarising sentence that drifted, and it is the third time in this PR. The pinned tree list now says it fails on an *added* tree too, which is the intended direction.

**Rebase:** onto #716 and #721. `docs/bugs/MINOR.md` conflicted and both sides are kept. #721's new `no-hand-rolled-comment-strippers` detector passes against this branch unchanged — the concealment gate parses `.gitattributes` paths and file basenames, not comments. The tree grew from 336 to 340 files, so every stated count in the stranding test was re-measured (1015 bare, 313 one-line, 1370 openers); margins are unchanged to a tenth.

ignaciohermosillacornejo:
Round-6 finding addressed in `0406986`.

**The `unreadable` collector had never been observed collecting.** Correct, and it is your round-4 argument about `stranded` applied to the list sitting next to it. I verified your mutation: replacing the wrapped read with the literal `catch { continue }` this PR exists to remove left **every** assertion in the file green — files, perTree, all three delimiter floors, `stranded`, both unit controls, the end-to-end stranding control, and `unreadable` itself. An empty list is what a clean tree produces, so it cannot distinguish a working collector from a deleted one.

There is now a positive control for it: a synthetic tree with a dangling symlink named `*.ts` — the exact case the `lstat` docblock claimed was handled, which lstat admits to the file list and which fails at the read. It asserts the entry names the file and carries `ENOENT`, and that the readable sibling in the same tree is still swept, so the failure is scoped to the file that caused it rather than to the tree. M25 (your mutation) fails it.

You were right that M19 had been covering this by hand since round 3, and right that this is the same relationship round 3 named about M15. Six rounds, and the recurring lesson has been the same one every time: the assertion that has only ever seen the negative case has not been tested.

Issue #454, opened 2026-06-10
Token extractor surfaces foreign-project tokens as PROJECT_NUMBER_MISMATCH instead of 'no Copilot session'


ignaciohermosillacornejo:
## Symptom

With no Copilot session in any browser, `bun run smoke` (and any authed GraphQL call) fails with:

```
Firebase token exchange failed (400): PROJECT_NUMBER_MISMATCH
```

This misattributes the failure: it reads like an API/key drift problem when the actual state is "user is logged out."

## Root cause

`src/core/auth/browser-token.ts` falls back from Copilot's IndexedDB to the browser-wide `Local Storage/leveldb` directory, which contains every site's storage. Any other Firebase-backed site leaves `AMf-` refresh tokens there. When no Copilot session exists, the extractor returns a foreign-project token (the `longest match` heuristic makes this worse), and the securetoken exchange correctly rejects it with `PROJECT_NUMBER_MISMATCH`.

Reproduced 2026-06-10: Copilot IndexedDB empty in all browsers; the only two `AMf-` tokens on the machine were in Chrome's global Local Storage, both rejected with `PROJECT_NUMBER_MISMATCH` against the Copilot web key.

## Expected behavior

- On `PROJECT_NUMBER_MISMATCH`, treat the candidate as foreign: discard it and continue searching remaining candidates/paths/browsers.
- If no candidate exchanges successfully, fail with the existing actionable message: "No Copilot Money session found... log into https://app.copilot.money".
- Error taxonomy (C2, #441): this is `AUTH_FAILED` / user-action-required, never a schema/API-drift attribution.

## Ratchet

- Unit test: extractor candidate list containing only a foreign-project token must produce the "no session" error, not a raw Firebase 400.
- Conformance ledger: the "refresh token found in browser storage belongs to copilot-production-22904" assumption is currently implicit and unverified — add an entry for it.

Found while running the B2 live-gate baseline (Epic C, #429 context).

PR #726, opened 2026-09-14
fix(auth): spend the exchange budget on Copilot's own store first

### LINKED ISSUE #722: Token-candidate ordering makes MAX_EXCHANGE_CANDIDATES unsafe: a real session can be crowded out by foreign tokens
Found during the #720 review (docs PR; deliberately not fixed there — it is a behaviour change, not a docs change).

`FirebaseAuth` tries `candidates.slice(0, MAX_EXCHANGE_CANDIDATES)` (`src/core/auth/firebase-auth.ts:70`, cap = 10 at `:24`). The candidate list comes from `BROWSER_CONFIGS` in path order, and `getChromiumProfileStoragePaths` emits **per profile** `[<profile>/IndexedDB/https_app.copilot.money_0…, <profile>/Local Storage/leveldb]` (`src/core/auth/browser-token.ts:78-81`), which `searchBrowser` flatMaps (`:246-254`).

So the "prefer the Copilot-scoped store" ordering is **within a profile**, not global. The resulting list interleaves scoped and browser-wide sources across profiles:

```
Chrome/Default/IndexedDB, Chrome/Default/LocalStorage, Chrome/Profile 1/IndexedDB, …
```

**The failure mode:** a user whose first profile's `Local Storage/leveldb` holds ten or more distinct `AMf-` tokens from other Firebase-backed sites exhausts the cap before the loop ever reaches the genuine Copilot token in a later profile's scoped IndexedDB. Every attempt fails `PROJECT_NUMBER_MISMATCH`, so they get `noCopilotSessionError` — which tells them to log in at `app.copilot.money`, when they already are. That is the same misleading-error shape as #454's `PROJECT_NUMBER_MISMATCH` class.

The cap itself is right (it bounds sequential exchanges for a logged-out user); the ordering is what makes it unsafe.

**Suggested fix:** sort candidates so Copilot-scoped paths precede browser-wide ones *before* the slice, so the cap can only ever discard low-probability candidates. That makes the bound safe rather than merely finite.

**Test gap:** `tests/core/auth/firebase-auth.test.ts:175` covers the cap, but not the interaction — no test has a valid token positioned past ten foreign ones. A regression test should place the genuine token after the cap boundary and assert it is still found.

Not user-facing until someone hits it, but when they do the error actively points them away from the cause.

ignaciohermosillacornejo:
# Summary

A user who **is** logged into Copilot Money could be told to log in. Candidates came back in disk-discovery order, which interleaves each profile's Copilot-scoped IndexedDB with the browser-wide `Local Storage/leveldb` every site writes to — so ten other sites' `AMf-` tokens in an early profile consumed the whole ten-candidate exchange budget before a later profile's real session was ever tried. Every attempt failed `PROJECT_NUMBER_MISMATCH`, and the user got `noCopilotSessionError`, which names the one action that cannot fix it.

Candidates now carry provenance (`TokenResult.scoped`) and every origin-scoped one is exchanged before any browser-wide one, so the cap can only ever discard low-probability candidates. **Nothing about which files are read from disk changes.**

Closes #722

## External assumptions

Three, all about **Google's securetoken endpoint** — not Copilot's surface, but external, and the cold-path auth loop branches on all of them. Two were already load-bearing and un-ledgered; the third is new here. All three are now entries in `src/conformance/ledger.ts` under `Securetoken.v1Token:*`, closing the gap the #478 post-mortem recorded ("the ledger has no category for client-side assumptions"). The ledger header and the `operation` naming convention were widened to admit a third-party `<Service>.<path>:<aspect>` surface; no ratchet applies to them (they carry no `toolParams` and no oracle, and the bijection tests walk the write-tool schemas).

| Surface | Assumption | Evidence class |
|---|---|---|
| `Securetoken.v1Token:foreignProject` | A token from another Firebase project is rejected with `PROJECT_NUMBER_MISMATCH` | **live observation** — seen repeatedly against production in #454/#478 with real foreign tokens; pre-existing, unchanged here, nothing re-checks it |
| `Securetoken.v1Token:invalidCandidate` | An `AMf-`-shaped string that is not a usable refresh token is rejected **400 `INVALID_REFRESH_TOKEN`**, *not* `PROJECT_NUMBER_MISMATCH` | **probe transcript** — one-shot probe run for this PR: a synthetic non-token posted to the documented endpoint with the public web API key returned `{"error":{"code":400,"message":"INVALID_REFRESH_TOKEN","status":"INVALID_ARGUMENT"}}`. No user data, no real token |
| `Securetoken.v1Token:endpointFailure` | The endpoint reports its *own* unavailability with a 5xx, a **429**, or no response — never by 4xx-ing a valid candidate | **`unverified`** — inferred from standard Google API HTTP semantics; no probe induced either. 429 is carved out explicitly because it is the known counterexample to the simpler "any 4xx is about the candidate" reading (`RESOURCE_EXHAUSTED`, `TOO_MANY_ATTEMPTS_TRY_LATER`). If it is wrong the failure is benign in the privacy direction: fewer requests and a raw error surfaced early |

One assumption is **removed**: that a non-`PROJECT_NUMBER_MISMATCH` failure implies the candidate was a Copilot-project token. The probe above refutes it — candidates are regex-scraped from raw LevelDB bytes, so a truncated match from any site's storage lands exactly there.

## Bug fix?

```text
Root cause:       `candidates.slice(0, 10)` applied to a list in disk-discovery order, which
                  interleaves Copilot's origin-scoped store with the browser-wide one per
                  profile — so the budget could be spent entirely on other sites' tokens.
Bug class:        ambiguous-candidate-selection (RECURRENCE — second instance after #478,
                  which shipped an instance-only ratchet)
Detector added:   tests/core/auth/candidate-ordering.test.ts — mutation-verified in four
                  independent directions. Encodes the class invariant, not the instance:
                  no single candidate may end the search for a valid one behind it,
                  whatever its position in the list or its failure mode.
Siblings checked: Found and fixed here — the loop aborted on any non-mismatch error, which
                  is the same starvation by another door (one garbage/truncated string from
                  any site's storage ends the search for a real session). Also swept: the
                  de-dup path (a token seen in both stores was pinned to the weaker
                  provenance) and the legacy single-pick `extractRefreshToken()` wrapper,
                  both fixed. `extractRefreshTokenCandidates` has 6 script call sites; all
                  construct FirebaseAuth the same way and inherit the fix.
Ledger updated:   src/conformance/ledger.ts — 3 new Securetoken.v1Token:* entries (see above).
```

Post-mortem: `docs/bugs/722-token-candidate-crowding-out.md`. The `ambiguous-candidate-selection` row in `docs/bugs/README.md` moves from **Detector: none** to the test above.

## Failing-then-passing evidence

The new tests were mutation-tested by reverting each half of the fix independently — so neither ordering site is dead code masked by the other.

| Mutation applied | Result |
|---|---|
| **both** ordering sites reverted | **3 fail** / 51 pass — and the failure message is the user-visible bug verbatim: `No Copilot Money session found… Please log into Copilot Money` |
| extractor ordering only reverted | **1 fail** — `the origin-scoped store is preferred across profiles, not just within one` |
| `FirebaseAuth` ordering only reverted | **1 fail** — `an unsorted extractor result does not defeat the budget` |
| 4xx-continuation reverted | **3 fail** — truncated-token / 403 / unexplained-rejection cases |
| de-dup provenance promotion reverted | **1 fail** — `a token seen in both stores is ranked by its strongest provenance` |
| origin match reverted to a bare substring | **2 fail** — lookalike prefix / subdomain rank as scoped |
| `scoped` dropped from a test literal | **build error** — the auth tests are now inside `tsconfig.tests.json`, so the required field is enforced |
| fix restored | **63 pass / 0 fail** in `tests/core/auth/`; `bun run check` exits 0 (3282 tests) |

The reproduction is deliberately not heavily mocked: it runs the **real** extractor over a real temp profile layout built by the production `getChromiumProfileStoragePaths`, and only `fetch` is faked — deciding accept-vs-reject from the token the request actually carries, which is the one part of the loop that cannot run locally. No token value is logged, asserted on, or put in a failure message; the fake records a provenance label (`session` / `foreign`), never the bytes.

## Three judgment calls

**1. Keep the cap at 10, and keep it a single global budget — not per-source.** Once scoped candidates hold the first slots, a per-source budget cannot improve their chances; its only effect would be to let *more* browser-wide tokens (other sites' tokens) reach Google's endpoint. It buys nothing for the bug and costs privacy, so the cap stays one shared budget of ten. The `MAX_EXCHANGE_CANDIDATES` docblock now says *why* the bound is safe rather than merely finite.

**2. An unscoped candidate is still tried when a scoped one exists.** Skipping them would regress the users the browser-wide read exists for: the `Local Storage` path is kept because older Firebase SDK versions and `browserLocalPersistence` store auth there, and a user can hold a stale scoped token alongside a live unscoped one. Skipping would be the same bug with the roles swapped. Ordering already captures the whole benefit — in the normal logged-in case the first exchange succeeds and **no** other site's token leaves the machine — without the risk.

**3. Users with no scoped candidate are not regressed.** *(This is the one the review sharpened — see the review round below.)* The partition is stable, so a list with zero scoped candidates comes out byte-identical to before and behaves exactly as it did. Two tests pin it: `all-foreign candidates still yield the actionable "no session" error`, and the pre-existing 25-candidate cap test (all browser-wide) still asserts exactly ten exchanges. The 4xx-continuation change strictly helps this group — a stale foreign token no longer ends their search.

## Review round

Four findings from the automated review, all accepted. Details in [this comment](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/pull/726#issuecomment-5672441245); in short: a browser-wide candidate's unexplained 4xx is no longer surfaced raw (it was #722's symptom from the other side), 429 left the per-candidate class, the origin match became host-delimited, and the auth tests joined `tsconfig.tests.json` — they were in **no** typecheck program, which is the actual reason a new required field on `TokenResult` broke nothing anywhere.

## Privacy surface

**Unchanged** — no file, directory, or traversal depth differs; only the order candidates are returned in. `PRIVACY.md` is updated anyway, because #720 documented two properties this PR makes more specific: the Copilot-scoped preference is now **global** rather than per-profile, and the ten-candidate budget is now described as a worst-case ceiling (a logged-in user normally sends zero foreign tokens to Google) with an explicit note that it is one shared budget, deliberately not one per source. `bun run check:privacy-endpoints` passes.


ignaciohermosillacornejo:
Thanks — all four accepted, with one correction to the diagnosis on #3 that turned out to be the more useful finding.

**1. `firstUnexplainedRejection` ignores provenance — accepted, implemented as suggested.** You're right that this is #722's symptom reached from the other side, and the predicate is exactly the one the rest of the PR argues for. `candidate.scoped &&` now gates the promotion. The test you predicted would need to flip did: `an unexplained rejection is surfaced…` became **`a SCOPED candidate's unexplained rejection is surfaced rather than flattened`** (still passing — that's the expired-Copilot-token case), and its mirror image is new: **`the same rejection from browser-wide candidates only still says "log in"`**, which is the logged-out user you described. Both assert the message, not just the throw.

**2. 429 — accepted.** `TOO_MANY_REQUESTS` is excluded from `isCandidateRejection` with a named constant explaining that it's a 4xx by number and a statement about the endpoint by meaning. New test mirrors the 503 one: exactly one request, error surfaced raw. I took your "narrowing the code is cheaper than defending the assumption" **and** amended the `endpointFailure` ledger evidence, since the entry's wording was the thing that made 429 a counterexample — it now names 429 explicitly as the case that was considered.

**3. `uid-transition-sweep.test.ts` — accepted, but the mechanism is worse than `as TokenResult[]`.** I first "fixed" this with `satisfies TokenResult[]`, then mutation-tested it: **deleting the field still compiled.** The reason is that `tests/core/auth/*` and `tests/integration/uid-transition-sweep.test.ts` are in **no typecheck program at all** — `tsconfig.tests.json` has an explicit include list and none of them were on it, so neither `as` nor `satisfies` was ever evaluated by anything. A typed contract that no program typechecks is not a contract.

So the real fix is the include list. Those four files are in it now, which surfaced 11 errors worth having: three unsound `as typeof fetch` assertions on bun `Mock`s (they genuinely lack `preconnect`), two dead imports in `tests/helpers/test-db.ts`, and — the interesting one — five `Type 'false' is not assignable to type 'true'` errors, because `mock()` inferred its return type from the default extractor's literal `scoped: true`, so every `scoped: false` override was a type error waiting to happen. Both extractor mocks are now typed by return annotation (`(): Promise<TokenCandidates> =>`) rather than assertion. Mutation-verified: dropping `scoped` from a literal is now a build error.

**4. Bare substring match — accepted, with a change to the proposed fix.** Anchoring on the two known directory forms doesn't close it: a Firefox lookalike at `app.copilot.money.example.com` produces `https+++app.copilot.money.example.com`, which *contains* `https+++app.copilot.money`. And enumerating encodings rots the day a browser adds a suffix — Firefox's `^partitionKey=…` already does. So it's a host-delimited match instead: `/(?:^|[^A-Za-z0-9.-])app\.copilot\.money(?![A-Za-z0-9.-])/`, which relies only on the property every encoding shares (the host is wrapped in characters that can't appear in a hostname). Seven cases pinned, including both lookalike shapes and the partitioned-origin form; mutation-verified against the old substring test.

**5. Duplicated partition — accepted.** `orderByProvenance()` is exported from `browser-token.ts` and called from both sites. The independence you and I both wanted is intact: `selectExchangeCandidates` still applies it itself and still owns the cap, so the bound doesn't rest on the extractor having called anything — only on the two agreeing about what "first" means, which is now enforced by them being the same function.

**6. `MAX_EXCHANGE_CANDIDATES` mirrored — changed anyway.** You called it fine as-is, but the drift risk isn't symmetric: *lowering* the cap keeps the test valid, *raising* it above 12 would make the fixture fit inside the budget and the test would pass without any ordering at all. It's exported and imported now, so the fixture scales with it.

**Docblock rewrap** was already fixed in `01409e9`, which landed after the review job started.

Full `bun run check` exits 0 (3282 tests).

ignaciohermosillacornejo:
Both accepted, and both turned out to be probeable rather than arguable — so I probed them instead of reasoning about Google's semantics.

**1. Key-level failures arrive as 4xx — confirmed, and the regression is real.** Two one-shot probes against the documented endpoint, synthetic input only:

| Request | Result |
|---|---|
| invalid API key + synthetic token | **`400`**, `"reason": "API_KEY_INVALID"` |
| no key at all + synthetic token | **`403`**, `PERMISSION_DENIED` |

So a dead key returns the *same status* as a bad refresh token, and status cannot be the discriminator. Your walkthrough holds exactly as written: with `candidate.scoped &&` in place, a user with no scoped candidate would have burned ten exchanges and then been told to log in. That is this PR's own symptom, and I'd rather have found it here than in the release.

`isCandidateRejection` now classifies in three steps: non-4xx → endpoint; `403`/`429` → endpoint; 4xx whose body names an `ENDPOINT_LEVEL_ERROR_CODES` entry (`API_KEY_INVALID`, the three `API_KEY_*_BLOCKED` forms, `SERVICE_DISABLED`) → endpoint; everything else → candidate. Note `403` moved wholesale rather than by code list: securetoken's per-token verdicts (`INVALID_REFRESH_TOKEN`, `TOKEN_EXPIRED`, `USER_DISABLED`, `PROJECT_NUMBER_MISMATCH`) are all 400s, so nothing about a candidate arrives as a 403 — which also means **the `403 PERMISSION_DENIED` row I had pinned as *continue* was wrong**, and it is now pinned as *stop*. Two 400 rows replaced it (`TOKEN_EXPIRED`, `USER_DISABLED`) so the continue-class still has real coverage.

**I did not take the uniformity alternative, and I think it's actively unsafe here.** "Every candidate failed with the byte-identical non-mismatch message" is exactly what a genuinely logged-out user looks like: their candidates are all truncated fragments, all rejected `INVALID_REFRESH_TOKEN`, all with identical bodies. That heuristic would hand them a raw 400 — reinstating finding #1 from your previous round, on the population this change exists to protect. It also arrives too late to help: the uniformity is only observable *after* all ten requests, so it fixes the message and not the ten tokens sent. The code list stops on the first one. Its cost is depending on Google's vocabulary, and the failure mode of an incomplete list is the old behaviour, not something worse — recorded as such in the ledger, which is now `PARTLY probed` and states plainly which halves remain unverified.

**2. Trusting the collaborator to de-duplicate — accepted, and you're right that it's the cheaper half of the same invariant.** `dedupeByToken()` joins `orderByProvenance()` as an exported helper, and `selectExchangeCandidates` is now `orderByProvenance(dedupeByToken(candidates)).slice(0, MAX)`. The extractor calls both too, which let its inline `seen` Map go — so there is one definition of each half rather than two that must agree, while both sites still apply them independently. One behavioural nicety: the shared helper keeps the first sighting's position *and browser*, lifting only `scoped`, and returns new objects instead of mutating — the old inline version mutated a candidate that callers also held.

New test in the FirebaseAuth-only block, where you pointed out the gap was: `duplicate sightings of one token cannot spend the budget more than once` — fifteen copies of one token ahead of the session, asserting `['foreign', 'session']`.

**On the three docs-only commits:** correct reading, and the mutation-count one was the same rot the ledger header warns about — it said "four directions" while the table had grown to five.

Rebased onto `origin/main` (#716, #721) — one conflict, in the ledger header comment #721 rewrote; kept main's more precise wording and re-appended the `<Service>.<path>:<aspect>` exception. **Re-ran the mutation after the rebase, not just before it:** removing the ordering still fails three tests with `No Copilot Money session found`. Added the `### Fixed` CHANGELOG entry. `bun run check` exits 0 (3348 tests).

ignaciohermosillacornejo:
The last review round on this PR (the `TOKEN_EXPIRED` / `INVALID_REFRESH_TOKEN` finding plus three nits) landed after the merge, so it is in #736 rather than here. Nothing in that round changes what merged — it narrows the raw-error branch that #726 had already narrowed twice, closing the last route to the same unactionable message.

Issue #663, opened 2026-08-21
get_holdings and get_recurring_transactions still report the institution label, not the nickname


ignaciohermosillacornejo:
## Summary

#660 (merged in 71579ca) made cache-mode `get_accounts` prefer the user's `nickname` over the institution-supplied `name`. Two other cache-mode surfaces that report an account name were not updated and still return the institution label:

- `resolveAccountName()` — `src/tools/tools.ts:637`, returns `account?.name`; feeds `get_recurring_transactions`' `account_name`
- `getHoldings()` — `src/tools/tools.ts:2553`, `account_name: acct.name ?? acct.official_name`

A third surface is already correct and is the reference implementation: `CopilotDatabase.getAccountNameMap()` (`src/core/database.ts:1357`) does `nickname ?? name`, feeding `get_balance_history`.

## Impact

The same `account_id` can be reported under two different names by two tools in one session, which breaks correlation for any client joining holdings or recurrings against the account list.

**Neither reproduces on the cache I checked** — 0 of 59 recurring rows carry an `account_id` at all, and the accounts holding securities have no nickname that differs from `name`. So this is latent, not currently user-visible. Filing it because the repo's ritual is to fix the class, not the instance, and because the "0 of 59 recurring rows have an `account_id`" observation is itself worth a look — `resolveAccountName` may be effectively dead code on real data.

## Notes

Prefer routing both through one helper rather than inlining a third and fourth copy of the precedence rule — but see the coupling warning in the companion issue on consolidating the display-name definitions, since the obvious shared helper is also what keys the decoder's dedup.

Found while reviewing #660.

ignaciohermosillacornejo:
Adding a sub-finding to this class, surfaced while reviewing #728 (which deletes the dead `getAccountDisplayName` / `withAccountDisplayName` / `AccountWithDisplayName`).

This issue calls `getAccountNameMap()` the reference implementation. It is nickname-aware, but it does **not** agree with the `get_accounts` handler on an empty-string nickname:

| site | code | `nickname: ''` |
|---|---|---|
| `getAccounts()` — `src/tools/tools.ts:1400` | `account.nickname ? { ...account, name: account.nickname } : account` | truthy check fails → falls back to `name` |
[...]

Issue #662, opened 2026-08-21
decodeAccounts dedups on provider name, ignoring nickname — silently drops account documents


ignaciohermosillacornejo:
## Summary

`decodeAccounts()` deduplicates accounts with the key `` `${getAccountDisplayName(acc)}|${acc.mask ?? ''}` `` — the institution-supplied `name` (falling back to `official_name`), **never the user's `nickname`**. Any two account documents sharing a provider name and a mask collapse into one, and the loser is discarded before anything downstream can see it.

Two copies of the same key:
- `src/core/decoder.ts:325` (`decodeAccounts`)
- `src/core/decoder.ts:2977` (`decodeAllCollections`)

## Evidence (real cache, 2026-08-20)

- Cache mode returns **18** accounts; `get_accounts_live` returns **20** for the same user.
- The decoder emits an `unread field: collection=accounts docId=…` warning for at least one document that is **absent from its own return value** — i.e. the document is present and processed, then dropped by the dedup, not missing from the cache.
- The collapsing group is three same-named investment accounts with `mask: null`, so the key degenerates to `name|`.

The dropped accounts currently carry no balance, so net worth is unaffected **today**. That is luck, not design: the key drops any same-name/same-mask pair regardless of balance.

## Why this matters

This is the #72 / #82 class — *account silently missing from `get_accounts` and net worth*. There is no error, no warning, and no count mismatch surfaced to the caller; cache mode and live mode simply disagree about how many accounts exist.

It also partially defeats #660 (merged in 71579ca): a user who renames two identically-named accounts precisely so they can be told apart gets no benefit, because the dedup runs on the provider name **before** `nickname` is ever consulted.

## Fix sketch

Key the dedup on `account_id` (the Firestore documentId), which is unique by construction — the (name, mask) heuristic predates having stable ids in the decoded model. If the heuristic exists to merge genuine duplicate documents for the same account, that intent needs to be stated and tested; right now it is indistinguishable from data loss.

Detector for the class: assert decoded account count equals the raw `accounts` document count minus documents that are tombstones/parent-pointers, so a dedup that eats real rows fails loudly. Compare against `get_accounts_live` in the live smoke as a second signal.

Found while reviewing #660. Refs #72, #82, #624.

PR #728, opened 2026-09-14
refactor(models): delete the nickname-unaware account display-name helpers (#664)

### LINKED ISSUE #664: Four competing definitions of "account display name" — consolidate, but only after the dedup key is decoupled
## Summary

"The name of an account" is now resolved four different ways:

| site | rule |
|---|---|
| `getAccountDisplayName()` — `src/models/account.ts:115` | `name ?? official_name ?? 'Unknown'` |
| `CopilotDatabase.getAccountNameMap()` — `src/core/database.ts:1357` | `nickname ?? name` |
| `decodeAccounts()` dedup key — `src/core/decoder.ts:325`, `:2977` | `getAccountDisplayName(acc)` |
| `getAccounts()` — `src/tools/tools.ts` (added by #660) | inline `account.nickname ? … : account` |

Any future change to the precedence rule has to be applied in all four or the surfaces drift apart again — which is exactly how the sibling-drift issue filed alongside this one came about.

## The trap

The obvious move — put nickname precedence into `getAccountDisplayName()` and have everyone call it — **is not a safe refactor**. That function is also what keys the decoder's dedup, so changing it changes which account documents survive decode and therefore how many accounts `get_accounts` returns. That is a deliberate, test-covered behavior change, not a cleanup.

Sequencing suggestion: resolve the dedup-key issue first (key on `account_id`, decoupling dedup from display naming), *then* consolidate the display-name rule into a single helper, at which point it is a genuine no-op refactor.

Also unused: `withDisplayName()` / `AccountWithDisplayName` (`src/models/account.ts:122-129`) — exported, re-exported, and unit-tested, with zero consumers in `src/`. It implements the "add a derived `display_name`, keep `name`" shape. Either adopt it or retire it.

Note the deliberate decision in #660 not to use it: `get_accounts` overwrites `name` rather than adding `display_name`, because that matches what live mode does (GraphQL `Accounts.name` returns the user label and carries no separate provider field). Worth keeping that parity rationale if this is revisited.

Found while reviewing #660.

ignaciohermosillacornejo:
# Summary

Deletes three dead symbols from `src/models/account.ts` — `getAccountDisplayName`, the account `withDisplayName` (barrel-exported as `withAccountDisplayName`), and the `AccountWithDisplayName` interface — plus their barrel re-exports and their unit tests. No deprecation shim, no tombstone comment: git history is the record.

Closes #664

> **Rebased onto #716.** This PR was opened before #716 landed and has been rebased onto it, resolved deletion-wins. See "What #716 changed" below — it makes the case for deleting *stronger*, not weaker.

## Why deletion and not a warning

These implement `name ?? official_name ?? 'Unknown'` — **nickname-unaware**. #716 established `preferredAccountName` as the canonical rule and routed all four live sites through it:

| site | now |
|---|---|
| `getAccounts()` — `src/tools/tools.ts:1405` | `preferredAccountName(account)` |
| `getHoldings()` — `src/tools/tools.ts:2734` | `preferredAccountName(acct)` |
| `getAccountNameMap()` — `src/core/database.ts:1386` | `preferredAccountName(account)` |
| `resolveAccountName()` — `src/tools/tools.ts:785` | `preferredAccountName(account)` |

So after #716 these three symbols are **the only account-name definition left in the tree that disagrees with the canonical one** — and nothing calls them. That divergence has already been the mechanism behind two bugs: #663 ("the name rule was at 2 of 4 sites") and #662, where `getAccountDisplayName` was built into the account dedup key and silently dropped documents.

## What #716 changed, and why I'm not keeping its remedy

#716 chose to **defend** these symbols with JSDoc rather than delete them, on the grounds that the models test pinned their behaviour. Its own warning text makes the opposite case better than I can:

> *"two exported name helpers in one module, one of them literally called 'display name', is exactly how a future surface picks the wrong one. It has happened."*

A JSDoc warning is read only by someone who has already opened the file and scrolled to the symbol. The failure mode being defended against is the *opposite* of that: someone typing `displayName` into autocomplete while building a user-facing row, and taking the first hit. Autocomplete does not render JSDoc warnings as a reason to stop. And the test that "pinned the behaviour" pinned behaviour that nothing depends on — it was the only consumer keeping the symbols alive.

Deleting the symbols removes the choice rather than annotating it, and takes the warnings with it, since they exist only to describe this code. Per the maintainer: *"retire dead code please. or just delete it, no need in documenting it and retiring, that's what git is for."*

#664's stated blocker has also expired — it said to act "only after the dedup key is decoupled", and #668 rekeyed `deduplicateAccounts` on `account_id`.

## Conflict resolution (deletion-wins)

Everything #716 added is preserved; only what described the deleted symbols went with them.

| kept | dropped |
|---|---|
| `preferredAccountName`, **including its trim-for-blank detection** (`nickname?.trim() ? nickname : undefined`, returning the untrimmed value) | the two JSDoc blocks warning about the deleted helpers |
| `isVisibleAccount`, `isVisibleAccountNode` and the `ACCOUNT VISIBILITY` block comment | the two trailing `// not the user-facing label… (#663)` comments on the barrel export lines |
| the `describe('preferredAccountName', …)` test block | the `getAccountDisplayName` and `withDisplayName (account)` describe blocks |

Three edits beyond pure deletion, all in the surviving `preferredAccountName` test block, all removing references to the code this PR deletes:

- a comment read *"The documented difference from getAccountDisplayName below…"* — now states its intent directly without naming the deleted symbol;
- the block's header comment cited *"Its deprecated neighbour below had the behaviour table"* — that neighbour is gone, so the comment now keeps only what still holds (these tests pin the canonical helper's input classes, and blank-vs-absent is the axis worth covering);
- `type Account` became an unused import once the deleted blocks went, so it is dropped.

Worth noting that **no gate catches that last one**: eslint ignores `tests/**`, `tsconfig.json` excludes `tests`, and this file is not in `tsconfig.tests.json`'s include list. `bun run check` is exit 0 either way. Fixing it by hand is the point — a dead symbol left behind by a PR about dead symbols being traps.

## Evidence that this is safe

**Zero consumers.** Grepped all three names repo-wide before and after:

- `getAccountDisplayName` — one remaining hit, `docs/bugs/662-account-dedup-drops-documents.md:43`, correctly left as historical narrative.
- `AccountWithDisplayName`, `withAccountDisplayName` — zero hits anywhere.
- `scripts/` and `skills/` — zero.

**Not a published API.** `package.json` ships `files: ["dist", "README.md", "LICENSE", "PRIVACY.md"]`, `main` is `dist/server.js`, `bin` is `dist/cli.js`, and there is **no `exports` map** and no `types` field exposing `src/models`. Nothing downstream can import these symbols, so removing them is not a breaking API change. This is the load-bearing fact for deleting rather than deprecating. Confirmed from the other direction too: `bun run build` succeeds and `grep` finds zero traces of the deleted names in the built `dist/` — they were never reachable in the shipped bundle either.

## The transaction twin is untouched

`src/models/transaction.ts` has its own `withDisplayName` (re-exported as `withTransactionDisplayName`) alongside `getTransactionDisplayName` and `TransactionWithDisplayName`. Different function, different entity — entirely out of scope. Verified two ways: SHA-256 of the file is byte-identical to `origin/main`, and `git diff origin/main..HEAD -- src/models/transaction.ts` is empty. Its tests still pass. The `withDisplayName()` reference in `docs/firestore-collections.md` is in the **transactions** section and remains accurate.

## External assumptions

None. This PR touches no Copilot API surface, no GraphQL operation, no response shape, no enum, and no cached-data assumption. It only deletes local TypeScript that nothing calls.

## Verification

Re-run after the rebase onto #716:

- `bun run check` — exit 0 (3316 tests, 20 skip, 0 fail)
- `bun run check:skills` — exit 0 (6 skills validated)
- `bun run build` — exit 0, zero traces of the deleted names in `dist/`
- Even with `origin/main` (0 commits behind at push time)
- Diff: 3 files, +6 / -150


ignaciohermosillacornejo:
Thanks — all four notes land. Taking them in order.

**1. `preferredAccountName` does not exist — agreed, and already corrected.**

You're right, and the correction was in flight before your review posted: I grepped for it while the review job was running, found zero hits, and rewrote the description. The claim came from my task brief, which asserted that helper as the canonical one; I repeated it without checking, which is exactly the failure mode this repo's "claims must name a runnable check" rule exists to catch. The body now carries the real table of inline sites instead.

**2. `resolveAccountName` — confirmed, and it's already filed as #663.**

I verified it independently rather than taking it on faith: `db.getAccounts()` (`src/core/database.ts:1152-1173`) only does type filtering — no nickname mapping — so `resolveAccountName` returning `account?.name` really does hand `get_recurring_transactions.account_name` the provider label, while `get_accounts` shows the nickname.

This is **#663**, still open, which names `resolveAccountName()` and `getHoldings()` as exactly these two sites. So the class is tracked, not lost. Your framing is fair: deleting the dead helper removed the autocomplete trap, not the class — and I've rewritten the PR description to say precisely that rather than implying a consolidation happened.

**3. Empty-string divergence — confirmed, and it was genuinely new.**

Verified: `??` keeps `''`, then the `if (displayName)` guard on the next line drops the account from the map entirely, so `get_balance_history`'s `account_name` is `undefined` for it — while the truthy check in `tools.ts` falls back to `name`. That's a third outcome from one nominal rule, and it is *not* in #663, which calls `getAccountNameMap()` the reference implementation. Added it there as a comment (#663) with the cross-surface assertion I'd want pinned when the shared helper lands.

**4. Class-level ratchet — declining, and I'd rather say why than add a weak one.**

Two reasons:

- The ritual is scoped to bug-fix PRs. This is a tech-debt deletion; nothing was broken, so there is no root cause to detect. You flagged this yourself as arguably N/A and I agree.
- More importantly, the obvious ratchet is fakeable. A test asserting `src/models/account.ts` exports no `*DisplayName` symbol constrains the *name*, not the defect — reintroduce the same nickname-unaware rule as `accountLabel()` and the guard stays green while the bug is fully back. Per this repo's "guard remedies must not be fakeable" principle, the cheapest way to turn that guard green is to rename, which means it isn't really guarding.

The honest ratchet for this class is the one in #663: a single exported helper, plus a test asserting `get_accounts`, `get_recurring_transactions.account_name`, and `get_balance_history.account_name` agree on the same `account_id` across nicknamed / un-nicknamed / empty-nickname. That fails on any site drifting, which is the actual class. It needs a behavior change to live code, so it belongs in #663's PR and not in a deletion-only diff.

**On verification:** `bun run check` exit 0 (3235 pass / 20 skip / 0 fail), `bun run check:skills` exit 0, `bun run build` exit 0, and `grep` finds zero traces of the deleted names in the built `dist/` — they were never reachable in the shipped bundle either. CI agrees: Tests, Quality Checks, PR sections, and codecov/patch all green.

No code changes from this round — notes 2 and 3 are pre-existing live behavior and out of scope for a deletion-only diff; they're recorded in #663 where the fix has to happen.

ignaciohermosillacornejo:
**Rebased onto #716** (which merged while this was open), resolved deletion-wins. Force-pushed, so the earlier approval was dismissed. Description rewritten for the post-#716 tree.

Two things worth flagging for whoever reviews the new head:

**1. #716 fixed the two live findings from the previous review round.** That review flagged `resolveAccountName` as a third nickname-unaware site, and an empty-string divergence between `getAccounts` (truthy) and `getAccountNameMap` (`??`). #716 closed both by introducing `preferredAccountName` and routing all four live sites through it — `getAccounts` (`tools.ts:1405`), `getHoldings` (`:2734`), `resolveAccountName` (`:785`), `getAccountNameMap` (`database.ts:1386`). I'd added the empty-string case to #663 as a comment before #716 landed; it is now fixed, and `preferredAccountName`'s doc comment records the reasoning.

That also retires the correction I posted earlier in this thread: `preferredAccountName` **does** exist now. It did not when I first wrote the description, which is why that paragraph read the way it did.

**2. I did not keep #716's remedy for these three symbols.** #716 added JSDoc warnings and kept them, reasoning that the models test pinned their behaviour. This PR deletes them instead, per the maintainer's instruction ("retire dead code please. or just delete it, no need in documenting it and retiring, that's what git is for"). The warnings go with the code they describe, since they exist only because it does.

I'd argue the deletion is the stronger reading of #716's own warning text — it says two exported name helpers in one module, one "literally called 'display name'", is *"exactly how a future surface picks the wrong one."* The reader that warning protects is one who already opened the file; the reader who causes the bug is one taking the first autocomplete hit. Deleting removes the choice rather than annotating it, and the test that pinned the behaviour was the only thing keeping the symbols alive.

**Preserved from #716, verified explicitly:** `preferredAccountName` including its trim-for-blank detection, `isVisibleAccount`, `isVisibleAccountNode`, the `ACCOUNT VISIBILITY` block comment, and the `describe('preferredAccountName', …)` test block. `git diff origin/main..HEAD -- src/models/transaction.ts` is empty.

The one non-deletion edit is a comment inside a surviving `preferredAccountName` test that pointed at "getAccountDisplayName below" and would have dangled; it now states its intent without naming the deleted symbol.

`bun run check` exit 0 (3316 tests), `bun run check:skills` exit 0, `bun run build` exit 0, branch even with `origin/main`.

ignaciohermosillacornejo:
Both notes fixed, amended into the single commit. Head is now `966d92e`.

**Note 1 — unused `type Account` import.** Confirmed and dropped. Your analysis of why no gate catches it is right and I verified it independently: eslint ignores `tests/**`, `tsconfig.json` excludes `tests`, and this file isn't in `tsconfig.tests.json`'s include list. `bun run check` was exit 0 with the dead import sitting there, which is precisely the reason to fix it by hand — a PR arguing that dead symbols are traps should not ship one, and the one automation that would have told me doesn't run here.

**Note 2 — dangling "deprecated neighbour below".** Confirmed and rewritten. Good catch that I got the sibling at line 56 and missed this one. It now keeps only what still stands:

> The canonical helper. It sat in six files, none of them a test — covered only through the four-surface parity tests, which exercised two of its input classes. These tests pin the rest directly. Blank-vs-absent is the axis worth the coverage: that is where the rule's trimmed truthiness parts company with a nullish check, and where #663 went in reverse.

**On the stale description:** that was a race rather than a disagreement — I rewrote the body for the post-#716 tree at roughly the moment your job started, so you read the pre-rebase version. The body you quoted three claims from is already gone. It now carries the four-site `preferredAccountName` table, the deletion-wins resolution table, and the rebase banner. I've corrected the diffstat again (`+6 / -150` after these two fixes — thanks for catching that it was stale even in the version I'd just written).

**On the ratchet (your note 3):** agreed, and your version is better than mine. I argued the honest guard couldn't be a name-check; you pointed out it already exists as behaviour — a reintroduced `accountLabel()` has to displace one of the four call sites to do harm, and the parity tests fail when it does. That's the non-fakeable form, and it's already in the tree. Nothing to add.

Gates after the fixes: `bun run check` exit 0 (3316 tests, 0 fail), `bun run check:skills` exit 0, `bun run build` exit 0, `bun test tests/models/models.test.ts` 13 pass (6 `preferredAccountName` + 7 transaction). Branch even with `origin/main`.

PR #736, opened 2026-09-15
fix(auth): a dead Copilot token is the logged-out state, not a mystery

ignaciohermosillacornejo:
# Summary

Follow-up to [#726](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/pull/726), which merged while its last review round was still being addressed. One finding from that round, plus three nits it raised.

After #726's `403`/`429` and API-key carve-outs, the codes that could still reach the "surface this raw" branch were mostly `INVALID_REFRESH_TOKEN` and `TOKEN_EXPIRED` — a **dead token**, whose remedy is exactly the action `noCopilotSessionError` names. So a user who logged out while Copilot's own IndexedDB still held the now-dead token got a raw Firebase 400 instead of "please log into Copilot Money": [#722](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/issues/722)'s symptom, from the last side it had left. The branch justified itself as *"telling the user to log in would contradict evidence we hold"* — but for those two codes it doesn't contradict it, it **is** it.

Those two codes now count as explained by being logged out and fall through to the actionable message. `USER_DISABLED` stays raw, because logging in cannot revive a disabled account — and so does any code Google adds that we have never seen, since an unrecognised rejection on a token from Copilot's own store is precisely where a raw error beats a guess.

Not a regression from #726: the branch predates it (`tests/core/auth/firebase-auth.test.ts` has pinned it since [#478](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/pull/478)), and #726 only narrowed who reaches it. The reachable path today is the mid-session re-extraction one, where `runServer`'s startup hint isn't there to compensate.

## External assumptions

**One, and it is a narrowing of an existing entry rather than a new surface.** `Securetoken.v1Token:invalidCandidate` already records (probe transcript, #726) that an unusable `AMf-` string is rejected `400 INVALID_REFRESH_TOKEN`. This PR adds the reading that `INVALID_REFRESH_TOKEN` and `TOKEN_EXPIRED` mean *the token is dead* while `USER_DISABLED` means *the account is disabled* — Firebase's documented vocabulary, **`unverified`** as a claim about which codes securetoken can emit for a live-project token. No new ledger entry: the existing `endpointFailure` entry already states that the body-code lists are not known to be exhaustive, and this list has the same failure mode — an unlisted dead-token code degrades to the pre-PR behaviour (a raw 400), never to a wrong answer.

## Bug fix?

```text
Root cause:       The raw-error branch was defined by exclusion ("not a foreign project"),
                  so after #726 carved out the endpoint-level codes, what remained was
                  mostly the one state the actionable message already describes.
Bug class:        ambiguous-candidate-selection (same class as #722; this is the last
                  route to its symptom, not a new mechanism)
Detector added:   Extends tests/core/auth/candidate-ordering.test.ts — the class invariant
                  it encodes is unchanged; two cases now assert the message a scoped dead
                  token produces, in both directions. Mutation-verified.
Siblings checked: The #478 test that pinned the old behaviour is reconciled rather than
                  deleted — it moves to USER_DISABLED and says why, so the property it
                  was protecting still has a test. Swept the other callers of
                  isForeignProjectError: none; the only consumer is this loop.
Ledger updated:   No new entry — narrowing sits inside Securetoken.v1Token:invalidCandidate
                  and endpointFailure, both of which already state the lists are not known
                  to be exhaustive.
```

## Mutation evidence

| Mutation | Result |
|---|---|
| `isExplainedByLoggedOut` → `isForeignProjectError` (i.e. revert this fix) | **3 fail** — the two `%s still says "log in"` cases and `a dead Copilot token yields the actionable message` |
| restored | **71 pass / 0 fail** in `tests/core/auth/`; `bun run check` exits 0 (3344 tests) |

## Also, from the same review round

- **The API-key test pins the probe's verbatim body.** It was hand-shaped (`{message, status}`); Google actually nests the reason in `error.details[].reason` with `error.status: "INVALID_ARGUMENT"`. The substring check matched either way, but the ledger's claim rests on a probe transcript — so the test should be evidence that production still matches, not that a plausible body does.
- **`MAX_EXCHANGE_CANDIDATES`' docblock stops over-claiming.** "Everything the cap drops came from a store every site writes to" is true only while fewer than ten *distinct* scoped candidates exist. Now stated with that condition, and why de-duplication makes it the normal case.
- **`isCopilotScopedPath`'s docblock says what a whole-path match does and doesn't support.** It claimed "only app.copilot.money could have written the tokens under it"; Chromium and Safari match on the full path, so a directory component literally named `app.copilot.money` would also match. Harmless — it fails toward pre-#722 behaviour and reads nothing extra — but the docblock shouldn't assert more than the test does.
- The post-mortem's mutation-direction count is **dropped** rather than re-numbered. It had gone stale twice.

Refs #722 (already closed by #726 — no closing keyword, this is the tail of that work, not a reopening).


ignaciohermosillacornejo:
Four of five taken. Finding #2 was the valuable one and I confirmed it by running the mutant before fixing it.

**2. The denylist mutant — confirmed, and worse than the note suggested.** You flagged it as unverified; it survives the **entire suite**, not just the auth tests: `3324 pass, 0 fail` with `isExplainedByLoggedOut` replaced by `!message.includes('USER_DISABLED')`. That is the repo's own `vacuous-assertion` class — a guard executed by every test and detected by none — and it inverted the allowlist's *second* claim, which is the one the docblock says is the reason for an allowlist at all.

Fixed with the case you named (`SOME_CODE_GOOGLE_HAS_NOT_SHIPPED_YET` → raw). Re-ran the mutant: now `71 pass, 1 fail`, and it is that test. Row added to the post-mortem's mutation table.

**1. Fast path — taken, not documented away.** You're right that the docblock's *"any failure here is a genuine error, not a foreign one"* is the same reasoning this PR refuted: a server-issued token is known-good only until the user logs out. It now falls through to a cold re-extract on `isExplainedByLoggedOut`, so a dead cached token produces either a live session from another profile or the actionable message — and any other failure still propagates untouched. The docblock says why the fall-through cannot loop (`exchangeToken` has already cleared `refreshToken`).

Two cases, both directions, mutation-verified (rethrowing instead of falling through fails both): `a dead CACHED token falls through to a cold re-extract when nothing else works` / `…when another profile has a live session`. The second is the one worth having — it proves the fall-through can actually recover a session, not merely relabel the error.

**3 and 4 — taken.** The ledger sentence now says `isForeignProjectError` is *one of two* reasons and names `isExplainedByLoggedOut` as the branch condition. The post-mortem's prose gained the paragraph its table already implied: that *unexplained* had to be narrowed twice more, each round of carve-outs changing what was left in it, and that the same reasoning then reached the fast path. Line rewrapped.

**5 — left alone, agreeing with your reasoning and adding one point.** Substring-matching the whole error message can match a code mentioned anywhere in a future body, and a misfire falls toward the actionable message rather than away from it. The point that decides it for me is that parsing `error.status` / `error.details[].reason` would be a *narrowing*: the three predicates would then only fire on codes appearing in the field we chose to read, and a body shape we did not anticipate would stop matching entirely — trading a rare over-match for a silent under-match on the surface where under-matching produces the unactionable message. Worth doing when a fourth predicate arrives and the verbatim probe fixture can anchor the parse; not worth it for three.

Rebased onto `origin/main` (it moved twice during this round). `bun run check` exits 0 — 3365 tests.

Issue #610, opened 2026-08-03
Landing page advertises the total tool-definition count no mode actually exposes


ignaciohermosillacornejo:
Carried over from the #588 review (non-blocking there): `docs/index.html` (meta description + features subtitle) advertises the total number of tool *definitions* (currently 50), but no single mode exposes that many — cache mode serves 14, `--write` + `--live-reads` (the largest real surface) serves 44 (19 write + 25 live-mode reads). Since #588 merged, `scripts/check-tool-counts.ts` (~lines 120-126) actively enforces the current framing, so fixing the copy means updating both the HTML and the checker's expected strings together.

**Work:** reframe the landing-page copy to per-mode counts (e.g. "14 read tools locally, 44 in full write mode" or similar), update the corresponding assertion block in `scripts/check-tool-counts.ts`, and verify with `bun run check:tool-counts`. Small, docs + one script, no GraphQL surface. The checker itself tells you every place that must stay in sync.

PR #746, opened 2026-09-16
fix(gates): close two unguarded classes — prose tool counts, and scripts git silently drops

### LINKED ISSUE #723: Close the unguarded-tool-count class: EXAMPLE_QUERIES.md (also incomplete) and CONTRIBUTING.md
Found by the #720 review. #610 closed the unguarded-count class for `docs/index.html` — all three counts on that page are now derived from the registry and asserted by `scripts/check-tool-counts.ts`. Two prose surfaces still restate tool counts with no needle behind them, which is the same class #610 was filed for.

Kept out of #720 deliberately: that PR's scope was the landing page (#610), PRIVACY.md (#615), and DESIGN_NOTES (#712). These are different files, and one of them needs a content fix rather than a count fix.

### 1. `docs/EXAMPLE_QUERIES.md:541` — the count is guarded by nothing *and* the table behind it is incomplete

```
When you ask questions, Claude uses these 12 tools automatically:
```

Verified: the table below names 12 distinct tools, so the number is self-consistent — but it omits `get_balance_history` and `get_goal_history`, both of which default mode does list (`READ_TOOL_DEFS.length` = 14). So a reader gets a surface that is 12/14 complete with no signal that it is partial.

This one is not just a missing needle: adding a needle for "12" would ratchet in an understatement. Either add the two rows and let the count be 14 with a needle, or reword so no count is asserted ("the tools behind the most common questions").

### 2. `CONTRIBUTING.md:58`, `:72`, `:110` — correct today, unguarded

All three lines currently match the registry ("33 base tools (14 read + 19 write)", "17 GraphQL-backed live read tools"), so this is not a live bug. But `CONTRIBUTING.md` restates those numbers in the same shape `CLAUDE.md` does, where they *are* guarded, and the checker's file list is package.json / README.md / CLAUDE.md / docs/index.html / docs/graphql-live-reads.md. Three `expectSubstring` blocks derived from `READ_TOOL_DEFS` / `LIVE_TOOL_DEFS` / `WRITE_TOOL_DEFS` close the class for the last prose surface that restates counts.

Also add `CONTRIBUTING.md` to `CHECKED_FILES` in `tests/scripts/check-tool-counts.test.ts` so the behavioural tests copy it into their temp tree.

### Why bother

#610 existed because a human-typed count drifted and nothing failed. After that fix, the gate covers every surface that restates counts except these two. Closing them makes "a count in prose is registry-derived or it does not ship" true without exception — which is the property that actually prevents the next instance, rather than fixing this one.

### LINKED ISSUE #729: A new scripts/ file is silently un-committable: nothing asserts that a script package.json references is tracked by git
`.gitignore:66` ignores `scripts/*` behind a hand-maintained per-file allowlist. A new
script that nobody remembers to allowlist is a **silent** `git add` no-op: exit 0, no
warning, file stays local. Every local gate then passes — the file is on disk — and CI is
the first thing to notice.

## It already happened

While implementing #692/#643 (PR #727), `scripts/check-workflows.ts` — the entire point of
that PR — was never committed for exactly this reason. `bun run check` and the pre-push
hook were both green on the author's machine. CI failed with `Module not found`, and the
AI reviewer independently diagnosed it on both passes. The instance was fixed by adding
`!scripts/check-workflows.ts`; the class was not.

That is `silent-failure-masking`, which is the most-recurring class in
[`docs/bugs/MINOR.md`](../blob/main/docs/bugs/MINOR.md) and still has no full entry.

## Suggested detector

For every `scripts/*` path referenced from a `package.json` `"scripts"` command, assert the
file is tracked by git:

```ts
// git ls-files --error-unmatch <path>  → must succeed
```

Roughly ten lines. It generalizes a little past `.gitignore` too: it also catches a
`check:*` entry pointing at a path that was renamed or deleted, which today fails only when
something runs it.

Placement is a judgment call — a sibling of `check-deps-pinned.ts` (which already asserts a
property *of* `package.json`) seems more natural than folding it into
`check-workflows.ts`, whose domain is `.github/workflows/`.

## Why it is worth the ten lines

A gate CI cannot run is worse than no gate: it is a green box over an unenforced invariant.
This repo now has nine `check:*` scripts and adds them regularly, so the exposure grows
with every one.

## Provenance

Raised by the AI reviewer on PR #727, twice, and deliberately deferred there rather than
growing a two-issue PR with a third gate in a different domain.

ignaciohermosillacornejo:
# Summary

Two claims the repo makes about itself that nothing checked.

- **#723** — the last two prose surfaces restating tool counts. `docs/EXAMPLE_QUERIES.md` said "12 tools" over a table of 12 that was missing `get_balance_history` and `get_goal_history`, so a needle for 12 would have ratcheted an understatement in. Table fixed first, then the count *and the table's contents* derived from `READ_TOOL_DEFS`. `CONTRIBUTING.md`'s three counts get needles.
- **#729** — a new file under `scripts/` is silently un-committable. `.gitignore` ignores `scripts/*` behind a hand-maintained allowlist, so `git add -A` skips an un-allowlisted script and exits 0. Every local gate passes, because every local gate reads the file off disk; CI, which clones, is the first to notice. New `check:tracked-files` gate, running inside `bun run check` — which is the point, since CI catching it is the status quo.

Closes #723
Closes #729

## External assumptions

None. Nothing here touches Copilot's API, data, or wire shapes — both gates read this repo's own files and its git index.

## Bug fix?

Root cause:       `.gitignore:66` ignores `scripts/*` behind a hand-maintained allowlist, so `git add` on an un-allowlisted path is a no-op that exits 0 and prints nothing; local gates read the file from disk and never notice it is not in the commit.
Bug class:        `silent-failure-masking` — an allow-list drop turning a real failure into a green signal. Most-recurring class in `docs/bugs/MINOR.md`, detector column was `none`.
Detector added:   `check:tracked-files` (`scripts/check-tracked-files.ts`, in `bun run check` and in CI), covering the class rather than the file: **any** file the repo's tooling reaches that git is untracked or ignore-matched, with "reaches" derived from package.json script references, the relative-import closure, and `scripts/…` string literals. Mutation-verified in `tests/scripts/check-tracked-files.test.ts` against real synthetic git repos.
Siblings checked: The gate's first run found eight latent instances, all tracked-but-ignore-matched and so one rename away from leaving the repo — seven scripts (three of them run by `package.json`: `pack-mcpb.ts`, `sync-manifest.ts`, `snapshot.ts`) plus `tests/unit/manifest-sync.test.ts`, swallowed by `MANIFEST-*` on a case-insensitive filesystem. All fixed. `MANIFEST-*` tightened to `MANIFEST-[0-9]*`, which describes the LevelDB artifact instead of guessing at it and retires the trailing whitelist that existed only to undo the over-match.
Ledger updated:   n/a — not an external-assumption bug. `docs/bugs/README.md` class table updated (`silent-failure-masking`: `none` → `partial`, scoped honestly to the one mechanism now closed).

---

## #723 — what changed

`docs/EXAMPLE_QUERIES.md` gains the two missing rows and states 14, derived from `READ_TOOL_DEFS.length`.

The interesting part is that a count needle alone would not have been enough. The old number was *self-consistent* — 12 over a table of 12 — which is exactly how it went stale. So the gate asserts the table for **set equality** against the registry, not just the digit above it. Adding a tool now fails `check:tool-counts` until the table gains a row.

`CONTRIBUTING.md` restated three counts in the same shape `CLAUDE.md` does, where they are guarded. Three `expectSubstring` blocks, and both files added to `CHECKED_FILES` in `tests/scripts/check-tool-counts.test.ts`.

Every count in this PR was derived from the registry this session, not typed from memory: 14 read, 17 live, 19 write, 6 swapped, 25 in live mode, 33 base, 44 in write mode, 50 definitions.

## #729 — what the gate does

Fails when a file the repo's own tooling needs is untracked, or is matched by a `.gitignore` rule. "Needs" is derived, never listed:

1. every repo-relative path named in a `package.json` script (a dangling one is reported too — that catches a renamed `check:*` target, which today fails only when something runs it);
2. the relative-import closure of those, plus every **tracked** file under `scripts/` and `tests/`;
3. `scripts/…` and `.github/…` paths named as text in those files, and in tracked `.github/workflows/*.yml` and `.husky/*` — the spawn-and-read case that no import graph reaches: `check-skills.py` → `dump-tool-names.ts`, `required-sections.yml` → `check-pr-sections.sh`, and `audit-severity-gate.test.ts` → `.github/audit-severity-gate.jq` (#744). Each path is resolved both repo-relative and relative to the file that named it, through a containment guard, so a `../`-prefixed literal cannot name anything outside the checkout.

The prefix list stays short on purpose. A general "any path-like literal" sweep would pull in the directories tooling *writes* — snapshots, generated fixtures, the demo database — which are ignored deliberately, and that is the cry-wolf failure.

Roots are tracked files, not everything on disk, so untracked scratch that nothing references stays unreported. That is what `scripts/local/` is for, and a gate that failed on it would be switched off within a week.

### Mutation results

Red direction, each verified by hand on the real repo and pinned as a test:

| perturbation | result |
|---|---|
| new `scripts/` file referenced from `package.json`, not allowlisted | fails, names the file (`git add -A` confirmed to skip it silently) |
| same file reached only by a side-effect `import` from a tracked test | fails, names the file |
| same file named only as a string literal in a tracked script | fails, names the file |
| tracked file that a `.gitignore` rule matches | fails, names the file and the rule |
| `package.json` script pointing at a path that does not exist | fails, names the path |

| script named only from a workflow, or only from a git hook | fails, names the file |
| `../`-prefixed literal naming an in-repo file | resolved to its in-repo path, not reported as an escape |

Green direction (cry-wolf controls): untracked `scripts/local/` scratch → passes; `dist/` build output named by the `build` script → passes; an actual `import` of a scratch file → still fails, because the carve-out applies to the string-literal sweep only.

### A bug this PR caused and fixed

The first version of the test inherited `GIT_DIR` from the pre-push hook, so `git init` in a temp directory re-initialised **the real repository** instead — `core.bare = true` in the shared config, plus a fixture identity appended to it. Every worktree of the repo stopped working and the next commit anywhere would have been authored by the test. Repaired, and fixed at the root: git is now spawned with every `GIT_*` variable stripped and `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` at `/dev/null`, in the test **and in the gate** — which runs from that same hook, and which would otherwise answer about whichever repository the hook belonged to. Pinned by a test that runs the gate under a bogus inherited `GIT_DIR`; drop the scrub and it goes red.

## Cheapest way to make each gate pass wrongly

**`check:tool-counts`** — edit the expected string in `scripts/check-tool-counts.ts` to match whatever the doc says. Nothing gates the gate; this is true of every needle #610/#720 added and is not new here. What *is* closed is the cheap path that used to exist: editing one digit in the doc. That now fails, because the table assertion is over set equality with the registry and does not read the digit. Adding a fake row (`` `get_balance_history` `` in a comment) would satisfy it — but that is a lie someone has to write deliberately into a doc, not an omission someone makes by accident, which is the class the issue is about.

**`check:tracked-files`** — add a `!path` negation to `.gitignore` **without** `git add`ing the file. That still fails, since untracked is checked independently of ignore-matching. The genuinely cheap false-green is deleting the reference that put the file in the closure (remove the `package.json` entry, or the import), at which point the file is by definition no longer needed and the gate is right to be quiet. The uncomfortable one is `GENERATED_PREFIXES`: adding a prefix makes everything under it invisible. That is hand-maintained and ungated — but its failure mode is the opposite of `.gitignore`'s, which is the whole point: an unlisted generated path fails loudly and gets added deliberately, rather than being dropped in silence.

## Review findings addressed

All three from the AI review, in `6838ced`, each mutation-verified:

1. **Workflow-invoked scripts were outside the closure (medium).** `scripts/check-pr-sections.sh` is named only from `.github/workflows/required-sections.yml`; it passed incidentally because it is already tracked. Tracked `.github/workflows/` and `.husky/` files are now sweep sources.
2. **The literal sweep had no repo-containment check (low).** Resolved both ways through the existing guard rather than rejecting `..` — which also generalised the sweep to `.github/…` and so covered the #744 `.jq` file.
3. **Ignore-matching is platform-dependent (low).** Documented: it honours `core.ignoreCase`, so local can be red while CI is green and never the reverse. All eight instances this gate found were macOS-only, which is why nothing had caught them.

Closure went 357 → 370 files.

## Also

- `docs/bugs/MINOR.md`'s intro asserted `silent-failure-masking` "appears **eight** times below" — nothing derived it and it had drifted by two. Replaced with the ordering claim (it is the most frequent class), which is the part worth stating and cannot go stale by one.


ignaciohermosillacornejo:
All three findings were real and are fixed in `6838ced` (branch also rebased onto `origin/main`, which had picked up #744/#745).

**1 — workflow-invoked scripts outside the closure (medium).** Correct, and the sharper version of it is that `scripts/check-pr-sections.sh` passed only *incidentally*: it is already tracked, so it entered through `TOOLING_ROOTS` rather than through any reference. Tracked files under `.github/workflows/` and `.husky/` are now sweep sources — text-swept, not import-walked, since neither has an import graph. Two new tests: a script named only from a workflow, and one named only from a hook. Mutation: drop the new roots and both go red.

**2 — no repo containment in the literal sweep (low).** Fixed, and fixing it properly turned out to close an adjacent hole. Rather than rejecting `..` tokens, the sweep now resolves each path **both** repo-relative and relative to the file that named it, through the same `toRepoRelative` guard the import resolver already used. So `'../scripts/target.ts'` inside a test resolves to the real in-repo file instead of escaping — which also generalises the sweep from `scripts/…` to `scripts/…` + `.github/…`, and that is exactly how `tests/scripts/audit-severity-gate.test.ts` reaches `.github/audit-severity-gate.jq` from #744: a `join()`ed path, invisible to the import graph, and a file a clean checkout must have. It was untested by this gate before your finding; it is covered now. Mutation: resolve only repo-relative and the containment test goes red.

The prefix list stays short on purpose (`scripts/`, `.github/`) — a general "any path-like literal" sweep would pull in the directories tooling *writes*, and that is the cry-wolf failure.

**3 — platform-dependent ignore-matching (low).** Documented in the `## Two failure kinds` block, with the direction stated: matching honours `core.ignoreCase`, which git sets at clone time from the filesystem, so local can be red while CI is green and never the reverse. Your Linux `git ls-files -c -i --exclude-standard` result is the confirmation — all eight instances were macOS-only, which is precisely why nothing had ever caught them.

Closure went 357 → 370 files. `bun run check` and `bun run check:skills` both exit 0.

PR #749, opened 2026-09-16
chore(audit): clear the LOW/MED audit backlog — seven issues, one sweep

### LINKED ISSUE #715: [Audit] Unaddressed suggestions from PR #713
## Unaddressed review suggestions from PR #713

**PR title:** docs(v3): migration guide + stale-claim sweep (#597 Tasks 7/8)

### LOW: The claim that the CHANGELOG's measured 9-key row carries `pending` is an unverified guess presented as fact, when the arithmetic only forces one of `pending`/`internal_transfer` to be present.

- **File:** `docs/tools-by-mode.md`:9
- **Reviewer said:** > the CHANGELOG's measured 9 is a real 100-row page whose rows carry `pending`
- **Evidence not applied:** docs/tools-by-mode.md:9 still reads "...the CHANGELOG's measured 9 is a real 100-row page whose rows carry `pending`" verbatim, unchanged from the reviewed text; no commit after the round-5 review (last commit is 08915fe, reviewed in that same round) touched this file.

### LOW: A comment reflow left a dangling one-word clause ("and the guide") on its own line, an edit scar from inserting sentences ahead of an already-wrapped paragraph.

- **File:** `tests/tools/tools.test.ts`:1228
- **Reviewer said:** > and the guide
- **Evidence not applied:** tests/tools/tools.test.ts:1227-1229 still contains the line break exactly as flagged: "...This is the third condition on that number,\n      // and the guide\n      // framed the dropped `category_name` as live mode's \"mirror case\"...", and the round-5 review (submitted 08:33:28) was the last review before merge (08:33:34) with no intervening commit.

---

### LINKED ISSUE #730: [Audit] Unaddressed suggestions from PR #716
## Unaddressed review suggestions from PR #716

**PR title:** fix(holdings): stop reporting positions on accounts get_accounts hides (#683)

### MEDIUM: preferredAccountName only trims/blank-checks the nickname field, so a whitespace-only `name` (e.g. `'   '`) still wins over a valid `official_name`, defeating the same identifiability rule the nickname trim was added for.

- **File:** `src/models/account.ts`:220
- **Reviewer said:** > The blank rule is applied to one of the helper's three inputs... `name: '   '` is truthy, so it wins over a perfectly good `official_name`... One predicate applied three times (`const usable = (s?: string) => (s?.trim() ? s : undefined)`) closes it and makes the rule match its stated reason for all three inputs.
- **Evidence not applied:** src/models/account.ts:220 only trims `account.nickname` (`account.nickname?.trim() ? account.nickname : undefined`) before the `||` chain on line 221 (`nickname || account.name || account.official_name`); `account.name` and `account.official_name` are used raw with no blank-after-trim check, and src/core/decoder.ts:1089 only drops an account when both `name` and `official_name` are absent, so a whitespace-only `name` alongside a real `official_name` is representable and unhandled.

### LOW: The migration guide and CHANGELOG still state the account-name resolution rule as the literal formula `nickname || name || official_name`, which is no longer exact now that blank-detection trims the nickname (a whitespace-only nickname is also treated as blank).

- **File:** `docs/MIGRATING-v3.md`:150
- **Reviewer said:** > The prose formula `nickname || name || official_name` is now inexact, in three places — two of them written by this commit... Cheapest version: say "prefers your nickname when you have set a non-blank one, otherwise the provider label" in the two docs (no formula to drift)
- **Evidence not applied:** docs/MIGRATING-v3.md:150 and CHANGELOG.md:64 both still contain the verbatim string "nickname || name || official_name", which omits the trim-based blank check added in src/models/account.ts:220, so the documented formula no longer matches the implemented rule.

### LOW: The AccountNode.name:resolvesNickname ledger entry doesn't note that cache mode now applies a whitespace-blank rule to nicknames that live mode (which does no nickname handling) cannot replicate, so the two modes could disagree on a whitespace-only nickname.

- **File:** `src/conformance/ledger.ts`:793
- **Reviewer said:** > the cache half now applies a blank rule the live half doesn't, so the two modes agree on every input except this one... It's worth a clause on that ledger entry, which already exists to say "this claim rests on the server and nothing re-checks it"
- **Evidence not applied:** The ledger entry at src/conformance/ledger.ts:793-813 for AccountNode.name:resolvesNickname describes cache mode mapping nickname to name and live mode doing no nickname handling, but contains no mention of whitespace/blank-nickname handling or the resulting mode divergence introduced by the trim added in src/models/account.ts:220.

---

### LINKED ISSUE #737: [Audit] Unaddressed suggestions from PR #732
## Unaddressed review suggestions from PR #732

**PR title:** chore(test): finish the review of #721

### MEDIUM: New file tests/helpers/ts-files.test.ts was never added to tsconfig.tests.json's include list, so it sits outside every static gate in `bun run check`, and the 'who imports what' comment still lists only the two prior importers of ts-files.ts as 'both on the list below'.

- **File:** `tsconfig.tests.json`:50
- **Reviewer said:** > Adding "tests/helpers/ts-files.test.ts" and updating the map is a two-line edit... an assertion moved out of tests/helpers/strip-comments.test.ts (on the list) and into a file that is not.
- **Evidence not applied:** tsconfig.tests.json's include array (lines 25-72) contains tests/helpers/ts-files.ts but not tests/helpers/ts-files.test.ts, and the comment at lines 43-45 still says ts-files.ts is imported by exported-constants.test.ts and no-hand-rolled-comment-strippers.test.ts 'both on the list below' with no mention of the third importer.

### LOW: The 'guards the guard' test's comment claims an empty EXTENSIONS would let the second assertion (JSX filter) pass, but toEqual(['.tsx']) would fail on an empty array too, so the stated rationale for the test is incorrect.

- **File:** `tests/helpers/ts-files.test.ts`:47
- **Reviewer said:** > the second would pass on an empty filter
- **Evidence not applied:** Line 47 of tests/helpers/ts-files.test.ts still reads 'the second would pass on an empty filter', unchanged from the flagged version, even though expect(jsx).toEqual(['.tsx']) with jsx=[] would fail, not pass.

### LOW: Object.fromEntries(EXTENSIONS.map(...)) still resolves to the untyped `any` overload because the callback's array-literal return isn't annotated as a tuple, so toEqual against `mapped` performs no compile-time check.

- **File:** `tests/helpers/ts-files.test.ts`:28
- **Reviewer said:** > Annotating the callback return (`(ext) => [ext, scriptKindFor(\`f${ext}\`)] as const`) restores the typed overload
- **Evidence not applied:** Line 28 is unchanged: `const mapped = Object.fromEntries(EXTENSIONS.map((ext) => [ext, scriptKindFor(`f${ext}`)]));` has no `as const`, so the map callback still returns ScriptKind[] arrays rather than tuples.

### LOW: regexLiteralsIn duplicates the AST walk that scanRegexLiterals already performs instead of having scanRegexLiterals also return the collected literals, leaving two near-identical tree walks in a file whose stated thesis is 'one lexer, not four'.

- **File:** `tests/no-hand-rolled-comment-strippers.test.ts`:236
- **Reviewer said:** > Returning `literals` alongside `considered`/`findings` from the existing scan removes the duplicate and gets `considered === literals.length` as a free invariant.
- **Evidence not applied:** regexLiteralsIn (lines 236-251) re-implements ts.createSourceFile + forEachChild to collect RegularExpressionLiteral text, identical in structure to scanRegexLiterals (lines 178-228), rather than reusing/extending scanRegexLiterals's output.

---

### LINKED ISSUE #738: [Audit] Unaddressed suggestions from PR #724
## Unaddressed review suggestions from PR #724

**PR title:** fix(guards): a security gate that vouched for files it never opened, and a fields fragment with no typo guard

### LOW: The perTree per-tree counter has a floor but no positive control, so moving its increment above the file read would leave every assertion in the file green while it silently overstates.

- **File:** `tests/docs/jsdoc-stranding.test.ts`:241
- **Reviewer said:** > perTree is floored but never controlled, and it is the one counter here that can still overstate
- **Evidence not applied:** Both end-to-end control tests (lines 312-339 and 341-369) assert probe.stranded, probe.files, and probe.unreadable but never probe.perTree, so the mutation the reviewer describes (moving result.perTree[tree]++ before the read) is unguarded.

### LOW: SCANNED_TREES is pinned with one literal array and the per-tree floor loop iterates a second, separately-written literal, so updating the pin after adding a tree still leaves the new tree with no floor.

- **File:** `tests/docs/jsdoc-stranding.test.ts`:287
- **Reviewer said:** > line 287 pins the tree list with a literal and line 288 loops over a SECOND separate literal of the same three names
- **Evidence not applied:** Line 287 is expect([...SCANNED_TREES].sort()).toEqual(['scripts', 'src', 'tests']) and line 288 is for (const tree of ['src', 'scripts', 'tests']) — two independent literals rather than one shared PINNED_TREES constant.

### LOW: linkTarget's bare catch block conflates 'not a symlink' with 'lstat/readlink failed' with no comment noting the safety of this depends on a co-dependency with the caller.

- **File:** `scripts/check-concealment.ts`:435
- **Reviewer said:** > linkTarget ends in a bare catch { return undefined }, which conflates "not a symlink" with "could not tell"
- **Evidence not applied:** Lines 435-442 still show `catch { return undefined }` with no explanatory comment, unchanged from what the round-7 review flagged.

---

### LINKED ISSUE #740: [Audit] Unaddressed suggestions from PR #739
## Unaddressed review suggestions from PR #739

**PR title:** fix(stranding): pin the one counter in the sweep that could still overstate

### LOW: The new Round-7 perTree comment paragraph was left between the pre-existing 'readable sibling' comment and the assertions it explains, instead of being reordered so each comment sits directly above the assertion it justifies.

- **File:** `tests/docs/jsdoc-stranding.test.ts`:363
- **Reviewer said:** > the new 9-line block was inserted between the pre-existing two-line comment and the assertion it annotates ... Putting the new paragraph after the old one but above perTree, or leading with it, would keep each comment adjacent to what it justifies.
- **Evidence not applied:** Lines 363-375 still show the 'And the readable sibling is still swept...' comment (363-364) immediately followed by the new Round-7 perTree paragraph (366-373), both sitting above both expect(probe.files) and expect(probe.perTree.src) — the reordering was not applied.

---

### LINKED ISSUE #742: [Audit] Unaddressed suggestions from PR #736
## Unaddressed review suggestions from PR #736

**PR title:** fix(auth): a dead Copilot token is the logged-out state, not a mystery

### MEDIUM: No test pins the fast-path catch's rethrow direction, so a swallow-all mutant of the catch block passes the whole suite.

- **File:** `tests/core/auth/candidate-ordering.test.ts`:456
- **Reviewer said:** > The new test (`candidate-ordering.test.ts:456`) pins *fall-through on `INVALID_REFRESH_TOKEN`*; nothing pins *rethrow on anything else*.
- **Evidence not applied:** The only fast-path test in the file (test.each at line 456, 'a dead CACHED token falls through...') only exercises INVALID_REFRESH_TOKEN; no test in the file sends a 403/endpoint-level error through the fast path to assert it is rethrown rather than swallowed.

### MEDIUM: The fast-path catch guard only checks isExplainedByLoggedOut, unlike the cold path which gates on isCandidateRejection first, leaving the two paths structurally inconsistent.

- **File:** `src/core/auth/firebase-auth.ts`:132
- **Reviewer said:** > The fast path skips the `isCandidateRejection` gate the cold path applies first — src/core/auth/firebase-auth.ts:132
- **Evidence not applied:** Line 132 reads `if (!isExplainedByLoggedOut(err)) throw err;` with no isCandidateRejection check, while the cold-path loop at line 151 checks `if (!isCandidateRejection(err)) throw err;` before ever calling isExplainedByLoggedOut — the suggested `isCandidateRejection(err) && isExplainedByLoggedOut(err)` combination was not added.

### LOW: The fast-path docblock still calls isExplainedByLoggedOut without naming or explaining that its first disjunct (isForeignProjectError) is claimed impossible on this path.

- **File:** `src/core/auth/firebase-auth.ts`:118
- **Reviewer said:** > The fast path calls a predicate whose first disjunct its own docblock declares impossible — src/core/auth/firebase-auth.ts:118
- **Evidence not applied:** The docblock at lines 116-125 still says the server-issued token 'can never be foreign' and the guard at line 132 still calls the combined isExplainedByLoggedOut (which starts with isForeignProjectError) with no added clause reconciling the two claims.

### LOW: A redundant third getIdToken() assertion in the fast-path-fallthrough test runs cold-path-only and does not actually exercise the fast path it claims to pin.

- **File:** `tests/core/auth/candidate-ordering.test.ts`:517
- **Reviewer said:** > The second assertion in the new fast-path test doesn't test the fast path — tests/core/auth/candidate-ordering.test.ts:513
- **Evidence not applied:** Lines 513-517 still contain the same `await expect(auth.getIdToken()).rejects.not.toThrow('Firebase token exchange failed')` as a third getIdToken() call, by which point this.refreshToken is already null from the prior call, so it runs cold-path-only exactly as the reviewer described.

---

### LINKED ISSUE #743: [Audit] Unaddressed suggestions from PR #725
## Unaddressed review suggestions from PR #725

**PR title:** fix(test): the suite's own cleanup and assertions were tied to moments bun does not give

### MEDIUM: The specimen fixture never pins the `ts.isExportAssignment` (`export default x`) exclusion, so deleting that clause from the detector's exclusion list would leave the whole suite green instead of failing.

- **File:** `tests/fixtures/collection-time-assertion-sample.ts`:40
- **Reviewer said:** > ts.isExportAssignment is NOT pinned: there is no export default anywhere in the specimen or in tests/, so deleting that clause leaves the suite green. Add a must-NOT-flag line `export default assertRow;` to the specimen
- **Evidence not applied:** tests/fixtures/collection-time-assertion-sample.ts contains only `export { assertRow };` and `export const assertions = { assertRow, row: assertRow };` (lines 40, 45) — no `export default` line was added, so the ExportAssignment branch in tests/no-collection-time-assertions.test.ts:236 remains unpinned by any test.

### LOW: `testFiles()` returns files in `readdirSync` filesystem order rather than sorted, so a multi-finding failure lists files in an order that differs between a dev machine and CI.

- **File:** `tests/no-collection-time-assertions.test.ts`:100
- **Reviewer said:** > `testFiles()` returns `readdirSync` order, not sorted... A `.sort()` on the return would cost nothing. (Third round open.)
- **Evidence not applied:** tests/no-collection-time-assertions.test.ts:100-108's testFiles() pushes `full` in readdirSync iteration order and returns `out` with no `.sort()` call anywhere in the function.

### LOW: The reviewer repeatedly noted (rounds 5 and 6) that `isNamedWithoutBeingCalled` is a denylist that keeps growing by one AST shape per round and proposed inverting it to a two-shape allowlist (callee or call argument), or alternatively documenting the remaining gaps (property-name, property-access, binding-element, type-query, array-element, variable-initializer positions) in the 'WHERE IT MAY OVER-REPORT' block — neither was done.

- **File:** `tests/no-collection-time-assertions.test.ts`:231
- **Reviewer said:** > Invert to an allowlist: only flag an identifier that is the callee of a CallExpression or one of its arguments... take it or add the positions to the over-report block.
- **Evidence not applied:** tests/no-collection-time-assertions.test.ts:231-238 still implements the five-clause denylist (VariableDeclaration name, FunctionDeclaration name, ExportSpecifier, ExportAssignment, ShorthandPropertyAssignment, PropertyAssignment initializer) with no allowlist inversion, and the header's 'WHERE IT MAY OVER-REPORT' section (lines 52-60) does not mention the property-name, property-access, binding-element, type-query, array-element, or variable-initializer shapes the reviewer identified as still unflagged-for-the-wrong-reason.

---

ignaciohermosillacornejo:
# Summary

Triage and clearance of the `[Audit] Unaddressed suggestions` backlog: 12 open issues, verified one finding at a time against current `main`.

**Five closed without code** — every finding already satisfied or overtaken, each closed with the `file:line` that settles it: #700 (stale — the hand-rolled three-pass `stripComments` those docblocks described was replaced by the parser-owned helper in #721), #719, #731, #734, #735.

**This PR fixes the seven that were still live.** Per the repo convention, LOW/MED batch into one cleanup PR; nothing in the backlog graded HIGH, so nothing is split out.

Two findings reach runtime. The rest are gates that could not fail, and comment/doc precision.

### Runtime

- **`preferredAccountName` applied its own rule to one of three inputs** (#730). The docblock argues blank-is-not-a-name from identifiability, then trimmed only `nickname`, so `name: '   '` was truthy and beat a valid `official_name`. Representable: `processAccount` drops an account only when `name` **and** `official_name` are both absent. One `usable()` predicate now runs on all three. Mutation-verified — reverting either widened operand kills two tests.
- **The auth fast path skipped a gate the cold path applies first** (#742). `isExplainedByLoggedOut` is a substring test over any `Error`'s message, so an endpoint-level 403 whose body quoted a dead-token code was read as a verdict on the token, swallowed into a cold re-extract, and reported as "log in" for an outage logging in cannot fix. Both paths now gate `isCandidateRejection` first, in the same order.

### Gates that could not fail

- **Nothing sent a non-dead-token error through the fast path** (#742), so `catch {}` passed the whole suite. New `test.each` covers both directions; mutation-verified: swallow-all kills two rows, dropping the new `isCandidateRejection` gate kills the row that targets it specifically.
- **The collection-time specimen had no `export default`** (#743), so deleting the detector's `ts.isExportAssignment` clause left the suite green. Added — verified the clause's deletion now goes red.
- **`tests/helpers/ts-files.test.ts` sat outside every static gate** in `bun run check` (#737): added to `tsconfig.tests.json`, with the "who imports what" map corrected to name all three importers.
- **The stranding sweep pinned its tree list with one literal and floored a second** (#738), so a new tree could arrive with no floor. One `PINNED_TREES` feeds both — verified that adding a tree to both now fails on the missing floor rather than passing.

### One finding measured and rejected

#737 also prescribed `as const` on the `Object.fromEntries(EXTENSIONS.map(...))` callback, on the premise that it resolves to the untyped `any` overload and leaves the table below it unchecked. Measured instead of taken: `tsc -p tsconfig.tests.json` reports `mapped` as `{ [k: string]: ts.ScriptKind }` **with the annotation and without it** — probed by assigning `mapped` to a `string` and reading the type out of the error, which `any` would not have produced. The premise does not hold, so the annotation is not applied and the measurement is recorded where the next round will read it. The real half of that finding stands and is fixed above: the file was outside every program in `bun run check`, so whichever overload it resolved to was moot.

### Comment and doc precision

The account-name formula in three docs, no longer exact once blank-detection trims (#730); a ninth-key claim the arithmetic does not force (#715); a wrong vacuity rationale — the assertion it credited would fail, not pass (#737); a duplicated AST walk folded into the single scan, in the file whose thesis is one lexer not four (#737); the six denylist positions documented in the over-report block instead of growing a seventh clause (#743); `readdirSync` order sorted so a multi-finding failure reads the same on CI as locally (#743); a bare catch whose safety is a co-dependency with its caller, now stated (#738); a comment reunited with the assertion it justifies (#740); an edit scar (#715); and a ledger entry that did not record the whitespace-nickname divergence only cache mode can create (#730).

Closes #715
Closes #730
Closes #737
Closes #738
Closes #740
Closes #742
Closes #743

### Round-1 review addressed

Five comment findings and one substantive one, plus a disagreement between the two reviewers that I settled by measuring:

- **Importer map was short one importer** — `tests/helpers/strip-comments.ts:55` imports `scriptKindFor`, so there are **four**, not three. Reviewer A said four; reviewer B verified "three is accurate" after reading `strip-comments.test.ts`, which names the module in prose without importing it. A is right. Both lists (the tsconfig map and the contract test's docblock) now name the same four.
- **Denylist count off by one** (both reviewers) — six clauses, so the next is a seventh. The PR body was right and the comment had drifted, which is the failure mode that block exists to prevent.
- **My replacement vacuity rationale overcorrected** — the table test compares against a hand-written four-key literal, so dropping `.ts` *does* fail it; only the JSX test is derived-and-silent. Corrected, with the reason the floor is still worth keeping.
- **`linkTarget`'s comment claimed loudness for a disjunct that doesn't earn it** — narrowed to what the caller supports, residual named rather than waved at.
- **`as const` (reviewer A, finding 4)** — reviewed against the pre-revert commit; the measurement in `13c7fb4` is the answer. No action.
- **#737 was fixed instance-only** (reviewer B) — fair, and the one finding that wasn't a comment. A helper module judged worth typechecking whose own contract test is in no program is a class, and nothing ratcheted it. Added a second rule to `tests/unit/tsconfig-tests-sync.test.ts`: a `tests/helpers/<name>.test.ts` importing a `<name>.ts` on the include list must be on it too. Scoped narrowly so it holds today without fighting the ~87 pre-existing errors. **Mutation-verified** by removing `ts-files.test.ts` from the list — it reproduces the #737 state and fails.

**Filed, not fixed:** #751 — `exchangeToken` clears a known-good refresh token on any non-OK response, so an endpoint-level 403/429 costs a browser-wide re-extract. Pre-existing and out of scope for a LOW/MED sweep; it is the third place the #722 endpoint-vs-token distinction isn't drawn.

## External assumptions

None. The one ledger edit (`AccountNode.name:resolvesNickname`) records that the existing probe could not cover whitespace-only nicknames and that cache mode now applies a blank rule live mode cannot replicate — it narrows an existing `verified-once` claim rather than adding a new assumption.

## Verification

- `bun run check` — exit 0 (3396 pass, 0 fail)
- `bun run check:skills` — exit 0
- Mutation-verified, honest direction, each killing the test written for it: swallow-all catch, dropped `isCandidateRejection` gate, reverted `usable()` operands, deleted `ts.isExportAssignment` clause, and a tree added without a floor.


ignaciohermosillacornejo:
Round-1 review addressed in `bbf4c76`. One note first, because the two reviews contradicted each other and I'd rather record the measurement than the verdict.

**The importer map: four, not three.** Review 1 said `tests/helpers/strip-comments.ts:55` also imports `scriptKindFor`; review 2 explicitly verified the opposite ("Good catch that it *is* three") after reading `strip-comments.test.ts`, which names the module in prose without importing it. Grepping import statements only:

```
tests/exported-constants.test.ts:143
tests/no-hand-rolled-comment-strippers.test.ts:96
tests/helpers/ts-files.test.ts:21
tests/helpers/strip-comments.ts:55
```

Review 1 is right. Both lists — the tsconfig map and the contract test's own docblock, which named a different three — now name the same four and say so, since a curated list read as a dependency map is exactly where this costs something.

**Fixed as flagged:** the six-vs-five denylist count (both reviews; the body was right and the comment had drifted, which is the shape that block exists to stop); my replacement vacuity rationale, which overcorrected — the table test pins a hand-written four-key literal, so dropping `.ts` *does* fail it and only the JSX test is derived-and-silent; and `linkTarget`'s comment, narrowed to the claim the caller supports with the residual named rather than waved at.

**#737 was fixed instance-only — fair, and now ratcheted.** The class is "a helper module judged worth typechecking whose own contract test is in no program", and `tsconfig-tests-sync.test.ts` only knew about mock-graphql adoption. Added the second rule roughly as proposed: a `tests/helpers/<name>.test.ts` importing a `<name>.ts` on the include list must be on it too — scoped that narrowly so it holds today without fighting the ~87 pre-existing errors, and saying nothing about the two scanners that are off the list on purpose. Mutation-verified by removing `ts-files.test.ts` from the include list: it reproduces the #737 state exactly and fails with the message naming the file.

**`as const` (review 1, finding 4):** that review ran against the pre-revert commit. `13c7fb4` has the measurement — probing `mapped` through a deliberate type error under `tsconfig.tests.json` reports `{ [k: string]: ts.ScriptKind }` with the annotation *and* without it, and `any` would have produced no error at all. Happy to be shown a config where it resolves differently, but under this repo's own it does not.

**Filed rather than fixed:** #751 for `exchangeToken` clearing a known-good refresh token on any non-OK response — correctly identified as pre-existing and out of scope. It is the third place the #722 endpoint-vs-token distinction isn't drawn, so it wanted an issue rather than a footnote.

`bun run check` and `bun run check:skills` both exit 0.

ignaciohermosillacornejo:
Round-2 addressed in `0748edb`. All four taken, plus the optional one.

**The vacuity gap is the one worth calling out.** You're right, and it's worse than a nit: the ratchet I added to close an instance-only fix floored its *walk* but not the *filter the rule turns on*. `pairs` comes off the filesystem, so it stays full regardless; `included.has(subject)` is what can go empty, and that file's own header calls expanding to `tests/**/*` tracked work. That's the could-not-fail shape four of this PR's seven fixes were about, reintroduced by the fix for the fifth. Floored on the filtered list now, and verified the way the rest of this PR was — turning the include list into a glob makes it fail with the message naming the glob case, rather than passing while checking nothing.

**Also taken:** the unsorted `readdirSync` (fair — same commit range, same argument, and it feeds a multi-line failure message); the two colliding sixes, now named separately as denylist DISJUNCTS vs missed SHAPES; and the floor comment's misattribution — what the table derives from `EXTENSIONS` is the key set, the values are hand-listed, and the comment says that instead of crediting the docblock with a claim it doesn't make.

**And the optional gate, taken.** You're right that it was the one thing in the PR still resting on a probe I ran once by hand — the exact category the ledger exists to distrust, in a file whose neighbours are gates against it. It's now a `@ts-expect-error` on an assignment that compiles only while `mapped` is not `any`:

```
=== MUTANT: force `mapped: any` ===
tests/helpers/ts-files.test.ts(48,5): error TS2578: Unused '@ts-expect-error' directive.
```

So if `Object.fromEntries` ever does fall to the untyped overload — the thing review 1 believed was already happening — `bun run check` says so, instead of a comment claiming it doesn't.

`bun run check` and `bun run check:skills` both exit 0.

ignaciohermosillacornejo:
Round-3 addressed in `235172d`. Both findings and the suggestion.

**The type gate was testing a weaker claim than the paragraph above it** — you're right, and I measured it rather than reasoning about it, since that's the whole point of that gate existing. With the old one-level form (`string = mapped`) and a values-only widening (`mapped: { [k: string]: any }`), `tsc` reports **nothing**: the directive stays used, the gate stays green, and the four values go unchecked. That is exactly the failure it was added to catch, so the gate was decorative for the case most likely to actually happen.

Indexed now. Verified against both mutants:

```
MUTANT A (whole expression any):  TS2578: Unused '@ts-expect-error' directive.
MUTANT B (values-only any):       TS2578: Unused '@ts-expect-error' directive.
```

**`walk()` sorted** — fair, and slightly embarrassing nine lines above the walk that got `.sort()` in the previous commit for the identical reason.

**The suggestion taken.** `tests/unit/tsconfig-tests-sync.test.ts` is now on the include list it is itself the gate for. Your argument is the one the header already makes for the two other gates: it casts the parsed JSONC (`as { include: string[] }`) and builds two failure messages by interpolating derived values, which is the #725 shape verbatim. And you were right that the new pairing rule can't reach it — scoped to `tests/helpers/` on purpose — so it wanted a deliberate manual add with a note saying why, not a ratchet. Verified type-clean before adding rather than after; the header now names all three and the reason for the third.

`bun run check` and `bun run check:skills` both exit 0.

ignaciohermosillacornejo:
Round-4 addressed in `6a88a28`. All five.

**The self-membership ratchet** — you're right that it can ratchet itself, and that it isn't circular: `bun test` collects the file from the filesystem regardless of any tsconfig, so removing the entry leaves the assertion running and red. Verified exactly that way. Keyed off `import.meta.path` as suggested, so a rename reports the rename rather than a false membership failure.

**The type gate was still one implication short.** Correct, and the same shape as round 3 one level down — a suppressor fires on *any* error, so `string = mapped['.ts']` says "not assignable to `string`", which `unknown` satisfies. Took the pair:

```
MUTANT (values become `unknown`, with the suppressor alone):  passed silently
MUTANT (values become `unknown`, with the pair):              TS2322: Type 'unknown' is not assignable to type 'ScriptKind'.
```

Neither half is the claim — the positive assignment rejects `unknown`/`string`/a widened union, the suppressor rejects the `any` the positive one waves through. Together they're exact, which is what the comment claims.

**The stale "17-file allowlist"** — dropped rather than updated, for the reason you gave: membership is the load-bearing half and is checkable, while a bare count of another file's list is the exact drift shape this PR has now corrected three times (denylist five→six, importers three→four, and this). Same treatment the header paragraph got — **finding 5** is now names, not a count over a fuzzy antecedent.

**Two parses, two Sets** — hoisted to one module-level `included`. Fair; it's the `PINNED_TREES` argument applied to something this PR itself introduced.

`bun run check` and `bun run check:skills` both exit 0.

ignaciohermosillacornejo:
Round-5 addressed in `31a517a`. All three.

**The "two temp-db support modules they read" claim was wrong about the mechanism**, and you were right to flag it as load-bearing rather than cosmetic. Nothing imports them — `temp-db-suite-teardown.test.ts:52` holds the probe as a path string it *spawns*, and `:185` names the teardown module in prose while bunfig.toml preloads it. That matters because being followed in by `tsc` is exactly the mechanism the importer map two paragraphs down turns on, so describing these two as "read" pointed at the wrong reason for their being on the list. They're listed in their own right; the header says that now.

**"The third is…"** — an ordinal indexing into a sentence that now names five things, introduced by the edit that was removing a fuzzy count. Replaced with the name.

**The self test moved to its own `describe`.** Agreed — it was printing `a typechecked helper brings its own contract test onto the list > this gate is itself typechecked`, a title contradicting the test under it, and `included` being module-scope now makes the move free. Its failure message also names the glob case, since it and the helper-pairing floor both compare literal paths and would go red together; they shouldn't send the reader to two different remedies.

Re-verified the ratchet still fires after the move: removing the entry from the include list goes red with the new message.

`bun run check` and `bun run check:skills` both exit 0.

ignaciohermosillacornejo:
Round-6 addressed in `8bceb58`. All three.

**The generalization, properly this time.** You're right that round 4 proposed it and only the self half landed — and that the one entry I ratcheted was the one that happens to hold the parsed list, which is the least interesting of the five. The other four were sitting exactly where this file sat two commits ago. One pinned literal now names all five, read by the assertion, with `import.meta.path` still keying this file's own entry so a rename reports the rename rather than a phantom membership failure. Added an on-disk check alongside it for the same reason, generalized to the other four.

Mutation-verified on entries *other* than self:

```
delete "tests/no-collection-time-assertions.test.ts"  -> red
delete "tests/fixtures/temp-db-leak-probe.ts"         -> red
```

**And the count-about-counts.** This one is genuinely funny and entirely fair: the parenthetical I wrote to justify deleting `17-file` said "had been 29 for a while" — and the branch was already **31**, because this PR's own two additions moved it between `origin/main` and the commit that wrote the sentence. A sentence arguing that counts of another file's list go stale, going stale before it was pushed. Deleted rather than corrected to 31, which would just have restarted the clock; the shape is the argument now.

**Header subject** — named two categories over a three-item list whose third member is neither. Rewritten, and it now also points at the pin, so the header and the ratchet reference each other rather than the header standing alone as a request.

`bun run check` and `bun run check:skills` both exit 0.

ignaciohermosillacornejo:
Round-7 addressed in `f878d5f`. All three.

**Re-derived rather than taken**, since the finding is itself a counting claim. Applying the pin's stated criterion — not reached by any import, not a mock-GraphQL adopter — to all 31 include entries returns **13**. Two are the `tests/helpers/` contract tests the helper-pairing rule already covers, leaving **11** for this pin, which named 5. Your six are exactly the difference:

```
tests/core/auth/browser-token.test.ts
tests/core/auth/candidate-ordering.test.ts
tests/core/auth/firebase-auth.test.ts
tests/exported-constants.test.ts
tests/integration/uid-transition-sweep.test.ts
tests/no-hand-rolled-comment-strippers.test.ts
```

And you're right that the auth three are the sharpest case — the tsconfig header justifies their membership with the #725 story verbatim (a required field appeared on `TokenResult` and nothing failed, because the files were in no program). They were the entries most worth pinning and were the ones not pinned. Mutation-verified on two of the six newly added, not on ones already covered.

The criterion is now written next to the literal, with why the list stays hand-maintained: deriving it needs a real import graph rather than a grep, so the pin and the header cross-reference each other and drifting apart is the intended failure direction.

**And the count, again — third time.** The clause I added *last commit* to point at the pin said "one of those **five** lines", a bare count over a list that had just moved into another file. That's the same shape as the `17-file` deletion it was written next to. Count-free now, phrased as the criterion instead.

**Finding 3 taken too** — the parenthetical had started narrating this PR's review history rather than stating the rule. Trimmed to the rule.

`bun run check` and `bun run check:skills` both exit 0.

ignaciohermosillacornejo:
Round-8 addressed in `03607e7`. Both, plus the optional third.

**The universal was false and the counterexample was one file away.** The clause I added to point at the pin said every unimported entry is pinned there — but the two `tests/helpers/` contract tests deliberately aren't, and the pin's own docblock carves them out *by name*. So the header overstated the pin's extent, and the file it was cross-referencing is the one that would have corrected it. It now says every such entry is **ratcheted** — by the pin or by the helper-pairing rule — and sends the reader to the file that says which. Same information, no claim the neighbouring file contradicts.

**And the count, a fourth time**, this one inside the paragraph explaining why the list *isn't* derived, added by the commit that deleted two other counts. "Thirteen entries meet it" adds nothing the criterion doesn't already say and rots on the next include-list edit. Gone, replaced by the rule and its one deliberate exclusion.

Third item taken too — "No count, twice learned" was narrating rather than stating.

Re-verified the ratchet still fires after the prose edits (deleting a pinned auth entry goes red).

`bun run check` and `bun run check:skills` both exit 0.

---

Worth noting where this has converged: rounds 1–2 found real defects (a wrong importer count, an off-by-one, a vacuous ratchet, a type gate weaker than its claim). Rounds 6–8 have been my own explanatory prose acquiring the exact drift shapes the PR set out to remove — four separate counts, in four commits whose subject was deleting counts. That is a fair cop each time and they are all now gone, but it does suggest the remaining surface is comment prose rather than anything the code does. Happy to keep going if there's a substantive finding left; otherwise this is ready from my side.

ignaciohermosillacornejo:
Round-9 addressed in `572d75d`. Both.

**The universal was false and in the wrong direction.** Verified your grep: `grep -rn "\.test\.js'" tests/ src/ scripts/` returns nothing, so "an entry here that nothing imports" is *every* test entry, not the thirteen I was reasoning about — which makes the pin a minority holder and "most" exactly backwards. Worse, the sentence whose job is to say which rule holds which didn't name the adoption rule at all, and it's the largest of the three.

Rewritten to state the mechanism rather than a distribution: nothing imports a `.test.ts`, so for every test entry the *line* is the only thing putting it in a program — delete it and the checking is gone, not relocated. Then all three ratchets named, with the adopter carve-out explicit so the header and the pin's criterion describe the same set. That's the part that matters: they're only a drift detector for each other while they're comparable.

**And "checkable" — you were right that the word was the finding.** A property, not a promise. `known-fields-wire-parity`'s `MIRROR_IS_EXACT` pins sit in `src/` *because* that file isn't typechecked, so adding it to the include list would leave those pins justified by a premise that had silently stopped holding. The negative is asserted now, mutation-verified:

```
add "tests/tools/live/known-fields-wire-parity.test.ts" to the include list
  -> red: "...is now typechecked, which invalidates the reason its MIRROR_IS_EXACT
     pins live in src/ instead of in it. Either move those pins into it and delete
     this assertion, or take it back off the include list."
```

The comment now says "IS checked" and points at it. That one was worth the two lines — it's the only claim left in the file that something else's correctness rests on.

`bun run check` and `bun run check:skills` both exit 0.

ignaciohermosillacornejo:
Round-10 addressed in `dd3a95a`. All three, and I agree with your read on where this stands.

**Finding 1 was the real one**, and it's the good kind: the negative assertion I added last round fails *open*. `included.has(absent)` is false whether the file is genuinely absent or merely spelled by a glob, so the one rule whose verdict is "still absent" would go green for exactly the wrong reason — while the two positive rules go red. Three rules, one shared premise, three prose caveats and no check.

So the premise is asserted now rather than warned about: `no include entry is a glob`. The caveats come out of the three failure messages, and the glob day still breaks this file — which is correct and is the point — but it breaks with one sentence saying what to do, instead of three rules failing for reasons that look unrelated to each other. Mutation-verified by turning one entry into `tests/helpers/*.ts`.

**Finding 2** — the same `describe`-title nit this file fixed in `31a517a`, and I reintroduced it one commit later by filing a "stays OFF the list" test under "entries whose membership is the coverage stay on the list". Its own block now.

**Finding 3** — "every test entry" over a list containing two non-test modules, which are precisely the ones where the argument matters most since nothing imports them either. Fixed to name them.

---

On convergence: agreed, and thank you for the ten rounds — rounds 1–2 caught four things that mattered (a wrong importer count both reviewers disagreed about, an off-by-one, a ratchet that was vacuous on the operand it turned on, and a type gate weaker than the claim above it), and rounds 3–10 have mostly been my own explanatory prose acquiring the drift shapes this PR set out to remove. Which is its own small lesson: the comments describing a gate rot faster than the gate, and four of the last six findings were counts I wrote while deleting counts.

From my side this is ready. Not merging or enabling auto-merge — leaving that to @ignaciohermosillacornejo.

ignaciohermosillacornejo:
Round-11 addressed in `c9028e9`. All three.

**A count, in the docblock about an unchecked premise, in a PR that has deleted six of them.** Fair and slightly absurd. Five assertions read `included`, and under either reading of "three" the one omitted is the adoption rule — the largest ratchet and the loudest false red on glob-day. Count-free now, and it matches the sentence directly below it that already said "Every rule here".

**The over-reach**, verified per entry rather than reasoned about. Applying the criterion to all 31 entries, exactly three are covered by no rule:

```
tests/helpers/mock-graphql.ts    IMPORTED by 15
tests/helpers/strip-comments.ts  IMPORTED by 3
tests/helpers/ts-files.ts        IMPORTED by 4
```

All three are helper *modules* with importers on the list, so `tsc` reaches them whether or not their own line survives — they need no pin, and my widened sentence claimed otherwise. The paragraph now scopes the argument to the entries it holds for and names the exception with its reason.

**And the rewrap scar** — "nothing imports them; the probe is a path string, the teardown module is preloaded" ended up standing in both halves of the paragraph I split. Consolidated.

`bun run check` and `bun run check:skills` both exit 0.

ignaciohermosillacornejo:
Round-12 addressed in `734c44f`. All three.

**An ordinal, two lines under the sentence explaining why this header names things instead of counting them** — and the same construct this PR already removed from this file once ("The third is tsconfig-tests-sync.test.ts"). You're also right that "needs" was false for one of the two: the teardown module is preloaded by bunfig.toml for every run, so `temp-db-suite-teardown.test.ts` is its *gate*, not its consumer. Both closed by naming the modules and asserting nothing about the relation — the mechanism paragraph below already carries it.

**"differently and silently"** — good catch, and the distinction is the one the `describe` exists for. A rule going red is loud; what's wrong is that it names the wrong cause. Now: "each breaks differently: the positive rules loudly and for the wrong reason, the negative one silently" — plus the sentence saying that loud-but-misattributed is precisely why this is one premise test rather than a caveat on each rule.

**The orphaned antecedent** — "that reason" lost its referent when I split the paragraph two commits ago, and the nearest "reason"-shaped thing a reader now lands on is the ratchet paragraph. Names `#725` directly.

`bun run check` and `bun run check:skills` both exit 0.

For the record, since it's been the through-line: this round's first finding is the fourth time an anti-drift sentence in this header has itself drifted — first the count, then the ordinal, then the count again, now the ordinal again. The prose explaining a ratchet appears to rot at roughly the rate the ratchet prevents, which is a decent argument for the ratchets and a poor one for the prose.

ignaciohermosillacornejo:
Round-13 addressed in `fead0c4`. All three — and finding 2 was right in a way worth showing rather than asserting, so I ran it.

Replacing the include list with `tests/**/*`, the scenario the tsconfig header names:

```
adoption rule                RED
pairing FLOOR                RED
membership pin               RED
pairing RULE                 green   <- typechecked is empty, missing is []
negative membership rule     green
premise test                 RED
```

So **three** outcomes, not two. The pairing rule is a positive rule that does not go red — it goes vacuously green, caught only by its own floor, which is exactly the defect that floor was added for earlier in this PR. My docblock was describing the file as it stood *before* that fix. And only the negative rule is silent end to end; the rest are legible once you know to look here. The docblock now lists the three cases and says which catches which.

**Finding 1** — "every rule in this file" is falsified by the premise test itself and the self-name test, neither of which reads the include list. Scoped to the include-list rules.

**Finding 3** — both rewrapped paragraphs left a short line mid-sentence. Fixed.

`bun run check` and `bun run check:skills` both exit 0.

---

This is the second time this round's substantive finding was "the comment describes the code as it was one fix ago" rather than a disagreement about what the code should do. Both times running the scenario settled it in a minute where reading it had produced a confident wrong sentence. Worth remembering next time I'm tempted to describe a failure mode I haven't triggered.

ignaciohermosillacornejo:
Round-14 addressed in `501bda9`. Finding 1 is the best catch of the whole review, and it's against the guard I added two commits ago to prevent exactly the thing it lets through.

`/[*?]/` does not match `tests`. And a bare directory in a tsconfig `include` is tsc's own shorthand for `tests/**/*` — so the single most likely spelling of the tracked follow-up sails past the premise test, and then the negative membership rule goes silently green while `known-fields-wire-parity.test.ts` is in fact typechecked and its `MIRROR_IS_EXACT` pins in `src/` rest on a premise that stopped holding. A guard for "the list stopped being literal paths" that only recognises one of the two ways that happens.

Now asserts the property the rules actually need — every entry names an existing **file** — so globs, directories and missing paths all fail alike. Mutation-verified on both spellings, and confirmed the old predicate's asymmetry:

```
old /[*?]/ on "tests/**/*"  -> flagged
old /[*?]/ on "tests"       -> NOT flagged     <- the fail-open
new check, both spellings   -> red
```

Non-vacuous by fact rather than construction: every entry today is a file, so the check is doing work on the real list rather than on an empty filter.

**Findings 2 and 3** taken too — the failure message was still carrying the universal the docblock three lines above had just been corrected for (the classic: fix the prose, leave the string), and the docblock said "the rules above" while its own third bullet is about the rule below.

`bun run check` and `bun run check:skills` both exit 0.

---

Round 10 flagged this same negative rule for failing open under a glob, I fixed the glob half, and the directory half survived four more rounds inside the fix. Worth stating plainly: the premise guard has now been wrong twice in the same direction, both times because I enumerated a failure mode instead of asserting the property. It asserts the property now.

PR #750, opened 2026-09-16
fix(ci): the bundle payload was a third door, and widening the token found a fourth

ignaciohermosillacornejo:
# Summary

Follow-up to #746, which merged while its second review round was still being addressed. These are that round's findings — one new hole in `check:tracked-files`, one latent defect widening it exposed, and three smaller ones. No new behaviour beyond closing them.

## External assumptions

None. Both gates read only this repo's files and its git metadata.

## Bug fix?

Root cause:       `check:tracked-files` derived its closure from `package.json` script paths, the import graph, and text under `scripts/`/`.github/` — none of which reaches the files `scripts/pack-mcpb.ts` stages into the shipped bundle. Separately, `PATH_TOKEN` required a directory segment, so root-level script arguments never seeded at all.
Bug class:        `silent-failure-masking` — the same class #729 filed, reached through doors the first pass did not watch. The alternation defect underneath is `silent-under-collecting-scan`: the scan took less than its own docblock claimed, and the shortfall was indistinguishable from a pass.
Detector added:   Three tests in `tests/scripts/check-tracked-files.test.ts`, each a deletion mutant of the guard it covers — a tracked file under a shipped directory that an unanchored rule matches; a root-level `package.json` path; and that same path's extension surviving intact. All three verified red on reverting their fix.
Siblings checked: The other `STAGED_FILES` entries (`icon.png`, `manifest.json`, `CHANGELOG.md`, `PRIVACY.md`, `README.md`, `LICENSE`, `SECURITY.md`) are root-level files with no directory segment; the widened `PATH_TOKEN` does not reach them either, because nothing names them in a `package.json` script. Recorded as a known limit below rather than claimed as covered. `skills/` was the one that was not hypothetical and is now covered.

---

## 1. The shipped bundle's inputs were outside the closure

`scripts/pack-mcpb.ts` stages `skills/` into the `.mcpb`. `skills/` is the only directory besides `scripts/` with a hand-written rule inside it, and the repo-wide **unanchored** rules — `LOG`, `CURRENT`, `LOCK`, `*.log` — match at any depth. So a skill reference file could be ignore-matched exactly the way `tests/unit/manifest-sync.test.ts` was, and nothing would have said so until `pack:mcpb` threw in CI.

`skills/` is now a text-sweep root. Be precise about what that buys:

- **Closes** the landmine half (tracked-but-ignore-matched) for every tracked skill file, at the cost of one array entry. `skills/user-profile.md` is untracked and roots are tracked-only, so it stays quiet.
- **Does not close** the untracked half. A whole-directory dependency like `'skills'` names no individual file for the closure to miss, and handling the directory case as such is a larger change than this follow-up should carry.

## 2. Root-level paths never seeded — and widening the token found a defect of mine

The header claimed "every repo-relative path named in a `package.json` scripts command." `PATH_TOKEN` required a directory segment, so `tsconfig.tests.json` and `tsconfig.scripts.json` — both tracked, both named by `typecheck` — were silently dropped. The docblock was one character wider than the regex.

Widening it exposed a latent defect in my own regex. Alternation is ordered, `js` precedes `json`, and so `tsconfig.tests.json` matched as `tsconfig.tests.js` — a dangling reference to a file that has never existed. The extension now has to end the word, which makes the match independent of the order the alternatives happen to be written in. Worth noting that the *only* reason this surfaced is that the fix for one finding ran the regex over inputs it had never seen; the defect had been sitting behind the directory requirement.

## 3. Three smaller ones

- **`expectToolTable` read every backticked token in the section.** It passes today only because the prose happens to use hyphen-leading flags; a future sentence mentioning `` `amount` `` or `` `merchant` `` would have failed as `names 1 unknown tool(s): amount`, which reads as a registry problem and is a trap for whoever writes that sentence. Scoped to table rows. Set equality — the part #723 is actually about — is unchanged.
- **The case-fold class was unobservable where CI runs.** `MANIFEST-*` → `MANIFEST-[0-9]*` produced seven of #746's eight sibling fixes, but over-matching needs `core.ignoreCase`, which is false on ubuntu-latest's ext4. The fixture now sets it explicitly. Verified first that the **config**, not the filesystem, is what drives exclude matching: same repo, `core.ignorecase=true` → ignored, `false` → not ignored.
- **The ignore-match message hardcoded `.gitignore`** while printing whatever source `-v` reported, so a `.git/info/exclude` hit rendered as `matched by .gitignore (/home/x/.config/git/ignore:3:…)`. Plus a `--` separator on that call, and `CONTRIBUTING.md` still described the pre-`6838ced` closure shape.

## Verification

`bun run check` and `bun run check:skills` both exit 0. Closure went 370 → 381 files, seeds 32 → 34.

Mutations, all run:

| revert | result |
|---|---|
| drop `skills/` from the sweep roots | shipped-directory test goes red |
| require a directory segment in `PATH_TOKEN` | root-level test goes red |
| remove the end-of-word anchor | root-level test goes red, naming `tsconfig.tests.js` |


ignaciohermosillacornejo:
All six fixed in `e0e8abf`. Two notes where acting on the review turned up something it didn't have visibility into.

**2 + 4 — the two `expectToolTable` guards were mutually masking.** Writing the detector you asked for showed that neither guard's deletion is observable while the other stands: delete row-scoping and the registry filter still swallows the prose tokens; drop the registry filter and row-scoping still hides them. Two overlapping defences, each making the other's deletion invisible — a vacuous *pair* rather than a vacuous assertion, and exactly the #596 shape one level up. So they are now separated by what only each can do:

- **row-scoping** stops prose satisfying `missing` on the table's behalf. That is #723's actual property, and it is the stronger of the two: without it, mentioning `get_balance_history` in a sentence lets its row go missing. Detector: row removed, tool named in prose → must still report `omits 1 tool(s)`.
- **the registry filter** stops a backticked non-tool *inside a row* being reported as unknown — `(with merchant filter)` is one edit from that shape, as you noted, and row-scoping cannot help there. Detector: backtick `merchant` in that row → must stay green.

Each goes red on its own mutation now. I took your "filter `extra` to the union of the tool defs" suggestion for the second; it closes the trap rather than narrowing it, so the qualifying sentence you asked for isn't needed any more.

**5 — the punctuation coupling was worse than "could stop covering".** Verified both directions: reword the message to `names ${rel} — which does not exist` and the old `not.toContain('tsconfig.tests.js,')` goes green *even with the alternation bug reintroduced*. The lookahead form survives the reword and still goes red on the bug. Good catch.

**3 — `trimStart`** done, and it needed a detector of its own: the real table isn't indented, so the mutation was invisible. Fixture indents the rows now; reverting to bare `startsWith` reports `omits 14 tool(s)` and the test goes red.

**1, 6 — docblocks.** Both updated, including the `(?![\\w])` over-escaping. `TEXT_SWEEP_ROOTS` now carries the shipped-bundle rationale for `skills/` alongside the invoked-by-name one, with the known limit stated where it will be read: roots are tracked-only, so `skills/` gets the ignore-matched half and not the untracked half. `PATH_TOKEN`'s block records both things you flagged — that the anchor is order-independent but not suffix-independent (`foo.yml.j2`), and that dropping the directory requirement widened the blast radius for `package.json`'s deliberate `_comment_*` prose.

**Smaller notes** both taken: line wrapped, and `CONTRIBUTING.md` no longer writes the roots as globs — they are recursive over every tracked file at any depth, and `.husky/pre-push` has no extension at all.

`bun run check` and `bun run check:skills` exit 0; 3435 tests.

ignaciohermosillacornejo:
All three fixed in `5593d32`, and finding 1 was a regression I introduced — worth naming as such.

**1 — the filter that silenced prose also silenced a typo.** You're right, and the framing is the useful part: "not a tool" covers two different things, and I collapsed them. Prose (`merchant`) is one; a name that *meant* to be a tool — typo, or a rename whose old row survived — is the other, and `extra` caught it before this PR. I broke that while closing the trap.

Took your discriminator. `extra` now reports a real tool in the wrong section **or** anything tool-shaped (`verb_noun`) the registry does not have, and the docblock states what the shape still misses: a misspelling with no underscore. Detector added using exactly your case — typo one of the duplicate `get_transactions` rows, `missing` stays quiet, `extra` must fire.

All three `named`/`extra` behaviours are now independently mutation-verified:

| revert | test that goes red |
|---|---|
| membership-only filter | `reports a tool-shaped name the registry does not have` |
| no filter at all | `a backticked non-tool inside a table row is not an unknown tool` |
| `trimStart` → `startsWith` | `reads an indented table, rather than reporting every tool missing` |

**2 — the hardcoded count.** Corrected, but not to 14. A number in that comment is the same defect the script exists to prevent, one level up, and the next registry change would stale it again; it now states the property ("reports every tool as missing") and notes in passing that it used to name a number and the number was wrong.

**3 — membership vs sweep.** Good distinction and I had genuinely conflated them. Both docblocks now say a root contributes twice — every tracked file under it becomes a closure **member** and a sweep **source** — and that `skills/` is listed for membership, with the sweep half called out as a harmless side effect for the reason you gave: `resolveLiteral` requires a `LITERAL_PREFIXES` hit **and** `isFile`, so narrative naming a nonexistent path is skipped rather than reported. That is the opposite of the `PATH_TOKEN` widening, and worth the contrast. `CONTRIBUTING.md` now separates the two reasons rather than implying the sweep is why `skills/` is listed. Also noted the real reference it pulls in (`scripts/decode-coverage.ts`, from `skills/boundary-audit/SKILL.md`).

**Smaller notes.** Fixture uses `replaceAll('\n|', '\n  |')` now — you were right that `| ` missed the `|---` delimiter row, so it wasn't quite the shape it modelled. On prettier/eslint not covering `scripts/**`: agreed it's the reason nothing will flag the next long line, and agreed it's out of scope here — it would reformat every script in the repo in a PR about two gates. Worth its own issue.

`bun run check` and `bun run check:skills` exit 0; 3436 tests.

ignaciohermosillacornejo:
All three fixed in `6734752`. Findings 1 and 2 turned out to be the same mistake pointing in opposite directions, which is worth saying out loud.

**1 — the dead arm, and what it was hiding.** You're right that `ALL_TOOL_NAMES.has(t)` is subsumed today. The part that matters isn't the dead code though — it's that the *premise* was incidental. A tool named `networth` would have slipped through `extra` in silence: a doc could name a tool that does not exist and the only gate that reads that table would say nothing, which is precisely what `expectToolTable` exists to prevent.

So I took your second option. `tests/tools/registry/tool-name-shape.test.ts` gates the convention, the arm is documented as unreachable-while-it-holds and load-bearing the moment it doesn't, and a single-word tool name now fails *where the convention is broken* rather than at the surface that depends on it. Mutation: rename `get_cache_info` → `cacheinfo` and the invariant test goes red.

**2 — the trap moved, and I'd have missed it.** This is the better catch of the two. The premise "non-tool prose is single words" holds for `merchant` and fails for the whole category it stands in for — 75 argument names in this registry, pervasively snake_case. And your point about which spelling is *likelier* is the convincing part: a doc author reaching for a backtick in a row reaches for the real parameter name at least as often as for an English word. Took your `ALL_ARG_NAMES` derivation verbatim. Detector backticks `account_id` and `start_date` in that row and expects exit 0; dropping the exclusion turns it red.

**3 — the no-op detector.** Fixed, and I swept the file as you asked rather than patching the one instance. Both exit-0 fixtures now go through a `mutate()` helper that throws when the anchor is gone, with the message naming the fixture rather than the gate. The exit-1 fixtures in this file and in `check-tracked-files.test.ts` fail loudly on a no-op, so they are left alone — stated in the helper's docblock so the asymmetry doesn't read as an oversight. Mutation: reword that cell in the real doc and the helper throws by name.

**Smaller note** taken — "every tracked file under it" instead of the file count, which was both wrong (9 tracked, not 7) and the kind of hand-typed number this PR is about.

`bun run check` and `bun run check:skills` exit 0; 3439 tests across 178 files.

ignaciohermosillacornejo:
All five fixed in `6f52d8d`. Finding 1 corrects a claim I made in the previous commit message, which is the useful part.

**1 — "exit-1 fails loudly on a no-op" was wrong, and wrong in the way this PR is about.** It holds for a single-edit fixture. It does not hold when edit A already produces the asserted failure and edit B is the thing under test — the test goes green having silently become a duplicate of another test. Both compound fixtures had that shape, including the row-scoping detector whose discovery I'd been pleased about. The discriminator is **whether any single edit in the chain is load-bearing on its own**, not the exit code; the helper's docblock now says that, and explicitly records that its earlier wording was wrong. Every non-trivial edit routes through it, and it takes a regex. Mutation: reword `With \`--live-reads\`` in the real doc and the fixture throws by name instead of passing.

**2 — nested arguments.** Walk recurses now (`properties` and `items.properties`), 75 → 77 names, `name_contains` among them. I took the recursive option rather than documenting the limit precisely because of your framing: "Derived from the registry, so a new argument needs no edit here" reads as total coverage, and a sentence that *reads* as coverage while not being coverage is worse than no sentence. Detector backticks `name_contains`; flattening the walk turns it red.

**3 — the cast.** You're right that it pointed the wrong way: it widened two *required* fields to optional, so renaming `properties` would have compiled to a silently empty set instead of a type error — under-collection indistinguishable from passing, in the script about exactly that. Dropped; `inputSchema.properties` typechecks directly.

**4 — the hardcoded filename.** Same defect as the `.gitignore` label one, caught the same way. Takes a `file` parameter now.

**5 — `toBeGreaterThan(40)`.** Gone to `> 0`, which is what "non-vacuous" actually means. A floor just under today's count carries no information and fails spuriously the day the surface legitimately shrinks — and I'd replaced a hardcoded `16` two commits earlier for the same reason, so thank you for not letting it pass.

On the prettier/eslint gap for `scripts/**`: agreed it's the root cause of both the long line and the trailing-comma style drift you spotted, and agreed it belongs in its own PR rather than this one — it would reformat every script in the repo inside a PR about two gates.

`bun run check` and `bun run check:skills` exit 0; 3440 tests across 178 files.

ignaciohermosillacornejo:
All four fixed in `1863edd`.

**1 — the charset premise.** Agreed, and its failure mode is the reason to take it seriously: an unmatched name isn't merely unreported, it's **unrecoverable**. `missing` names the tool forever and no edit to the doc can clear it, because the parser cannot see the token that would satisfy it — `bun run check` stays red until someone reads the regex. Gated, and the two premises are now documented as failing differently: relax the charset and the gate jams; relax the underscore and an unknown name goes silently unreported. Mutation: `get_cache_info2` turns it red.

**4 — out-of-section, and I'd rather close it than note it.** You were right that it's narrower than round 4's finding, but the fix is a few lines and it removes the last way that particular test can become a duplicate. `anchoredEdit` proves the doc changed; it can't prove the result is in scope. The fixture now asserts the inserted prose lands between the heading and the section's `---`. Mutation: move that sentence below the rule — exactly the plausible edit you described, since it's a "see the README" pointer — and it throws `landed outside the scanned section` instead of passing.

**2 — "at every depth".** Narrowed to the two keywords actually walked, with `oneOf`/`anyOf`/`allOf`/`patternProperties`/`$defs`/schema-valued `additionalProperties`/tuple `items` recorded as the uncovered routes and "none does today — every `additionalProperties` here is boolean `false`" as the evidence. Taking the clause rather than the code for the same reason you gave: this PR opened by fixing a docblock that claimed "every repo-relative path" over a regex taking a subset, and I'd just written the same shape one file over.

**3 — both overstatements.** `EXAMPLES` is the doc every `anchoredEdit` *names*, not every file the fixtures touch, and the summary sentence was false for five fixtures the docblock's own discriminator correctly exempts. I kept your sharper reason for one of them: `reports every mismatch` edits two files but asserts on each name separately, so either edit going stale fails loudly, which is a different exemption from "single edit" and worth writing down.

**Smaller notes.** `anchoredEdit`'s replaceAll/replace asymmetry documented. Added the visited set — you're right there's no cycle today and the registry only *shares* fragments, but it's two lines and it means a by-reference schema build wouldn't hang.

One I'm leaving: `ALL_ARG_NAMES` shrinking `extra`'s reach by one string per new argument, already noted as a residual. The recursive walk makes it grow slightly faster, which as you say is the right trade.

`bun run check` and `bun run check:skills` exit 0; 3441 tests across 178 files.

ignaciohermosillacornejo:
All three fixed in `9ffffd4`. Finding 1 is the sharpest of the round — it lands on something the *previous* commit added, which makes it the fourth consecutive round where the fix for one class reintroduced it one level over. Worth naming, because that pattern is more interesting than any individual defect here.

**1 — the dedupe guard was scoped to where nothing is reachable twice.** All three sub-points hold and I verified each:

- `seen` was a default parameter, so the loop made a fresh one per tool. Per-tool is exactly the scope with no revisits.
- `BULK_TARGET_PROPERTIES` has one use site and it is a spread, which copies. And `review_transactions` — which that const's own docblock names as the other sharer — duplicates the block inline *above where the const is defined*. The cited evidence was a stale comment; I fixed it at the source too, so the next reader isn't sent looking for sharing that isn't there.
- `TRANSACTION_FIELDS_PARAM_SCHEMA` / `RECURRING_FIELDS_PARAM_SCHEMA` are the real by-reference sharing, across tools.

Took the hoist. Measured rather than asserted: **3 revisits skipped with one hoisted set, 0 with the per-tool default.** So the arm is live now instead of unobservable. Argument count unchanged at 77, which is the expected result — skipping a revisit loses nothing because a parent adds a child's name before recursing.

**2 — the mirrored boundary.** Took the clause, as you suggested for scope. Worth recording *why* extraction isn't available here: the test can't import the script because running it is the check, which is the whole reason these tests spawn a subprocess. The docblock now distinguishes the two halves — a heading divergence throws, a boundary divergence doesn't — so whoever changes `^---$` in the script sees what else moves.

**3 — the enumeration.** Gone. The clause up to "not every file the fixtures touch" is exact and can't stale; the list was the part that could, and you're right that I'd removed two of those in the preceding two commits for the same reason.

**Smaller note on the argument charset** taken as a clause next to the derivation: the exclusion set is only ever compared against tokens harvested by the same `/`([a-z_]+)`/`, so an off-charset argument yields no token rather than a spurious one — benign by construction, not by luck. That was exactly the question the new tool-name invariant invites, and leaving it unanswered would have read as an oversight.

Agreed on keeping the charset and underscore tests separate, and on `scripts/**` lint/format being its own PR.

`bun run check` and `bun run check:skills` exit 0; 3441 tests across 178 files.

ignaciohermosillacornejo:
Took the delete, in `7e5c9c8`.

**1 — you're right that hoisting bought liveness but not observability.** I verified the leaf shape rather than taking it: each of the three revisited nodes contributes **zero** names. Which means the previous commit's "argument count unchanged at 77", which I offered as evidence nothing was lost, was equally the evidence that nothing was gained. That's the sentence I should have read both ways when I wrote it.

Deleting rather than documenting, because findings 2–4 are the tell. Two rounds of justification had grown three dependent problems — the `= new Set()` default was itself the per-tool scoping the last commit diagnosed as the bug, one forgotten argument from reinstating it with nothing able to detect it; the soundness argument covered a revisited node's own key but not its subtree, which survives only because every call shares one `into`; and the one claim worth testing is unreachable from a test for the same reason `assertInToolSection` is. A mechanism needing that much scaffolding to be defensible, while its deletion stays byte-identical, is one to remove. All three evaporate with it.

What's left is the sentence you drafted: the sharing is real but leaf-shaped, so a revisit costs one empty loop and can lose nothing — plus a note to add a guard if a shared fragment ever carries `properties`, and why such a guard is unobservable by construction.

Worth saying plainly, since it's now four consecutive rounds: the recurring pattern here was mine, not the code's. Each round's fix reintroduced the class one level over — a guard whose deletion nothing notices — and the `seen` set was the purest instance, since it was added in response to a *smaller note* and then defended twice. The right move on a mechanism that cannot be observed is to not ship it.

`bun run check` and `bun run check:skills` exit 0; 3441 tests across 178 files.

ignaciohermosillacornejo:
All three fixed in `1e6cc4e`. One rewrite of the paragraph, since all three were in it.

**1 — the correctness/cost confusion, and the hazard pointing the wrong way.** You're right on both halves and the second is the one that mattered: omitting the set is correct under *any* schema shape (a re-walk collects the same names into the same `into`), so leaf-ness bounds the cost, not the correctness — and my parting instruction recommended re-adding a set while the caveat that makes one dangerous had just been deleted along with the paragraph that held it. The docblock now says a future set would be a cost optimisation and never a correctness fix, and carries the `into`-sharing condition with the recommendation rather than a commit message ago.

**2 — three counted, two named.** `ACCOUNT_FIELDS_PARAM_SCHEMA`. Replaced with the property, as you suggested — and you're right that it's the fourth time in this PR: the hardcoded `16`, the four-file `EXAMPLES` list, the skill-file count, now this. Four instances of the same reflex in one PR about hand-typed facts is its own finding.

**3 — the measurement described the deleted program.** Correct, and the direction matters: "three" was taken with the guard short-circuiting before descent, so without it each fragment *and* its `items` child repeats. Immaterial to the conclusion, but "measured" was doing persuasive work in a docblock whose subject is claims wider than their code, and this one was narrower. The number is gone; the shape claim ("the fragment and its single `items` child, neither contributing a name") is exact and doesn't need one.

`bun run check` and `bun run check:skills` exit 0; 3441 tests across 178 files.

ignaciohermosillacornejo:
All three fixed in `eabe38a`.

**1 — the hand-rebuilt union.** Good catch, and the asymmetry you drew is the reason it's worth more than a tidy-up: the script's divergence would fail *loudly* (a tool missing from its local copy surfaces as a false-positive "unknown tool"), while the invariant test's would fail *silently* — it would simply stop gating the charset and underscore premises for the new tools while staying green. That's the silent-under-collecting-scan class, sitting in the test whose whole job is preventing the unrecoverable charset jam. Both import `ALL_TOOL_DEFS` now, and the script keeps the three arrays only for the per-mode counts that genuinely need them.

That is the fifth hand-maintained fact in this PR, after the hardcoded `16`, the `EXAMPLES` enumeration, the skill-file count and the sharer list. Same reflex every time, which at this point is the more useful finding than any of the five.

**2 — tree vs DAG.** You're right, and the sentence was self-refuting: by-reference sharing is exactly what makes these a DAG, and is why a visited set was on the table at all. Acyclicity doesn't follow from the shape either — it follows from `const` object literals initialised in module order, so a fragment can only embed one already defined. That's what the docblock says now.

**3 — the hardcoded file.** Fixed. Mildly embarrassing: `anchoredEdit` two functions up got that exact parameter one round earlier for that exact reason.

`bun run check` and `bun run check:skills` exit 0; 3441 tests across 178 files.

ignaciohermosillacornejo:
All four fixed in `0826c72`. Finding 1 is the one worth the round: I fixed the instance three commits ago and not the class, which is exactly what CONTRIBUTING's bug-response ritual exists to catch, and the sibling was two files away.

**1 — the other collector.** `scripts/dump-tool-args.ts` had the same one-level read under a docblock promising "every dispatchable MCP tool's INPUT ARGUMENT names". Both recurse now; `update_recurring` gains `name_contains` and the nested `rule` members. Your read on the direction is what makes it worth fixing rather than noting: it's an *exclusion* set, so a missing nested name becomes a false positive telling a skill author to add a `fields:` argument for a parameter. And you're right the absence of an instance is accidental — the one skill naming it writes `` `name_contains: "…"` ``, which the identifier regex doesn't match.

**2 — the cast, and the guard that couldn't see the shape.** The cast is the round-4 one I'd fixed in the sibling and missed here. The downstream half is the better catch: `collect_tool_args` refuses an empty *map*, but a broken collector produces a non-empty map of *empty lists*, which passes shape validation and the emptiness check and then silently turns every argument name in the repo into a candidate row field. Added `if not any(args.values())` with a message that names the degenerate shape, and mutation-tested it — point the collector at nothing and `check:skills` now fails with "50 tools and not one argument between them" instead of validating.

**3 — the parameterized axis.** Took the clause. `file` labels the messages; the heading and boundary stay module-global, so the function is EXAMPLES-specific by construction and passing another file gets you a loud wrong-file message rather than a silent pass. Worth saying because the signature invites the opposite reading.

**4 — the comment on the wrong line.** Alias deleted; the three tests read `ALL_TOOL_DEFS` directly, so the warning sits on the name a future rewrite would touch instead of one hop above it.

`bun run check` and `bun run check:skills` exit 0; 3441 tests across 178 files.

ignaciohermosillacornejo:
All four fixed in `372890e`. Took the extraction rather than the back-reference.

**1 — the duplication was the bug.** You're right that the "cannot extract" argument doesn't transfer: neither script can import the other, but a third module can hold the walk, and `scripts/manifest-utils.ts` is the precedent. `scripts/schema-args.ts` now holds the walk, the `SchemaNode` shape and the known-limit paragraph once.

The part that settled it: **extraction and testability are the same change.** That walk had never been directly tested because running either script IS its check — the same constraint I cited for `assertInToolSection` — and moving it into a side-effect-free module dissolves that. `tests/scripts/schema-args.test.ts` asserts the contract on synthetic schemas plus `update_recurring.rule.name_contains` against the real registry, with the top-level siblings asserted alongside so they can't mask a regression. Mutation: flatten the walk, it goes red on the nested case.

A nice incidental: `check:tracked-files` caught `scripts/schema-args.ts` before I committed it — the new file was un-allowlisted, the gate named it, `git add -A` had skipped it silently. That is #729's exact scenario, on the PR that fixed it.

**2 — the detector for the refusal.** Added to the class-level table as one more case, using the `argsBody` parameter that was already there. Mutation-verified: revert `not any(args.values())` and it goes red.

**3 — the silent false negative.** Recorded in the shared module's docblock, with your numbers: the global delta of two names understates it because the union is per-tool, and `update_transactions` goes from two to thirteen through `edits[].items.properties` — five of them off-default row fields, including the `tag_ids` that motivated narrowing the union in the first place. The trade matches the one the top-level exclusion already made, but the loud/silent asymmetry gets written down, per this PR's standard.

**4 — total vs partial.** Fair, and the honest framing is that a collector's correctness isn't observable from its output shape at all. The Python guard now says it catches the total break only, and points at `tests/scripts/schema-args.test.ts` for the partial one — which is now a real place to point, because of finding 1.

`bun run check` and `bun run check:skills` exit 0; 3449 tests across 179 files.

PR #753, opened 2026-09-17
fix(ci): ratchet the schema walk, and pin each lookup case to its own failure

ignaciohermosillacornejo:
# Summary

Final follow-up to #746 / #750, closing that review thread's last round. Three findings, no new surface — a ratchet for the class #750 fixed by instance, a detector-table trap, and a known-limit clause on the one sibling the sweep missed.

## External assumptions

None. Everything here reads this repo's own files and its tool registry.

## Bug fix?

Root cause:       #750 removed the two hand-rolled JSON-Schema argument walks into `scripts/schema-args.ts` but added nothing to stop a third, and `makeRepo` in the skills detector table created its synthetic `scripts/` directory only when `dumpBody` was set, so a case supplying `argsBody` alone silently ran a different case.
Bug class:        `silent-under-collecting-scan` for the walk (a one-level read is indistinguishable from a passing one), `vacuous-assertion` for the table case that degraded into a duplicate of another row.
Detector added:   `tests/scripts/no-duplicate-schema-walks.test.ts` — any file under `scripts/` reading `inputSchema.properties` must import the shared module, no allowlist; modelled on `tests/no-hand-rolled-comment-strippers.test.ts`. Plus an optional `because` per detector-table case, so a case pins itself to ITS failure rather than to invariants every row shares.
Siblings checked: `tests/conformance/ledger.test.ts` holds a third walk of the same shape. Deliberately left separate — it builds dotted parameter paths and enum sets, so folding it in would couple a coverage gate to a name-collection helper — with the shared limits recorded in its docblock and cross-referenced both ways. Nothing else under `scripts/` reads `inputSchema.properties` (the ratchet asserts this).

---

## Why a ratchet rather than "the copies are gone"

The class is unusually well evidenced here: the one-level-read defect was found in `check-tool-counts.ts`, fixed there, and survived **seven commits of review** in `dump-tool-args.ts`, because fixing an instance does nothing about the copy nobody is looking at. CONTRIBUTING's bug-response ritual says that explicitly, and this repo already has the pattern — `no-hand-rolled-comment-strippers.test.ts` opens with *"Fixing the four instances is not a class-level fix; nothing stops the fifth."*

The expectation is derived, not a text match on the walk's body: a fresh hand-rolled walk has to start by reading `inputSchema.properties`, so that is what is checked. **No allowlist** — `schema-args.ts` is excluded because it *is* the walk, and zero files need an exemption today.

## The table case that tested something else

Every row in the tool-lookup table asserts the same four invariants (exit 1, `linter self-check`, `Skill references were NOT validated`, not blaming a skill). That is what made the degradation invisible: an args-only case with no `scripts/` directory produces exactly those four. Fixed by gating on either body — and, because identical invariants were the enabling condition, by letting a case name its own failure.

## Mutation results

| revert | result |
|---|---|
| add a third hand-rolled walk under `scripts/` | ratchet fails, naming the file |
| a consumer drops the shared import | ratchet fails, naming the consumer |
| revert `check-skills.py`'s degeneracy guard | args case goes red |
| revert the `makeRepo` gate | args case goes red — it did **not** before this PR |

## Cheapest way to make each guard pass wrongly

- **The ratchet**: write the walk without the string `inputSchema.properties` — destructure it, or reach the schema through a variable. That is the honest limit of a source-level check, and it is the same one `no-hand-rolled-comment-strippers.test.ts` lives with. What it does close is the accidental path, which is how both existing copies were written. The non-vacuity test guards the other cheap green (a sweep that finds no files).
- **`because`**: omit it, and the case falls back to the shared invariants. Deliberate — most rows have no distinctive message — so it narrows rather than closes. The rows where degradation is plausible are the ones that set it.


ignaciohermosillacornejo:
All four fixed in `9676c2b` (branch also rebased onto the three dependency bumps that landed since). Findings 1 and 2 were correctness defects in the ratchet, not polish.

**1 — depth.** Right, and the consequence is the part that stings: `tsFilesUnder` recurses, `scripts/` has 26 TS files in subdirectories, and one of them importing the shared module correctly writes `'../schema-args.js'`. The guard would have reported it as an offender **for importing the shared module** — and with no allowlist by design, its only ways out would have been an exemption or hand-rolling the walk. A guard pushing toward the thing it exists to prevent. Depth-tolerant regex now; mutation-tested with a real `scripts/smoke/` consumer.

**2 — the accidental allowlist.** In a file whose docblock opens "THERE IS NO ALLOWLIST", which makes it worse than the usual version of this. Absolute path and strict equality. Mutation: a decoy `scripts/smoke/schema-args.ts` that hand-rolls the walk is now reported instead of exempting itself.

**3 — the untested trigger.** Took the `isOffender` extraction. You're right that the non-vacuity test checked the file list rather than whether the predicate can fire, and that mistyping the literal switches the ratchet off permanently with everything green. Four assertions now: fires on a hand-rolled walk, does not fire with the shared import at either depth, does not fire on an unrelated file.

**4 — `because` on one row.** I was wrong that most rows have no distinctive message; your table is correct and all five are now set. Also took the related gap — `_run_dump_script`'s "is missing" branch had only ever run for the NAMES dump, since case 1 removes the whole `scripts/` directory and the names lookup raises first. An `omitArgs` flag covers the args dump.

One thing worth flagging, found while making `because` required: **`tests/scripts/` is outside every tsconfig program here.** `tsconfig.tests.json` names its files individually and neither `check-skills.test.ts` nor the new ratchet is among them, so `because: string` documents intent and enforces nothing — I verified that deleting a `because` typechecks clean. It is guarded at runtime instead, with a message naming the row. That is the #725 shape your own tsconfig docblock warns about, one directory over; I've left it out of this PR rather than widen it, since adding entries interacts with `tsconfig-tests-sync.test.ts`'s membership rules.

`bun run check` and `bun run check:skills` exit 0; 3465 tests across 180 files.

ignaciohermosillacornejo:
All three fixed in `6f21fac`, plus round 2's two open items.

**1 — the exemption excused the two files the ratchet exists for.** This is the best finding on the PR and I'd missed the shape entirely. The exemption was file-scoped, not use-scoped, so an inline walk re-added to `dump-tool-args.ts` *alongside* its existing import was invisible — and that is the file which carried this exact defect through seven commits of review. Your framing is the part worth keeping: the class as the docblock stated it is "a third copy in a new file"; the class as this repo has actually experienced it is "a copy in a file that also legitimately touches schemas".

Took the deletion. It protected nothing (going through the shared module is what stops a consumer containing the literal), it is strictly stronger, and it takes round 2's comment hole with it — with nothing exempting, `IMPORTS_SHARED` matching inside a comment can no longer excuse anything, so the #691 stripper dependency isn't needed. One exclusion left, by absolute-path identity. Mutation: an inline walk beside the shared import is now reported by name.

**Round 2 #2 — descent.** Also taken: `length > 0` plus the module itself is satisfied by the top-level files alone, so the 26 subdirectory files could silently leave scope, and `ts-files.test.ts` covers extensions and ScriptKind rather than whether a caller's tree is descended.

**2 — the shared `because`.** Right, and it is the one shape the mechanism exists to rule out. `check-skills.py` raises that sentence for both dumps; qualified with the script name. Thank you for verifying the other six against the source rather than assuming.

**3 — the stale enumeration.** Took the first resolution rather than amending the prose, since your own header argues for it: the ratchet is on the include list and pinned in `MEMBERSHIP_IS_THE_COVERAGE`. Bonus exactly as you predicted — `isOffender`'s types are now read by something. Mutation: dropping the include line reddens the membership pin.

The test is renamed too — it asserted "every script reading the literal imports the shared walk", and the rule is now "no script but the module reads it at all", so the old name described a guard that no longer exists.

`bun run check` and `bun run check:skills` exit 0; 3465 tests across 180 files.

ignaciohermosillacornejo:
All four fixed in `ba4f241`.

**1 — the vacuous vacuity-closer.** Caught me cleanly. `startsWith(join(SCRIPTS_DIR, 'smoke'))` is satisfied by the top-level `scripts/smoke-graphql.ts`, so the line proving the sweep descends passed without the sweep descending. Measured by path separator now. Mutation: neuter `tsFilesUnder`'s recursion and it goes red — which it did not before.

**2 — the only realized hit is a comment.** This one reframed the exclusion for me. `scripts/schema-args.ts` names the literal in its *docblock*; its code reads `node.properties` and never writes it. So the identity exclusion has been suppressing a prose mention, not the shared walk, and the sweep's one lifetime match was a false positive. Stripping added, and you're right it is now purely a false-positive reducer — with nothing exempting, a walk cannot hide inside a comment. The concrete consequence is in the docblock: a new script documenting "uses schemaArgNames() rather than reading inputSchema.properties" would have been reported for saying so, with the shared module's own docblock as proof that the prose is natural to write.

**3 — 26 vs 28.** Right three rounds running, and I propagated it into a second docblock last commit. Deleted rather than corrected: finding 1's assertion measures it, so no number is stated anywhere.

**4 — the missing remedy.** Taken, modelled on `no-hand-rolled-comment-strippers.test.ts`. Your point about the load-bearing clause is the one that mattered: for a detector whose docblock spends eleven lines arguing there is no allowlist, a failure reading `expected [] received ['smoke/foo.ts']` invites precisely the fix those lines argue against. The message names the helper, says why a one-level read under-collects, and states that an exception is argued in the file rather than appended to a list.

`bun run check` and `bun run check:skills` exit 0; 3465 tests across 180 files.

ignaciohermosillacornejo:
All four fixed in `7e4ff49`.

**1 — the enumeration, again, three lines up.** Fair and slightly galling: adding `stripComments` made this a fourth importer of `strip-comments.ts` while that helper's list still said three — the same defect, same comment block, three lines above the `ts-files.ts` one I fixed two commits ago. Corrected, with the distinction you'd want: unlike the two scanners that stay off the list, the ratchet IS on it and so is actually checked.

That is the seventh hand-maintained fact this PR chain has corrected, and the second one I introduced *while* correcting another. The honest reading is that prose enumerations of a derived set are a defect generator, and this repo keeps paying for them one line at a time.

**2 — the surviving 26.** Deleted, and the paragraph around it was worse than stale: it argued for a behaviour the offender predicate no longer has. `IMPORTS_SHARED`'s docblock now says what it is actually for — the "both known consumers" test, and deliberately *not* the predicate.

**3 — the docblock contradicting its caller.** Right. `isOffender` takes whatever string it is given; the caller strips. Said that way instead, including why the predicate test's raw synthetic strings are equivalent (no comments in them).

**4 — the `> 0` floor.** Good catch and the fix is better than a bigger number: the assertion now names the set of subdirectories actually reached, so one tree leaving scope is reported by name rather than covered by the other. Mutation: exclude `graphql-capture` from the walker and it fails naming it; the old floor stayed green.

`bun run check` and `bun run check:skills` exit 0; 3465 tests across 180 files.

PR #754, opened 2026-09-17
feat(check): a registered safety invariant has to prove something notices its removal

### LINKED ISSUE #596: Mutation-guard registry: make safety invariants prove they have detectors

ignaciohermosillacornejo:
# Summary

`bun run check:mutation-guards` — a ledger of the safety invariants whose violation writes wrong data to a real account, each paired with the exact edit that disables it and the one test file that must go red when it does. The runner applies the edit, runs that file, and requires **both** directions: green unmutated, red mutated. A guard whose removal leaves its detector green is reported `VACUOUS`. Closes #596.

## External assumptions

None. Nothing here talks to Copilot's API, adds an operation, or reads a response shape — it edits local source files and runs local test files. No ledger entries touched.

## Bug fix?

Root cause:       Two assertions in #587 were predicates satisfied by both the correct and the broken behaviour (`\d+/8 succeeded` matches a stopped batch and a runaway one alike), so the guards they covered had no detector. "Mutation-tested" existed only as an unverifiable claim in commit messages, and one such claim was false.
Bug class:        `vacuous-assertion` (docs/bugs/README.md) — a test executes a guard but would still pass if the guard were deleted.
Detector added:   `bun run check:mutation-guards` (`scripts/mutation-guards.ts`), in `bun run check` and as a step in `.github/workflows/test.yml`. For each registered invariant it asserts the named test file passes unmutated and fails mutated; the class-level property is "a registered guard cannot be vacuous", not "this one guard is tested". `tests/scripts/mutation-guards.test.ts` mutation-tests the gate itself against synthetic repositories.
Siblings checked: All five guards proposed in the issue were re-derived against current source (three were unusable as written — see below), the `skipped`-rows entry was found to name two distinct guards and both are registered, and the second `stopOnError` check was probed and deliberately left out because it genuinely has no detector.
Ledger updated:   n/a — not an external-assumption bug. `docs/bugs/596-vacuous-assertions-bulk-edit.md` and the class table in `docs/bugs/README.md` now name this gate instead of "none — proposed registry in #596".

## The seed set was re-derived, not copied

Three of the five entries proposed in #596 violate the issue's own "matches exactly once" rule or do not disable what they claim. Verified against current source:

| Proposed | Problem |
|---|---|
| `if (stopOnError && failures.length > 0) return;` | matches **twice** in `src/tools/tools.ts` (the pre-task check and the in-catch one) |
| `if (result.skipped.length > 0) {` | matches **twice** — these are two different guards, `bulk_edit_transactions` and `review_transactions` |
| `bulkEditTransactions: refusing to send without explicit target ids` | a fragment of a `+` concatenation; removing it leaves the `throw` standing, so the entry would report a **false failure** |

The fix is the site marker: every guard carries a `// mutation-guard: <name>` comment at its site and each row's `find` must contain it, so an entry names one site by identity rather than by resemblance. Six rows now, each matching exactly once:

| Guard | File | Detector |
|---|---|---|
| update_transactions stops on first failure | `src/tools/tools.ts` | `tests/tools/update-transactions-batching.test.ts` |
| bulk write refuses an unbounded row set | `src/core/graphql/transactions.ts` | `tests/tools/bulk-edit-transactions.test.ts` |
| bulk filter carries only ids | `src/core/graphql/transactions.ts` | `tests/tools/bulk-edit-transactions.test.ts` |
| bulk_edit_transactions surfaces silently-skipped rows | `src/tools/tools.ts` | `tests/tools/bulk-edit-transactions.test.ts` |
| review_transactions surfaces silently-skipped rows | `src/tools/tools.ts` | `tests/tools/review-transactions-batching.test.ts` |
| bulk edit validates category ids client-side | `src/tools/tools.ts` | `tests/tools/bulk-edit-transactions.test.ts` |

**One guard is deliberately absent.** The *second* `stopOnError` check — the one inside the `catch` — was registered experimentally and the gate reported it `VACUOUS`, correctly: the error the caller sees is built from `failures[0]` either way, so removing it changes nothing observable. It is defence in depth, not an invariant with a detector, and registering it would mean registering a row that can only ever be green by accident. Transcript below.

## Making the gate itself non-fakeable

This gate's whole purpose is to catch guards that execute but cannot fail, so it must not have that defect.

1. **What the runner does.** Per row: run the named test file unmutated (must pass) → apply the edit → run it again (must fail) → restore. Both directions, because the repo's standing rule is both directions.
2. **A mutation must fail for a *behavioural* reason.** "The test file went red" is also what a `find` string that breaks the parse looks like — bun reports that as `Ran 1 test … 1 error` rather than as the file's real count. So the mutated run must execute the **same number of tests** as the baseline **and** report **no module-level error**. Without this, any row could be made green by choosing a `find` that breaks the parse, and the guard would never be exercised at all.
3. **Not trivially satisfiable.** The unmutated direction is checked first and must be green, so `expect(1).toBe(2)` cannot be registered as proof of anything. A `find` that matches zero times is an error, not a skip — a row that has quietly stopped matching is worse than no row, because it reports green.
3b. **The row names its own detector, not just a file.** `expectFails` alone only asserts that the *file* went red, and four of the six rows share `bulk-edit-transactions.test.ts` — so a row could keep printing a tick on a sibling test's failure while its own coverage rotted. Each row also names the test(s) that must be among the failures, checked statically against the detector's source (a renamed test fails the gate rather than silently unpinning it) and again at run time.
3c. **What this does NOT prove, stated outright.** For a `replace` row the `with` string is arbitrary, so what is established is "*some* edit at this marked site is detected", not "disabling this guard is detected". Nothing can check that a replacement is a faithful disabling. The mitigation is reporting rather than enforcement: every passing row prints the names of the tests that caught it, so a reviewer can confirm by eye that the detection is the one the row claims.
4. **Cheapest false green: deleting a row.** Stated plainly — it is deleting a row from `MUTATION_GUARDS`. That path **is** gated: the guarded site carries a `// mutation-guard: <name>` marker, and the gate requires a bijection between markers in `src/` and rows in the registry, so a dropped row leaves an orphaned marker and fails with a message naming it. This does not make removal impossible (nothing can), but it turns one silent deletion into a deliberate two-file edit a reviewer sees. The next-cheapest is deleting the marker *and* the row together — at which point the guard code itself is unprotected, and the only remaining backstop is that the guard's own detector still exists. Transcript below.
5. **The runner cannot hang holding the mutation.** The child test run is bounded (120s, SIGKILL). Several rows delete an early exit — a `return`, a `break` — which is exactly the edit class that turns a bounded loop unbounded, and bun's per-test timeout does not cover a module-scope hang. A timeout is reported as "the mutation hung the detector", never as a detection.
6. **Restoration.** Four layers: `finally`; `process.on('exit')` plus SIGINT/SIGTERM/SIGHUP handlers; an on-disk journal written *before* the file is touched, which the next run recovers from (this is the SIGKILL / power-cut case, where no handler runs); and content-addressed write-back — the original bytes go back only if what is on disk is still exactly the bytes this runner wrote, otherwise the original is dropped at `<file>.mutation-guard-original` and the run fails rather than clobbering someone's concurrent edit. A journal owned by a live pid is refused instead of raced.
   The journal is also read as **untrusted input**: it sits at a predictable path under `tmpdir()`, which on Linux is the shared `/tmp`, so obeying it verbatim would make "can create a file in /tmp" into "can write any file, as you, on your next `bun run check`". Entries must carry the three string fields and must resolve *inside* the run root — this runner only ever mutates files under the root it was given, so anything outside it is definitionally not ours.

   *Deviation from the issue, deliberate:* #596 asked for a clean-working-tree requirement and `git diff --quiet` verification. This does neither. `bun run check` is the pre-push hook and is routinely run mid-edit, so a clean-tree requirement would skip the gate exactly when someone is changing the code it guards; and byte comparison against the captured original is strictly stronger than `git diff --quiet`, since it holds for a file that was already modified before the run.
7. **Where it runs.** In `bun run check` (and so the pre-push hook) **and** as its own step in `.github/workflows/test.yml`. It is affordable there — one test file per row, baselines shared across rows naming the same file: **1.4s for all six**. It sits immediately before `bun test --bail` in the composite, so the full suite runs afterwards over the restored tree; if restoration were ever wrong, the very next command in `check` is a whole-repo test run over the damage. That backstop is local only — the CI `quality` job never runs `bun test` — so CI adds `git diff --exit-code` right after the step instead, which is free there because CI checks out clean and is an independent check rather than the runner's own bookkeeping.

## Demonstration

Catching a guard with no detector — the real second `stopOnError` check, registered temporarily:

```
  ✓ update_transactions stops on first failure
      tests/tools/update-transactions-batching.test.ts: 24 pass unmutated → 1 fail mutated
  ✗ only the first failure is recorded
      VACUOUS — disabling this guard left tests/tools/update-transactions-batching.test.ts
      entirely green (24 pass, 0 fail). Nothing detects its removal. Write a test that pins
      the PROPERTY ("Under stopOnError only the first failure is kept"), not the error message.
  ✓ bulk write refuses an unbounded row set
  ...
mutation-guards: 1 of 7 failed.   [exit 1]
```

Deleting a row while its marker stays:

```
  ✗ (registry)
      src/ carries a marker with no registry entry: "bulk edit validates category ids
      client-side" (src/tools/tools.ts). Either add the entry to MUTATION_GUARDS or delete
      the marker — an unregistered marker is a safety claim nothing verifies.

mutation-guards: 1 of 1 failed.   [exit 1]
```

The real registry, green (every passing row names what caught it, which is what makes a `replace` row reviewable):

```
  ✓ update_transactions stops on first failure
      tests/tools/update-transactions-batching.test.ts: 24 pass unmutated → 1 fail mutated
      caught by: update_transactions partial-failure accounting > default: stops — entries
                 queued behind the failure are never written
  ✓ bulk write refuses an unbounded row set
      tests/tools/bulk-edit-transactions.test.ts: 24 pass unmutated → 3 fail mutated
      caught by: bulkEditTransactions() refuses an unbounded write > throws on an empty array
                 rather than sending a filterless request | … | …
  ✓ bulk filter carries only ids
      tests/tools/bulk-edit-transactions.test.ts: 24 pass unmutated → 2 fail mutated
      caught by: bulk_edit_transactions sends one scoped request > filter carries ONLY ids —
                 never a widening key | …
  ✓ bulk_edit_transactions surfaces silently-skipped rows
      tests/tools/bulk-edit-transactions.test.ts: 24 pass unmutated → 1 fail mutated
      caught by: bulk_edit_transactions surfaces partial application > a silently-skipped id
                 fails the call
  ✓ review_transactions surfaces silently-skipped rows
      tests/tools/review-transactions-batching.test.ts: 16 pass unmutated → 1 fail mutated
      caught by: review_transactions surfaces server-side partial application > a
                 silently-skipped id fails the call instead of reporting success
  ✓ bulk edit validates category ids client-side
      tests/tools/bulk-edit-transactions.test.ts: 24 pass unmutated → 1 fail mutated
      caught by: bulk_edit_transactions validates before writing > unknown category is
                 rejected client-side — the server would persist it

mutation-guards: 6 safety invariant(s) each proved to have a detector — every one passes
unmutated and fails mutated.   [2.4s, exit 0]
```

Both failure demonstrations are permanent, not one-off: `tests/scripts/mutation-guards.test.ts` rebuilds each of them against synthetic throwaway repositories, so the gate re-proves in CI that it can still catch a vacuous guard, a red-either-way detector, a parse-breaking mutation, a mutation that hangs the detector, a row riding on a sibling test, a row naming a test that does not exist, a deleted row, a drifted `find`, and an entry that omits its marker — plus a planted journal pointing outside the run root, and a malformed one.

## Verification

- `bun run check` → exit 0 (3,466 pass, 0 fail, 180 files), after rebase onto `origin/main`
- `bun run check:skills` → exit 0
- `git status` clean after every runner invocation; no journal left under the temp dir
- No real financial figures, balances or account names anywhere in the diff


ignaciohermosillacornejo:
Review addressed in 6ed2791. Taking them in the order you gated on.

**1 — no timeout.** Fixed, and you were right that the edit *class* is what makes it more than hygiene. `TEST_TIMEOUT_MS = 120_000` with `killSignal: 'SIGKILL'`, plumbed through `RunOptions.timeoutMs` so the path is testable at 3s. I did not lean on the existing `ran !== baseline.ran` branch — a timeout gets its own branch and its own message ("the mutation hung the detector, not a detection"), because conflating it with an unparseable mutation would send the reader to the wrong fix. New test drives a synthetic repo whose only loop exit *is* the guard, and asserts the source file is back afterwards.

**2 — unvalidated journal.** Fixed as described. `validEntries()` requires the three string fields and requires `resolve(abs)` to be inside the run root; rejected entries are reported, not silently dropped. Two tests: a planted journal naming a file outside the root (which stays untouched) carrying a `{}` entry alongside it, and a truncated journal — the latter is discarded rather than thrown on, since a gate that dies on every `bun run check` until someone hand-deletes a file in `/tmp` would be a worse failure than the one it is recovering from.

**3 — "0 of 6 failed".** Already fixed locally before your review landed, by the route you suggested: a `(restore)` result row, so the count and the exit code agree and the last line names the sidecar.

**4 — `fail > 0` doesn't pin which test.** Took the stronger version rather than reporting alone. `expectFailingTests` is required on every row, checked twice: statically against the detector file's source, and at run time against bun's `(fail)` names. The static half earned its place immediately — it rejected my first attempt at the unbounded-write row, which pinned three `test.each`-rendered names that appear nowhere in the file. Passing rows also print `caught by: …`, and the docblock now has a **"What this gate does NOT prove"** section stating the `replace`/`with` arbitrariness outright: what is established is "some edit at this marked site is detected", the printed names are what make that reviewable, and nothing can check a replacement is faithful.

**Smaller notes.** `--guard` with no value now exits 2. CI gained `git diff --exit-code` after the step, with a comment recording your point that the `quality` job never runs `bun test`, so the docblock's "next command is a whole-repo test run" backstop is local-only. Signal handlers now exit 128+signo per signal instead of 130 for all three, and journal entries carry their own `root` so a multi-root process clears the right journal. The `does not PASS unmutated` branch now prints a tail of the child's output — agreed that it is the most likely contributor-facing failure, and agreed it does not belong on the mutated branch.

**The tsconfig one I did take.** You filed it as follow-up and consistent with its 26 siblings, but the argument you quoted cuts hardest exactly here: it is the sole verification of a gate that edits tracked source in place. `tests/scripts/mutation-guards.test.ts` is now on `tsconfig.tests.json`'s include list and pinned in `MEMBERSHIP_IS_THE_COVERAGE`, so the line cannot be dropped silently either.

`bun run check` exit 0 (3,466 pass), `bun run check:skills` exit 0.

ignaciohermosillacornejo:
Both new findings fixed in d73e729, plus the three smaller notes.

**The `restoreFailed` loop continuation.** You traced it exactly right, including which of the two outcomes is worse. Fixed both ways you suggested: `if (restoreFailed) break;` at the top of the loop, and the per-row body extracted into an `evaluate()` whose call is wrapped, so an unexpected throw becomes a failing row instead of taking every computed result — and the `(restore)` row — with it. Two rows over two files in one synthetic root, the first detector clobbering its own mutated source: the named row appears and the second row never runs.

**The untested `(restore)` row.** Agreed, and agreed on why it mattered here specifically. Covered now, by exactly the shape you described. The throw path got one too, and through a reachable route rather than a contrived one — a detector that rewrites its source during the *baseline* run leaves `find` absent by the time `applyMutation` reads the file, which is how I found the branch was live in the first place.

**CI post-condition skipped when it matters.** Right, and the sidecar point too. `if: always()`, and `git status --porcelain` rather than `git diff`, with both reasons in the comment.

**`res.signal === 'SIGKILL'` broader than "timed out".** `killedBy: 'timeout' | 'signal' | null`; the OR stays as the fail-safe, but an OOM no longer gets told it ran for 120s.

**Docblock drift.** Fixed — the restoration section now names SIGHUP.

**Untrusted input echoed raw.** Fixed. `safe()` strips control characters from `journal.startedAt` and `entry.abs` before either reaches a terminal. Marginal as you say, but the payoff is ANSI escapes rewriting the very message that reports the rejection, which would undo the reporting — so it is the one place the omission actually bites.

`bun run check` exit 0 (3,468 pass), `bun run check:skills` exit 0, tree clean after every runner invocation.

ignaciohermosillacornejo:
All of it fixed in 9f7c3e6, including the whole carried-over list. Thank you for re-listing it — you were right that it would otherwise have been lost to scroll.

**1 — the `break` made skipped rows invisible.** You're right that this is the same shape as the bug the file exists to prevent, which is the argument that made it worth fixing rather than noting. There's now a `(skipped)` row naming every guard that never ran ("they are neither proved nor disproved"), and the refused-restore test asserts it names the second guard rather than only asserting the silence.

**2 — `killedBy` only read on the mutated run.** Fixed, with its own branch before the red-either-way one and its own message ("this is the detector or the machine, not the guard"). And you spotted the live flake: `baselineTimeoutMs` is now separate, so the hang test shortens only the mutated bound and a cold runner cannot fail it with a message about the wrong thing. Tested via `baselineTimeoutMs: 1`.

**Journal write side in shared `/tmp`.** Taken, with your suggestion: `node_modules/.cache/mutation-guards/journal.json`. Per-worktree (which this repo needs), already ignored, no hashing, no `.gitignore` edit. You put the objection precisely — the docblock's "a blob found there is a suggestion, never a list of write instructions" was true in one direction only. `validEntries` stays, and its comment now says why: a journal is a list of "write these bytes to that path" whether or not today's directory happens to be trusted.

**Live-pid refusal before `validEntries`.** Fixed as you described — gated on `entries.length > 0`, since `writeJournal` never emits an empty list for a root, so an entry-less journal names nothing two runs could collide over. The existing test moved to a one-entry journal, and gained a sibling asserting that `{"pid":…,"entries":[]}` does *not* brick the gate. This was the one I most wanted too: "reachable without an attacker" is a different category from the rest.

**`--guard=name`.** Fixed. Same class as the bare-`--guard` bug, one spelling over, and the greener of the two — which is the worse direction.

**`clearJournal` serialising foreign entries.** Both `writeJournal` and `clearJournal` now scope to the root they're given.

**The `nonTerminating` fixture's unbounded `push`.** Capped, with the comment you asked for — raising the timeout being the instinctive response to a flaky hang test is exactly why it needed one.

`bun run check` exit 0 (3,470 pass), `bun run check:skills` exit 0, `--guard=` verified against a real row, tree clean after every invocation.

PR #755, opened 2026-09-17
fix(ci): one SHA was labelled two versions, and nothing could tell

ignaciohermosillacornejo:
# Summary

`actions/checkout` was SHA-pinned in **eight** workflow steps: six commented `# v7`, two `# v6`. Same action, same SHA, contradictory labels in one repo — and the wrong pair had been wrong across a major version. Dependabot then bumped the SHA (#671, v7.0.0 → v7.0.1) and faithfully preserved *both* comments, so the drift survived the bump.

Resolved against the upstream tag API rather than the bump message:

| SHA | Actual tag |
|---|---|
| `3d3c42e…` (new) | **v7.0.1** |
| `9c091bb…` (old) | **v7.0.0** |
| `d23441a…` | v6 — unrelated to either |

So `# v6` was simply false. Both now read `# v7`, matching their six siblings.

## Why it matters

This repo pins every action by SHA for supply-chain reasons (#654). **The pin is the security control; the trailing comment is the only human-readable half of it** — it is what a reviewer reads when asking "is this the version we meant?". A comment that disagrees with its SHA defeats that half silently, because nothing executes a comment.

## Class-level gate

`scripts/check-action-pins.ts`, wired into `bun run check`. Asserts every SHA-pinned action carries a version comment, and that the same `action@sha` carries the **same** comment everywhere.

**Deliberately NOT resolving SHAs against GitHub's tag API.** That would make `bun run check` require network and a token, and fail offline for reasons unrelated to the change under test. This is an *internal consistency* check: it cannot tell you a comment is right, only that the repo does not contradict itself about it. That is the achievable half — and it is the half that actually drifted. The limitation is stated in the script header rather than left to be discovered.

## Mutation-tested, both directions

| Mutation | Result |
|---|---|
| Reintroduce the v6/v7 split | **fails**, naming every location on both sides |
| Strip a version comment entirely | **fails**, naming file and line |
| A *different* SHA of the same action with a different label | **passes** |

The third is the cry-wolf control: different SHAs are different versions, and a gate that flags that gets disabled.

Swept all **21** pins in the repo; this was the only mismatch, so it is an instance with a gate rather than widespread rot.

## Two incidental notes

- The new script had to be allowlisted in `.gitignore`'s `scripts/*` rule — and **`check:tracked-files` caught it unallowlisted**. That gate shipped hours ago in #746 for #729; this is its first real outsider, and it worked.
- While pushing, `tests/core/database-errors.test.ts › getItems rejects when dbPath is undefined and cache is empty` failed **1 run in 3** under load, taking 5417ms for what should be instant. Pre-existing and unrelated to this change — filed separately.

## External assumptions

**None** about Copilot's API or data. The claim about which upstream tag each SHA corresponds to was verified live against `repos/actions/checkout/git/refs/tags`, and the gate deliberately does not depend on that API at runtime.

## Bug fix?

Root cause:       A SHA pin's version comment is free text that nothing parses, so `# v6` on a v7.0.0 SHA survived indefinitely and was then carried across a Dependabot bump.
Bug class:        `unparsed-claim-beside-verified-value` — a machine-checked value (the SHA) paired with a human-readable label that no check reads, so the two drift apart silently and the label is what humans trust.
Detector added:   `scripts/check-action-pins.ts` in `bun run check`, plus `tests/scripts/check-action-pins.test.ts`. Mutation-verified in both directions including a cry-wolf control.
Siblings checked: Swept all 21 SHA-pinned actions across `.github/workflows/`; `actions/checkout` was the only action labelled inconsistently. Actions whose tags do not resolve directly to the pinned SHA (`1password/load-secrets-action`, `actions/github-script`, `anthropics/claude-code-action`, `codecov/codecov-action`) are not mislabelled — their tags simply point at a different object — and the gate does not claim otherwise.
Ledger updated:   n/a — not an external-assumption bug; nothing here touches Copilot's API, schema, or data.


ignaciohermosillacornejo:
Rebased onto `main` (`bcc2b8d`) and closed the `codecov/patch` gap. Nothing about the fix itself changed: the two `# v6` comments on `3d3c42e…` still become `# v7`.

## The rebase — three resolutions, one of them semantic

**`package.json` (real conflict).** `main` added `check:mutation-guards` to the composite in the same position this branch added `check:action-pins`. Kept both, but *not* in the order git would have produced: `mutation-guards.ts` documents its placement as deliberate — immediately before `bun test --bail`, so the full suite runs afterwards over the restored tree, which is the backstop if its in-place restoration were ever wrong. Appending after it would have quietly taken that property away. Final order is `… && check:tracked-files && check:action-pins && check:mutation-guards && bun test --bail`, and the `check:action-pins` script entry moved out of the tail of the file into the gate group next to its siblings.

**`.gitignore` (auto-merged, then moved).** `!scripts/check-action-pins.ts` had landed next to `!scripts/launcher.sh`; it now sits in the `check-*` block between `check-tracked-files` and `mutation-guards`, matching the composite order above. `check:tracked-files` passes either way — this is legibility, not correctness.

**`.github/workflows/*` (auto-merged clean).** `main` edited `audit-reviews.yml` elsewhere in the file (the unclassifiable-finding escalation). No overlap.

Also updated the `bun run check` chain in `CLAUDE.md`, which `main` had started listing gate-by-gate — adding a gate and leaving that line short is the same drift class this PR is about.

The gate sees **23** pins now, up from 21, and all are consistent.

## Patch coverage

The twelve uncovered lines were the entire `import.meta.main` block. The tests covered `collectPins` and `findPinProblems` — the parsing and the comparison — so what went untested was which outcomes are *failures* and what CI is told about them. That is the operational half.

`runCheck()` now returns `{ code, stdout, stderr }` instead of printing and exiting, and the entry point is four lines that perform it. `WORKFLOW_DIR` honours `CHECK_ACTION_PINS_DIR`, matching `CHECK_WORKFLOWS_DIR` on the sibling gate, so the shipped script can be driven end to end against a synthetic tree instead of only against this repo.

Six tests: three in-process on `runCheck`, three spawning the real CLI. A clean tree exits 0 and says how many pins it saw; a contradictory one exits 1, carries `::error::` so GitHub annotates the failure rather than burying it in a collapsed log, and prints no reassuring line beside it; a tree with no pins exits 1 rather than reporting a clean scan over nothing.

Three more mutations on top of the three already in the description:

| Mutation | Result |
|---|---|
| Drop `process.exit(result.code)` | **fails** — 2 CLI tests. A gate that cannot exit non-zero stops nothing |
| Drop the zero-pins vacuity floor | **fails** — at both the `runCheck` and CLI levels |
| Drop the `::error::` prefix | **fails** — both annotation tests |

Unmutated: 18/18 pass.

Script line coverage 78.9% → 94.6%. The four still uncovered are the `import.meta.main` block itself, which the three spawn tests *do* execute — in a child process bun's instrumentation cannot see. Worth stating rather than chasing: the alternative is an in-process test of a two-line print-and-exit wrapper, which walks the line without being able to fail.

One thing deliberately not done: `tests/scripts/check-action-pins.test.ts` is not on `tsconfig.tests.json`'s include list, unlike its two `tests/scripts/` neighbours. Their argument is #725 — a failure message that silently reads `undefined` because a literal is missing a declared field. This file has no such message; a rename on `CheckResult` turns its assertions red at runtime rather than passing quietly. Happy to add it if you'd rather the convention be uniform.

`bun run check`: 3613 pass, 20 skip, 0 fail (184 files), via the pre-push hook.

ignaciohermosillacornejo:
Heads-up before you merge: the review findings above are **addressed in the worktree but not on the branch**. `git commit` has been returning `error: 1Password: agent returned an error` (exit 128) for the last ~15 minutes across eight attempts — Touch ID on the signing key, nobody at the machine. Two commits earlier in this session signed fine, so it is transient.

What is on the branch (`bcc2b8d`) is complete and green: the rebase and the patch-coverage fix. What is uncommitted, all passing `bun run check` (3616 pass / 0 fail):

- **Finding 1** — `.js` import, `tests/scripts/check-action-pins.test.ts` added to `tsconfig.tests.json` and pinned in `MEMBERSHIP_IS_THE_COVERAGE`
- **Finding 2** — folded into the `doc-reality-drift` row in `docs/bugs/README.md` rather than given a second name, with the reasoning written into the row
- **Finding 3** — `WORKFLOW_DIR` resolved from `import.meta.url`; two tests stop hardcoding the relative path; printed paths stay repo-relative via `displayPath()`
- **Finding 4** — quoted scalars collected (mutation-checked: reverting the quote classes fails the new test); `# v7` vs `# v7.0.1` on one SHA documented as a deliberate granularity rule and tested
- **Finding 5 / 6** — shebang; the stale `~21` count out of the test floor's comment; the three limitations (one label on two SHAs, unpinned refs outside the scan, lines-not-YAML) stated in the header as decisions

Merging now is safe and loses only that polish. The commit message is ready — the finishing move is `git commit -F …` and a `--force-with-lease` push, once the signing agent answers.

PR #760, opened 2026-09-17
fix: a decision derives its input from the wrong source, twice

### LINKED ISSUE #741: auto-merge's approver identity comes from github.repository_owner, which stops being a person if the repo moves to an org
`.github/workflows/auto-merge.yml` derives the approving identity from
`github.repository_owner` (set at `env.OWNER`, used in the job `if:` and in the `jq`
filter). For a personal repo that is exactly the current hardcoded login, which is why
PR #727 moved to it — one source instead of a literal in two places.

If the repository is ever transferred to an organization, `github.repository_owner`
becomes the **org login**, which is not a reviewer. Every review then reads `NONE` and the
workflow refuses to enable auto-merge on everything.

## Why this is low priority

It **fails closed**. The failure is loud (an `::error::` naming the state) and the
workaround is the same one-liner the workflow wraps:

```bash
gh pr merge <num> --auto --rebase
```

No PR can merge *without* review because of it. It is an availability bug, not a safety
one — which is the right direction for this particular guard to fail in.

## Options if it ever matters

- An explicit `APPROVERS` list in `env:`, which is also what would be needed the moment
  there is more than one maintainer — the current design assumes exactly one approver
  either way.
- Or resolve the owner's login once and pin it, accepting the literal.

Not worth doing speculatively: the repo is personal today, and the single-approver
assumption predates this and is encoded in `claude-review.yml` and `npm-publish.yml`'s
`TRUSTED_PUBLISHERS` too. Whoever adds a second maintainer will touch all three.

## Provenance

Raised by the AI reviewer on PR #727 (final pass, non-blocking) and deferred rather than
held, to stop the rebase cascade on that PR.

### LINKED ISSUE #751: auth: an endpoint-level failure discards a known-good refresh token, forcing a browser-wide re-extract
Surfaced by the review of #749 (the #742 fast-path guard), and pre-existing — not a regression from that PR.

## What

`exchangeToken` in `src/core/auth/firebase-auth.ts` clears `this.refreshToken` on **any** non-OK response, before throwing. That includes the endpoint-level statuses the rest of the module is careful to distinguish from candidate-level ones: `403 PERMISSION_DENIED`, `429` rate limit, `API_KEY_INVALID`.

## Why it matters

#749 made the fast path rethrow those rather than swallowing them (correct — a 403 is not a verdict on the token). But the refresh token is already gone by then, so the **next** `getIdToken()` cannot take the fast path at all. It falls into a full cold re-extract: a browser-wide Local Storage read plus up to `MAX_EXCHANGE_CANDIDATES` exchanges against the endpoint that just rate-limited us.

Both directions point the same way — a transient outage should cost one retried request, not a browser-wide scrape of every site's Firebase tokens.

## Shape of the fix

Gate the clear on the same predicate the two paths now share: only a candidate-level rejection (`isCandidateRejection`, or specifically a `DEAD_TOKEN_CODES` match) should invalidate the cached refresh token. An endpoint-level failure should leave it intact.

## Detector

A test that a `429` (or `403`) on the fast path leaves `refreshToken` intact, so the following call retries the fast path without a cold extraction — asserting the extractor was not invoked a second time, which is how #749's new fast-path tests discriminate.

## Class

Same family as #722: a signal about the *endpoint* being read as a verdict about a *token*. #722 fixed it in the candidate loop's continue/abort decision and #749 fixed it in the fast path's catch; this is the third place the same distinction is not drawn.

ignaciohermosillacornejo:
# Summary

Two bugs, one mechanism: a decision read a signal that merely *correlates* with the fact it needed. `auto-merge.yml` took the approving identity from `github.repository_owner`, and `exchangeToken` took "the request failed" as a verdict that the token it sent was dead. Both proxies are faithful in the configuration they were written in, which is why both survived review.

Closes #741
Closes #751

## External assumptions

Not None — both fixes rest on behaviour of an external system.

**#741 — GitHub (`GitHubApi.pullsReviews:standingApproval`, new ledger entry, `unverified`)**

- *probe transcript:* `GET /repos/ignaciohermosillacornejo/copilot-money-mcp` → `owner.login: "ignaciohermosillacornejo"`, `owner.type: "User"`. So `github.repository_owner` and the login now declared in `env.APPROVERS` are the **same string** today — this is why the change is a no-op on this repository. `GET /repos/{owner}/{repo}/pulls/744/reviews` → per-review `user.login` / `state`, the fields the filter now selects on.
- *unverified, carried into the ledger:* that reviews are returned chronologically (the filter takes the **last** decisive one); that a dismissed approval reports `DISMISSED` rather than vanishing; and that a job-level `if:` can never read the `env` context — which is the whole reason the approver login is repeated as a literal there. The first two predate this PR (#727). The third is documented by GitHub and deliberately not probed, because probing it means shipping a workflow that errors. All three fail closed: the wrong answer is a merge that does not happen.
- *live round-trip, on this PR:* the `claude-review.yml` step `if:` now reads `contains(fromJSON(env.APPROVERS), …)`. `pull_request` runs the PR's own workflow file, so if an approving review from `github-actions` appears on this PR, the `env`-in-step-`if:` assumption is confirmed by execution rather than by citation.

**#751 — Google securetoken (`Securetoken.v1Token:endpointFailure`, existing entry, still `unverified`)**

- The same classification that decided which errors to *report* now also decides which errors may *discard* the cached credential. The entry's evidence records that second consumer. A code missing from `ENDPOINT_LEVEL_ERROR_CODES` now costs one wasted fast-path request on the next call — the opposite direction from the bug, which spent a browser-wide re-extract.
- No new claim about the endpoint: the 403/429/`API_KEY_INVALID` shapes were probed for #722 and are unchanged here.

## Bug fix?

Root cause:       A destructive decision took a correlated signal as its input — `github.repository_owner` for "who approves", a non-OK response for "this token is dead" (`.github/workflows/auto-merge.yml`; `src/core/auth/firebase-auth.ts` `exchangeToken`).
Bug class:        `proxy-for-authority` — new, added to `docs/bugs/README.md`. The proxy agrees with the authority in the configuration it was written in and diverges silently in every other; distinct from `identity-resolution`, where the mismatch is visible in the values.
Detector added:   Two, one per substrate, both mutation-verified. (1) `scripts/check-workflows.ts` invariant 3: no workflow may mention `github.repository_owner` **anywhere** in the parsed document — not just in an `if:`, because `env:` laundering is exactly how this one hid — and every login literal compared in an `if:` must equal the declared `env.APPROVERS` / `env.TRUSTED_PUBLISHERS` set, in both directions. (2) `tests/core/auth/candidate-ordering.test.ts`: a matrix driven from `ENDPOINT_LEVEL_STATUSES` and `ENDPOINT_LEVEL_ERROR_CODES` themselves — every member must leave a cached credential intact, and every `DEAD_TOKEN_CODES` member must discard it — so a signal added to a list is asserted the day it is added.
Siblings checked: The workflow gate found two on its first run, both fixed here. `npm-publish.yml` declared `TRUSTED_PUBLISHERS` with the comment "only these users can publish packages" and **read it nowhere** — three gates carried their own copy of the literal instead; the verify step now tests membership in the declared list. `auto-release.yml` repeated the login across three job gates with nothing to check them against. On the auth side this is the third site of the endpoint-vs-token distinction (#722 fixed the candidate loop, #749 the fast path's catch); the other two are correct and now share this one's predicate.
Ledger updated:   `Securetoken.v1Token:endpointFailure` evidence extended with its second consumer; new `GitHubApi.pullsReviews:standingApproval` entry, `unverified`.

## Does this make auto-merge looser? No.

`github.repository_owner` on this repository evaluates to exactly the login now declared (probed above), so the set of PRs that can auto-merge is bit-identical: a standing `APPROVED` from that login, at the current head, verified against the API on both the review and dispatch paths. Nothing about the fork-PR posture moved — the cheap pre-runner gate still refuses to start a runner for a non-approver's review, SHA-pinned actions and the concealment gate are untouched, and `workflow_dispatch` still requires write access.

The only behaviour that differs is the failure this issue is about: after an organisation transfer, the workflow goes from refusing **everything** to admitting what one named human approved. That is the repair, and it is never wider than the policy the file always intended.

## Evidence

**#751** — 8 endpoint-level rows red before the fix (`expect(extractions).toBe(1)` → received `2`: the cached token was gone, so the retry re-read the browser), 0 after. Mutation-verified four ways: clear unconditionally → 8 red; never clear → 2 red; drop `isExplainedByLoggedOut` from the shared predicate → 2 red; drop `isCandidateRejection` → 1 red.

**#741** — with invariant 3 added and the workflows untouched, `check:workflows` failed naming `auto-merge.yml`, `claude-review.yml`, `npm-publish.yml` and `auto-release.yml`; it passes after. Mutation-verified: reverting each of the four workflow fixes turns the real-repository test red, and deleting the owner-proxy rule, weakening set-equality to membership, or narrowing the `if:` walk to job level each turns a synthetic case red. All nine workflow files re-validated with PyYAML, not by eye.

`bun run check` and `bun run check:skills` both exit 0. PRIVACY.md gains the read *frequency* it never stated, since this change reduces it.


ignaciohermosillacornejo:
Thanks — finding #1 is the best catch here, and it is the PR's own thesis turned on the line the PR wrote. All eight addressed; seven in code, one filed.

**1. The discard needed two facts.** Fixed as suggested: `if (isTokenFinished(failure) && this.refreshToken === refreshToken)`. Your interleave is now a test rather than a comment — two callers enter the fast path holding the same expired credential, the first clears/re-extracts/installs a fresh one, and the second's rejection is released only afterwards. With the identity half dropped the third call re-reads the browser and the test goes red; the assertion is the exact attempt sequence plus `extractions === 3`, so the failure names what happened. The fast-path invariant is restated as the stronger claim it now supports: *fell through ⇒ the token it failed on is no longer cached* — either this cleared it, or someone else had already replaced it with a live one.

**2. `fixed_in`.** Both links point at #760. (Caught independently a few minutes before your review landed — the entry was written before the PR had a number, which is its own small lesson.)

**3. Grammar narrower than the rationale.** Both gaps closed rather than documented away.
- `OWNER_PROXY` now matches `github['repository_owner']` too.
- A declared login may no longer appear **anywhere** except the declaration and an `if:` gate — `run:` blocks, action inputs, everything. That is the npm-publish shape exactly, and you are right that it survived only because the old declaration was a bare string. A companion test pins the remedy (read `$TRUSTED_PUBLISHERS` from the shell) so the rule is not just a wall. Verified there are no legitimate occurrences of the login outside gates today, so the rule lands clean.
- What remains genuinely unreadable — a login compared in a shell script by a workflow that declares *no* list, where there is nothing to compare against — is now stated in the invariant's header instead of implied, and in the `docs/bugs/README.md` detector column.

**4. Positive bot comparison.** Right, and it would have put a bot where the humans are named. `…[bot]` logins now sit outside the rule in both directions (GitHub logins cannot contain brackets, so the suffix is exclusive to Apps). Test covers `github.actor == 'dependabot[bot]' || github.actor == 'octocat'` passing with only the human declared.

**5. Two spellings.** Both are `IN(...)` now — `select(.user.login | IN($approvers[]))` and `jq -e --arg p "$PUBLISHER" 'IN($p; .[])'` — and the failure message for #3 hands out the same spelling. The `contains()` case-insensitivity asymmetry is recorded next to the expression, since a reader meets it there.

**6. `${{ github.event_name }}`.** Now `$GITHUB_EVENT_NAME`. Agreed that leaving one interpolation in the step whose point was to stop doing that reads badly regardless of exploitability.

**7. PRIVACY.md.** Corrected — the cached access token serves calls with **no network request at all** until roughly a minute before its hour is up. A frequency bullet that overstates frequency is the one kind of privacy-doc error worth fixing even though it errs toward caution.

**8. `scripts/` outside `format:check`.** Filed as #762 rather than folded in. The one-line glob change drags 21 reformatted files with it, and burying two fixes under that much whitespace is the opposite of what this PR is for.

Also: your note that `pull_request_review` runs the default-branch workflow while `claude-review.yml`'s auto-approve runs the head version is the right frame, and that second half gave the PR a live round-trip it did not have when opened — the auto-approve step fired on this PR, which is what confirms `env` is readable from a step-level `if:` by execution rather than by citation.

Not re-running `bun run check` in the sandbox is the honest call; it exits 0 locally, along with `check:skills`, and every mutation named in the PR body was re-run after these changes.

Issue #708, opened 2026-09-12
Quick start should include logging into a browser


tuxbook:
As a pre-requisite in the quick start, or an instruction line, it should mention that the user needs to log into https://app.copilot.money/ in a browser since that's where the MCP gets its auth from (at least for --live-reads, which I have configured).

I found this out from logs, but it should be in the setup guide.

ignaciohermosillacornejo:
Agree. it's mentioned on the readme but should probably be more prominent. Agree on the setup guide and will give it some thought from a product persepective as to the discoverability aspect.

Issue #705, opened 2026-09-11
mirror-pin ratchet: anchor discovery to line start — a pin written in prose counts as a pin


ignaciohermosillacornejo:
Follow-up to #703 (review round ten). Not blocking — filed rather than grown onto that branch.

## The hole

`tests/core/graphql/mirror-pin-coverage.test.ts` discovers mirrors and pins with unanchored regexes run over raw file text, so a `: ExactKeys<` written inside a **comment** counts as a real pin.

This is not hypothetical — it is already true in the tree, and the phantom comes from the ratchet's own documentation. The `Usage` block in `ExactKeys`' JSDoc (`src/core/graphql/queries/_shared.ts`) writes an example pin in prose, and the scanner counts it:

```
_shared.ts: unanchored=['FOO_NODE_MIRROR_IS_EXACT', 'MARKET_INFO_...', 'SECURITY_...']
            anchored  =['MARKET_INFO_...', 'SECURITY_...']
            PHANTOM: FOO_NODE_MIRROR_IS_EXACT
```

Reproduce by running the test file's own `PIN` regex with and without `^`/`m` over `src/core/graphql/queries/*.ts`.

**Inert today** — nothing declares a `FooNodeSchema`, so no lookup resolves against the phantom. But it reopens exactly the fakeable-remedy hole round nine closed: when the ratchet fails, its message names the missing pin, and the shortest silencing edit is now "write that identifier in a comment." The failure message even points the reader at the copy-pasteable usage block that demonstrates the trick.

## Fix

Anchor both regexes to line start:

```ts
const NAMED_MIRROR = /^export const (\w+NodeSchema)\b/gm;
const PIN = /^export const (\w+_MIRROR_IS_EXACT):\s*ExactKeys</gm;
```

All five real mirrors and all five real pins are at column 0, so nothing is lost — verify that before and after. Then extend the existing `a bare = true const is NOT counted as a pin` test with a third case proving a commented pin line (` * export const TAG_NODE_MIRROR_IS_EXACT: ExactKeys<…>`) yields zero matches, and say in the header that discovery is line-anchored so prose cannot satisfy it.

Mutation-check both directions, as that file's existing tests do: a real annotated pin must still be accepted, not just the fake rejected.

## Two smaller items from the same review, same file

[...]

PR #766, opened 2026-09-18
fix(server): a live-reads boot failure now reaches the agent, not just the log (#708)

ignaciohermosillacornejo:
# Summary

The runtime half of [#708](https://github.com/ignaciohermosillacornejo/copilot-money-mcp/issues/708), which #717 closed with docs only. `--live-reads` ran a boot probe and `process.exit(1)` on failure, so the one actionable fact — log into `app.copilot.money` — only ever reached a log file. The probe is now non-fatal **and** runs after the transport connects: the server comes up, live tools stay listed, and the first call returns the failure as an `isError` result the assistant can relay.

Relates to #708 (already closed for its docs half).

## External assumptions

None

## Bug fix?

Root cause:       `runServer` reported a live-reads boot-probe failure by killing the process, destroying the MCP transport that was the only channel able to carry the explanation to the model.
Bug class:        A failure inside the server reported by terminating the process rather than through an MCP result. Every member is invisible the same way, and the class is open — a future cache-load guard, schema check or config validation could each reach for `process.exit(1)` in `src/server.ts`.
Detector added:   `tests/unit/live-auth-boot-nonfatal.test.ts` walks the TypeScript AST of every module under `src/` and fails on any non-zero `process.exit` outside `src/cli.ts` (the **argv path**, which runs before a transport exists — the one place where exiting *is* the report). AST rather than a text scan for the reason #705 documented from the other side: a pattern that merely *contains* the call would count the JSDoc explaining its removal, so that case is asserted too. A second, unrelated class detector was added for a mistake made **during** this PR — see "Second class" below.
Siblings checked: The only other `process.exit` call sites in `src/` are `src/cli.ts` (argv/`--help`, legitimately fatal) and `src/server.ts`'s SIGINT/SIGTERM handlers (`exit(0)` — a clean shutdown the client asked for). Both are permitted by the gate and pinned by it. `cli.ts`'s `unhandledRejection` handler is *not* covered by the carve-out's reasoning, which is why the fire-and-forget probe carries its own `.catch`.
Ledger updated:   n/a — not an external-assumption bug; no GraphQL surface changed.

## What changed

- **`src/server.ts`** — the boot probe moves into an exported `preflightLiveAuthOrWarn()` that logs and returns instead of exiting. Deliberately not classified by error code: a `SCHEMA_ERROR` or `NETWORK` failure at boot has the same remedy shape as an auth one, and killing the transport is the wrong answer to all three. The stderr lines stay, reworded, for host-log debugging.
- **Connect first, probe after.** Exiting was one way to destroy the channel; blocking is the other. An offline boot spends ~125s in the probe (four 30s attempts plus backoff), and a host whose startup timeout fires during that wait sees the same closed transport. `runServer` now connects stdio first and fires the probe off afterwards, where it still writes its diagnostic and warms the token cache.
- **The probe carries its own `.catch`.** `no-floating-promises` defaults to `ignoreVoid: true`, so `void` is the suppression and cannot be a backstop. Without a handler, a rejection would reach `src/cli.ts`'s `unhandledRejection` → `process.exit(1)` *after* the transport connected — the same class, in its worst form.
- **Recovery needs no restart.** After a failed cold extraction `FirebaseAuth` has cached no token, so the next tool call re-runs browser extraction and picks up the session the user just created. Exiting made every failure permanent, including the transient ones (no network at launch, laptop asleep).
- **Nothing is served stale.** A live tool with no session fails; it does not fall back to the cache.
- **Docs** — `docs/graphql-live-reads.md` said the server "exits non-zero… check the MCP server logs", which is now false; README's prerequisite bullet said those modes "fail at startup". Both updated, plus CHANGELOG entries.

## Tests

- Four boot-failure shapes (no browser session, `AUTH_FAILED`, `NETWORK`, `SCHEMA_ERROR`) each assert the probe resolves and `process.exit` is not called, plus a success case asserting a healthy boot is **silent**.
- End-to-end: `get_transactions_live` on a session-less server returns `isError` with the login URL in the text — the only channel the model can read.
- Ordering is pinned by **running** it. `runServer` takes a two-function seam (`RunServerDeps`: client construction, transport connect — both defaulting to the real thing; `src/cli.ts` passes nothing) because `server.run()` claims this process's stdio. The test drives the real probe against a client whose `query` never settles, racing a 5s timeout so a reintroduced `await` fails rather than making the suite quietly take two minutes. The two claims are awaited separately, so the assertion does not depend on microtask depth in `fetchTransactionsPage`.
- Every gate mutation-checked rather than assumed: awaiting the probe, moving it before `connect`, and reintroducing `process.exit(1)` each turn a test red.
- **Two tests were written and then removed or rewritten for being green-by-construction**, with the reasoning kept in the file both times. A gate for the new `.catch` passed with the `.catch` deleted, so it went. An earlier AST ordering test asserted a literal `server.run()` call site, and its own anti-vacuity guard fired once the seam removed that spelling. A later "the two causes carry different messages" test compared two literals in its own file and never ran the script — verified by collapsing both message branches and watching all 27 tests pass — so the claim moved into the loop that executes the gate, as the negative assertion.

## Second class: generated output in the index

Mid-PR, a `git add -A` after a local `bun test --coverage` committed `coverage/` — 43,687 lines of machine output, including bun's in-progress `.tmp` scratch file — and **every gate stayed green**. `scripts/check-tracked-files.ts` asked whether every file the tooling *needs* survives a fresh clone; nothing asked the inverse.

It now also fails on any tracked file under a directory `bun run clean` deletes, **derived from that script** rather than from a list of its own — one definition, shared by the seeds sweep and the new rule, so the two cannot answer "is this generated?" differently. The parse fails loudly on a `clean` it cannot read, with one sentence per rejection cause: shell syntax (stated as the allowlist, with operators/globs/braces/quoting/`~`/escapes as examples) and targets outside the repo (a scan over `git ls-files` can only guard paths git emits). A `clean` naming no targets exits rather than scanning nothing. The failure text names the remedy that *stays* fixed — `git rm -r --cached` plus an ignore rule — because deleting the files alone is what let the next `git add -A` re-add them.

**Known gap, deliberately not closed here:** `scripts/` is typechecked but neither linted nor formatted, so it now holds five repo gates with no style enforcement. Widening the prettier glob rewrites 21 files unrelated to this PR, and burying a diff under machine-driven output is the mistake this branch already made once. It wants its own PR.

## Verification

`bun run check` — including `check:mutation-guards` — plus `check:skills` all pass; 3,588 tests pass. The single failure (`TransactionMetaStore` cap-valve) is pre-existing on `main` in this environment: it `chmod`s a directory to `0o555` to force a write failure, which root ignores. Verified by running that file on a clean tree; CI runs it non-root and it is green there.

https://claude.ai/code/session_01AUg7JiBqtEU8w8wUnMk2PP


ignaciohermosillacornejo:
All five findings addressed in `2467ffd`. Two notes where I did something other than what was suggested.

**Finding 1** was the right call and I took it as stated — the probe now runs after `server.run()`, fire-and-forget. The suggested test ("assert the transport connects without waiting on a hanging probe") isn't reachable behaviourally: `runServer` claims stdio, so it can't be called in-process. I pinned it structurally instead — an AST walk over `runServer`'s body asserting the `server.run()` call precedes the probe call and that the probe isn't awaited — and the docblock states plainly what that proves and what it doesn't (a probe moved into a helper awaited before `server.run()` would evade it).

**Finding 3** is closed for `process['exit'](…)` and named as a KNOWN LIMIT for the rest. The destructured/imported/aliased forms need binding resolution — a `ts.Program`, not a `SourceFile` — which is more machinery than this gate earns, and none of the three is a shape anyone here writes. The element-access branch is verified against a fixture run through `processExitArgs` itself rather than an inline copy of the matcher, with the near-misses (`process.exitCode =`, `process.emit('exit')`, a commented-out call) asserted uncounted.

On the minor about `toContain('app.copilot.money')`: correct that for a `GraphQLError` the URL comes from the second, static line — `graphQLErrorToMcpError` renders "Sign in to the Copilot web app" without one. I left the assertion as-is and said so in a comment, because the pair of lines is what a person reading the log actually sees; the per-error rendering isn't the claim.

All three gates mutation-checked rather than assumed: awaiting the probe, moving it back before `server.run()`, and reintroducing `process.exit(1)` each turn a test red.

Also worth recording, since your testing note flagged it — you were right that CI is the authority: `bun run check` can't pass in my environment either, but for an unrelated reason. `tests/core/persistence/transaction-meta-store.test.ts` `chmod`s a directory to `0o555` to force a write failure, which root ignores, so it fails on clean `main` here too. CI runs it non-root and it's green there.

---

ignaciohermosillacornejo:
Finding 1 was right on both halves, and I verified each rather than taking them on trust: `eslint.config.js` has no `no-floating-promises` override, so `ignoreVoid: true` stands and `void` is exactly the suppression — my comment named as its backstop the one thing it switches off. And `src/cli.ts:135-139` is as quoted, armed for the whole process lifetime, so a rejection would exit *after* `server.run()` connected. Fixed in `5658e97`: the probe carries its own `.catch`, which holds whatever the callee does later. The detector's carve-out is narrowed to cli.ts's **argv path** — you're right that `unhandledRejection` and `main().catch` aren't covered by "runs before any transport exists".

One thing I did differently, and it's worth recording since it cuts against adding coverage. I wrote the obvious test for the new `.catch` — listen for `unhandledRejection` across a `runServer` call with a rejecting probe — and then mutation-checked it. **It passes with the `.catch` deleted.** `preflightLiveAuthOrWarn` swallows everything, so through the seam its promise cannot be made to reject, and the only way to provoke the real failure is the later edit the `.catch` exists to survive. So the test could never fail, and I deleted it rather than ship a green-by-construction gate; the reasoning is in the file's docblock instead. The guarantee here is the handler's presence, not an assertion about it.

Separately, since it landed after your pass: `codecov/patch` went red at 42.85% on `fccf8fb`, and it was a fair signal. Reordering the probe rewrote most of `runServer`, which no test executes — so the ordering rule had no behavioural coverage at all, and the AST walk I'd added in its place was asserting a literal `server.run()` call site, i.e. a spelling. `647097a` replaces it: `runServer` takes a two-function seam (client construction, transport connect — both defaulting to the real thing, `cli.ts` passes nothing), and the test drives the real probe against a client whose `query` never settles, racing a 5s timeout so a reintroduced `await` fails the test instead of making the suite quietly take two minutes. The AST test's own gate-guard is what condemned it — once the seam removed the literal call site, `server.run() not found in runServer` fired. Patch coverage 42.85% → ~96%.

Also merged `main` (`fccf8fb`) — additive conflict in `tsconfig.tests.json`, both sides appending an include entry; kept both.

---

ignaciohermosillacornejo:
All four addressed in `24fa684`. Finding 3 (the `void`'s backstop) and finding 4 (temp dir) had already landed in `5658e97`, which your pass predated — 4 was still open in the tree you read, so I've re-checked it against the current one.

**Finding 1 was mine to have avoided entirely, and you're right that deleting the files isn't the fix.** `git add -A` after a local coverage run swept them in; without a `coverage/` rule the next one does it again. Untracked and ignored, with the note pointing at the `LOG` entry four lines above it — same class, same file, already documented, which is what makes this one embarrassing rather than unlucky. −43,687 lines.

**Finding 2 was the sharpest thing in this pass.** You traced it correctly: the assertion held only because `fetchTransactionsPage`'s first statement is `await client.query(...)` with nothing before it, so `'probe'` landed in the same microtask. A test that fails when an unrelated file gains an `await` — and fails with a message blaming an ordering rule that is perfectly intact — is worse than a weaker test, because it sends the next reader somewhere there's nothing wrong. Now awaited as two separate facts: the race proves `runServer` returned without waiting, a deferred resolved from `onQuery` proves the probe started. Both mutations re-verified after the change (awaiting the probe → timeout; probe before `connect` → order assertion fails), so the decoupling didn't cost the gate its teeth.

Nits taken too: the timeout helper clears its timer and both ordering tests share it, so the cache-mode case reports a hang in its own words rather than on bun's default; the seam's two defaults are named functions sitting together that `RunServerDeps` points at.

One I left: `callPos`'s `startsWith` is gone with the AST ordering test it belonged to, so there's nothing to tighten.

On the `FirebaseAuth` in-flight de-duplication note — agreed it's a real gap and agreed it's not this PR's. Worth filing; I haven't, so as not to open an issue on your repo unprompted.

---

ignaciohermosillacornejo:
The coverage blocker was already fixed in `24fa684`, which this pass predated — but the part of it I hadn't done was the part that matters, so thanks for pushing on it.

You flagged the class observation as *"optional, and arguably its own PR"*. I took it, because this PR's own bug-response ritual says an instance-only fix doesn't satisfy "Detector added", and I'd otherwise be arguing that standard applies to `process.exit` but not to the mistake I made three commits later. `cf6ad85` extends `scripts/check-tracked-files.ts` with the inverse assertion: no tracked file under a directory `bun run clean` deletes, **derived from that script** rather than from a list of its own — so a renamed build directory can't leave the gate guarding a name nothing produces. Your three consequences are quoted in the docblock, because "it's noisy" wouldn't have justified the gate and the stale-upload one in particular is a real correctness bug.

Two things I'd have gotten wrong without checking:

- The anti-vacuity guard (a `clean` naming no `rm -rf` targets exits 1) **broke 16 existing tests** — the synthetic fixture repos have no `clean` script, so every case failed on a message about `clean` instead of its own subject. Fixed by giving the fixture one, which models the real repo better anyway. Worth knowing that guard has teeth in both directions.
- The failure message names `git rm -r --cached` **plus** the ignore rule, not just the deletion. Deleting the files alone is what would have let the next `git add -A` re-add them — the thing that made my instance a class rather than a slip.

Four cases, all mutation-checked: tracked coverage file flagged; same directory on disk but absent from the index passes (a contributor who just ran the tests shouldn't fail a check they can't act on); a tracked `dist/` file caught by the same code path with no second rule, which is what "derived" actually buys; reworded `clean` trips the guard.

`bun run check` and all nine `check:*` are green on `cf6ad85`, 3,582 tests passing — the one failure is still the pre-existing `chmod 0o555` test that root ignores in my environment.

---

ignaciohermosillacornejo:
Both passes converged on the same thing and they were right: I argued that a list is one more thing to keep in sync, then left `GENERATED_PREFIXES = ['dist/']` standing 330 lines above my own derivation. All findings addressed in `77b2f6b`.

**The two definitions are now one.** I verified the contradiction before fixing it, and it's worse than drift: renaming the build output and updating `clean` makes rule (4) forbid tracking `build/` while `isGenerated('build/cli.js')` goes false, so the seeds sweep demands it be tracked — one run, two failures, opposite remedies. After the hoist that rename passes cleanly; I checked by actually doing it to `package.json` and running the gate, not by reading.

**The parser failed open exactly where you said.** I took the union of both passes' tables. `./` and trailing slashes normalise; quotes, globs and shell operators now say "this script cannot read that as a directory" and exit — a token the parse can't read isn't a directory it's guarding, which is the same claim the empty-parse guard makes. Also widened to `-rf?` so `rm -r` doesn't slip past.

**On the inline exit** (raised as "report them together" in one pass, "worth a sentence saying that's the intent" in the other): I kept it and wrote the sentence, because after the hoist it isn't just defensible, it's required. `isGenerated` now feeds the seeds sweep, so a bad derivation doesn't merely skip rule (4) — it makes the "needed but not tracked" report wrong too. Printing findings computed from a definition the gate just admitted it couldn't read would be worse than reporting one failure at a time.

**The `-f` really did read backwards.** Dropped it: the fixture has no `coverage/` rule, mirroring the real repo, and a plain `add` is how the file actually got committed — `-f` would have kept the test green if the fixture ever gained one. `dist/` keeps `-f` in a case of its own since there it *is* ignored, and the contrast is now the point rather than an inconsistency.

**On the derivation test proving its own title:** fixed with `out-stage`, a name in the fixture's `clean` and nowhere in this repo, so no hardcoded list could contain it. You were right that `dist` proved only "two directories share a code path".

One thing worth reporting because it went wrong: my first draft listed `./dist` *both* as an unreadable spelling and as the normalised one — two tests asserting opposite things about the same input. The suite caught it immediately. Noting it since the alternative reading is that I hand-verified the table.

The `check-ignore` strengthening you floated — asserting each generated directory is *ignored*, which would have gone red on `main` before anything was committed — is the better gate and I've deliberately not built it here. It changes what the check asserts rather than fixing this PR's findings, and it belongs with someone deciding whether `.bun-build/`-style exceptions should exist. Happy to file it.

`bun run check` and all nine `check:*` green; 3,585 tests passing, same one pre-existing root/`chmod` failure.

---

ignaciohermosillacornejo:
Both correct, both are my previous commit's bug one spelling further out, so both got its answer rather than a patch. `f169573`.

**The predicate split was real and I'd missed it while fixing its sibling.** I unified the two *lists* and left the two *membership tests* — prefix-only in `isGenerated`, exact-or-prefix in rule (4) — which agree for every directory-valued target and diverge the moment `clean` names a file. Exactly the failure I'd just written a docblock claiming to have designed out, reached through a file target instead of a rename. One predicate now; `GENERATED_PREFIXES` went with its last consumer.

**The denylist was the wrong shape twice over.** Writing an enumeration of shell metacharacters in a file whose whole argument is "derive, don't list" — and then having it miss `rm -rf {dist,coverage,.bun-build}`, which carries no operator, quote or glob character and is the most natural way to write that script. It now states what a readable token *is*, so braces, character classes, `~` and escapes fall out of one rule together with whatever nobody's thought of. Your `rm -fr` / `rm -rfv` observation is in the docblock too, since it isn't visible from the regex.

The file-target case is asserted from both sides in separate tests so a failure says which direction broke: the seeds sweep must stay silent about a generated file absent from a fresh clone, and rule (4) must still refuse to let it be tracked. Both new gates mutation-checked against the code they replace — reverting `isGenerated` to prefix-only fails the file-target case, reverting the allowlist fails the brace case. Neither is green by construction.

One correction to your finding 1's third bullet, which doesn't change the fix. You noted the remedy gate `failures.length > trackedGenerated.length` assumes one failure per tracked-generated file, and that a path in both `trackedGenerated` and `ignoreMatched` would push two. That's right as stated, but the double-push isn't reachable through the file-target route: `ignoreMatched` only pushes for paths in `candidates`, and a generated path never enters `candidates` in the first place — it's filtered at the seeds sweep by `isGenerated`, which is now the same predicate. So the gate rests on the agreement your finding is about, which is your point, but the specific two-failure path was already closed by the unification rather than needing its own handling. I checked rather than assuming, since the alternative was a latent bug in the gate I'd just added.

3,587 tests pass, all nine `check:*` green, same one pre-existing root/`chmod` failure.

---

ignaciohermosillacornejo:
Findings 1 and 2 fixed in `069a4ac`. Finding 3 I've deliberately not done, with reasons.

**Finding 2 is the same failure a third time and the categorical argument settles it.** `PLAIN_PATH` tested for absence of shell syntax; the property the guard needs is "names something `git ls-files` could print". `../dist`, `/tmp/build` and `.` clear the first and fail the second. You're right that the same test already sits 200 lines down in `toRepoRelative` — the parse was the one place that didn't ask. Mutation-checked: reverting to the bare `PLAIN_PATH.test` fails the three new cases.

**Finding 1 is fair and slightly embarrassing** — the commit that deleted a second definition shipped a second *name*, with two docblocks `{@link}`-ing an alias of the function they document. One name now, and both blocks name their consumers rather than pointing at themselves.

**Finding 3 I'm declining here, and splitting the reasons.** The 114-column line was mine and is fixed. The trailing commas in call arguments are not a divergence this diff introduced — `main` carries four of them in this file and I added four more matching what was already there. Matching the file you're editing is the right call for a contributor even when the file's style is unenforced.

The underlying point — `scripts/` holds four repo gates and is typechecked but neither linted nor formatted — is correct and worth fixing. But `prettier --check "scripts/**/*.ts"` reports **21 files**, none of them this PR's subject. Burying a diff under unrelated machine-driven output is the exact mistake this branch already made once with `coverage/`, and I'd rather not make it twice in the same PR to close a nit. It wants its own PR with a `bun run fix` behind it. Happy to open one.

Separately, and worth recording since it affects my earlier claims: `bun run check` now includes `check:mutation-guards` (CLAUDE.md was updated on `main` mid-PR), and I had not been running it in my manual list. I've run it now — it passes, 6 invariants each proved to have a detector — so the earlier "all checks green" claims were incomplete rather than wrong. Current run: 3,587 tests, all ten `check:*` green, same one pre-existing root/`chmod` failure.

---

ignaciohermosillacornejo:
All four fixed in `a406cb0`. Finding 1 is the previous commit's own finding one layer out, which is the uncomfortable pattern here — I split "readable" into two claims in the code and left the message making one.

**Finding 1.** For the three spellings `069a4ac` *added*, the message named a cause that doesn't apply and a remedy already followed. Two sentences now, the second explaining why (a scan over `git ls-files` can only guard paths git emits). You were right that the test couldn't catch it — `toContain('cannot read as a directory')` is the half true for every row — so the table is split per cause with a guard that the two expected sentences differ. Mutation-checked by routing outside-repo tokens through the shell-syntax message: the new assertions fail.

**Findings 2 and 4 are one rewrite.** `readableCleanTokens` was `const isGenerated = isGeneratedPath` again, in the commit that deleted that one. And your asymmetry argument is the part that made it worth more than a rename: a token in *both* lists is loud, a token in *neither* vanishes from `generatedDirs` unreported while the success line still prints a count. One partition pass, three buckets, exhaustive by construction. I kept the raw spelling for the messages and wrote the comment you suggested so it isn't simplified back.

**Finding 3** was straightforwardly wrong prose — quotes are rejected, not normalised, and the sentence sat next to a test comment reasoning from the opposite.

**On `scripts/` formatting:** you're right that I fixed the instance and left the class, and right that the parallel to my own `process.exit` argument is uncomfortable. The difference I'd defend: that argument was about a class *this PR created*, and the fix lived in the same file. This one predates the branch and the fix rewrites 21 files I haven't touched. I've said so in the commit message rather than leaving the glob to be inferred, which is what you asked for — but `package.json` can't carry a comment, so the decision only exists in the PR history. That's a real weakness in my answer and I'd rather the repo owner picked: widen the globs here and eat the noise, or take it as a follow-up. I'll do either.

Worth stating plainly: this is round six on a gate I added mid-PR, and the last four rounds have been me fixing my own fixes. The #708 change this PR is actually for has been settled since `5658e97`. Each finding has been correct and each fix small, so I've kept going — but the convergence is in the hardening, not the subject, and splitting `check-tracked-files.ts` into its own PR is a reasonable call if you'd prefer the boot-probe fix to land.

3,588 tests pass, all ten `check:*` green, same one pre-existing root/`chmod` failure.

---

ignaciohermosillacornejo:
All three plus the nit fixed in `3b5e048`. Finding 1 is the one that should not have happened.

**I verified it rather than reasoning about it.** Collapsing both `console.error` branches into one message containing both phrases passes all 27 tests — my "guards the gate" test included. So the regression it was written to catch survives it, exactly as you said. Two commits after deleting the `.catch` test for being unfalsifiable, I shipped an unfalsifiable guard, which makes that standard look like something I apply when convenient. The claim now lives in the loop that runs the script, as the negative assertion: each group's stderr must contain its own phrase and *not* the other's. Same collapse now fails on the first row. The table-shape checks survive as anti-vacuity on the loop, retitled, since you're right that's a different claim.

**Finding 2 is the enumeration trap one notch in**, and `~/dist` being a row in my own table while the message named four causes it doesn't have is the part that stings. Stated positively now. I checked it end-to-end against a synthetic repo whose `clean` carries `~/dist ../build` — both messages print, each naming its own cause:

```
… names `~/dist`, which this script cannot read as a directory: a target must be a plain path
of letters, digits, `.`, `-`, `_` and `/`. Shell syntax is not interpreted — operators, globs,
braces, quoting, `~` and escapes all reach this script verbatim. …

… names `../build`, which are plain paths but not ones inside this repository. This gate is a
scan over `git ls-files`, which emits no leading `/` and no `.`/`..` segment, so such a target
would be counted as generated and then never checked. …
```

**Finding 3 and the nit** — "Both of these" over three exits, `const quoted` wedged between the paragraph and its subject, and the singular over a joined list — all corrected.

Standing observation, unchanged from my last comment: this is round seven on a gate I added mid-PR, and the #708 fix has been settled since `5658e97`. The findings keep being correct, so I keep acting on them, but three of the last four rounds have been my own fixes needing fixes. That's a signal about the rate I'm introducing these, not just about the review catching them. Splitting `check-tracked-files.ts` out so the boot-probe fix can land remains available and I'd take that call.

3,588 tests, all ten `check:*` green, same one pre-existing root/`chmod` failure.

---

ignaciohermosillacornejo:
All three fixed in `2eb019c`, and the PR body is refreshed — it had been stale since `24fa684` (wrong test count, no mention of the seam, the `.catch`, or the second class).

Note 1 is the sharpest of the three: the contrast-by-`find` was the complementary-by-hand shape I removed from the script one commit earlier, reappearing in its own test. `filter` plus an inner loop now. Your aside that the uniqueness assertion is load-bearing for that selection was right and it had no stated reason — it does now.

Note 2 was one word, and the categorical argument is the one that matters: `\w` without `u` is ASCII-only, so the message promised a rule a non-ASCII name satisfies. Note 3 — the file typechecks clean under `tsconfig.tests.json`, so it's on the include list with its siblings' reasoning and pinned in `MEMBERSHIP_IS_THE_COVERAGE`.

**On splitting: you've convinced me, and I'll withdraw the suggestion.** Your argument is better than mine — the gate exists *because* of a mistake made in this PR, and this repo's bug-response ritual is what says the class gets a detector. Moving it to a follow-up would separate the detector from the incident that motivated it and force the whole argument to be re-made from context that no longer exists. I was weighing round-count against that and had the trade backwards.

**On `scripts/` formatting:** agreed it wants an issue so the decision isn't only in PR history. I haven't opened one — filing on the owner's repo unprompted isn't mine to do — but the PR body now states the gap and the reason explicitly under "Known gap", so it's at least recorded where a reader of this change will find it. Same for the two other follow-ups this PR surfaced and declined: `FirebaseAuth` in-flight de-duplication (a tool call during a slow probe starts a second cold extraction and can double the candidate budget sent to Google's endpoint), and asserting each generated directory is *ignored*, which would have gone red on `main` before anything was ever committed.

3,588 tests, all ten `check:*` green including `check:mutation-guards`, same one pre-existing root/`chmod` failure.

---

ignaciohermosillacornejo:
Both plus the nit fixed in `53dcac9`.

**Answering the one claim you flagged as most wanting CI:** the new include-list entry *is* type-clean — I ran `tsc -p tsconfig.tests.json` locally before adding it, which is what the "typechecks clean, so it joins the list" line in `2eb019c` was based on. Not asking you to take that on trust either; CI's Quality Checks on this SHA is the independent answer.

Finding 1 is the nicest instance of the class this branch keeps hitting, because the paragraph *argues against its own failure mode*: "named rather than counted, since a count drifts" — then drifts, as an enumeration, further than a count would have. The fix is the same move as `GENERATED_PREFIXES` and `readableCleanTokens`: the membership already exists once, as an inline comment per entry on the line where a reason can't separate from its subject. The header now points there instead of keeping a second copy. You were right that it's pre-existing and fine to decline — I took it because this PR widened it.

The title nit was the same staleness shape as "Both of these stop the world" over three exits, one file over. And the 101-column line: worth noting it re-broke within two commits of that file being hand-reflowed to zero, which is a sharper argument for putting `scripts/` behind the formatter than the 87-line count was. Still its own PR.

**On splitting — we agree, and it's settled:** not splitting. Your argument convinced me last round and your restatement holds: separating the detector from the incident that motivated it would strand both.

3,588 tests, all ten `check:*` green, same one pre-existing root/`chmod` failure.

---

ignaciohermosillacornejo:
Correct, and it's the sharpest version of this yet: the commit that fixed "prose asserting something the file doesn't bear out" introduced a fresh instance of it. I deleted the enumeration on the grounds that the naming belongs on the entries, and didn't put it on the entries. Fixed in `d157fe7`.

**Checking it properly found more than you named.** Seven pinned entries had no reason on their own line, not four. Two of those are genuinely fine — `exported-constants` and `no-hand-rolled-comment-strippers` sit under the `#691` block that heads their run and names them. But the auth trio and `uid-transition-sweep` were argued *only in the header*, which is the remote prose the move was supposed to retire. They now carry a block stating the `#722` reason.

So the pointer is widened to what the file actually does — "on its own line, or in the block heading the run it belongs to" — rather than to a claim that happens to be true of the five entries I looked at. `tsconfig-tests-sync.test.ts` is named as the one deliberate exception, for the reason you gave.

And I verified it this time instead of asserting it: a walk over `MEMBERSHIP_IS_THE_COVERAGE` back to each entry's heading comment now reports zero without one. That check is what turned up the extra three, and writing it took less time than the round it would have saved.

3,588 tests, all ten `check:*` green, same one pre-existing root/`chmod` failure.

---

ignaciohermosillacornejo:
Correct, and worse than you framed it: the block I added to *finish* deleting a second copy **was** a second copy, forbidden by the paragraph I'd written one commit earlier. Fixed in `34044a2`.

Your point that the two arrived already divergent is the part that matters — header naming two subjects, my block naming three, with the third's file on the list. That's the argument for the rule rather than an illustration of it, and the copy nobody touched was the incomplete one.

Two things fell out of trimming the header:

- **The adopter-scope claim.** You were right that the four aren't adopters — I confirmed `grep -c createMockGraphQLClient` is 0 for all of them. The scope paragraph had been introducing them as part of the adopter set while paragraph 2 excluded them from it; the header now states which side they're on.
- **A cross-reference my own edit falsified.** `tsconfig-tests-sync.test.ts` said "the tsconfig header justifies their membership", which stopped being true the moment I moved that justification. Re-pointed at the run's block, with a note saying where it moved from — a stale cross-reference is precisely what this sequence keeps turning up.

I checked rather than asserted this time: one copy of the rationale remains in the tree, and one instance of its closing phrase.

**Being straight about the pattern:** that's four consecutive rounds where my fix for "prose asserting what the file doesn't bear out" introduced a fresh instance of the same thing. Each has been smaller than the last, which is convergence — but the rate is mine, not the review's. The mechanical check I wrote last round found three problems I'd have missed by inspection; this round I went back to inspection for the duplicate-prose question and missed two. The lesson holds better than I applied it.

3,588 tests, all ten `check:*` green, same one pre-existing root/`chmod` failure.

---

ignaciohermosillacornejo:
Fixed in `b2f5f5c`, and this one is almost too neat: the sentence arguing that a bare count over a fuzzy antecedent goes stale was carrying a count, which my own previous commit falsified. Your reconciliation was exact — the nine was thirteen minus the four auth files, defensible only under the reading that `34044a2` removed.

The history is what the paragraph is for and the arithmetic never was, so it's `several files named against the whole category` now. I also added the reason it carries no number, so the next person doesn't helpfully put one back.

**Where I think this leaves the PR.** The last five rounds have each been one sentence smaller than the one before, and this one is a word count in a comment. That's converged. Every behavioural claim in the diff is mutation-checked, every structural claim has a check that fails when it's false, and the prose claims are now down to ones that can't go stale because they assert nothing countable. I don't intend to keep hardening past this unless something surfaces that isn't self-referential.

3,588 tests, all ten `check:*` green, same one pre-existing root/`chmod` failure.

---

ignaciohermosillacornejo:
Fixed in `5d53f25`. I said last round I'd stop unless something surfaced that wasn't self-referential — this qualifies, so I took it: unlike the last several, this one is caused by the *change* rather than by a fix to it. `2467ffd` added the sixth importer (responding to your tree-walker finding) and never touched the map 48 lines above the entry that declares that very import.

You're right that nothing was at stake for the conclusions — all six are on the list and clean, so the safety reasoning beside the map holds. What was at stake is the map's stated job, for a reader deciding whether `ts-files.ts`'s own line is load-bearing. Sixth named, count dropped, with the reason it carries none.

One thing worth reporting because it cuts against my own last comment: I reconciled this mechanically instead of by eye, and the check flagged `ts-files.test.ts` as missing. It isn't — the map names it as *"its own contract test"* rather than by filename. So the check produced a false positive where inspection would have been right. That's the counterexample to the lesson I drew two rounds ago; the honest version is that mechanical checks catch omissions and miss descriptive references, which is why I reported the reconciliation rather than adding it as a gate.

3,588 tests, all ten `check:*` green, same one pre-existing root/`chmod` failure.

---

ignaciohermosillacornejo:
Declining this one, and I think the disagreement is substantive rather than me being tired of rounds.

`The two scanners` is not the same shape as `all five`. It's an **anaphor**, not a completeness claim — the two are named in the immediately preceding clause of the same row (`decoder-field-completeness.test.ts, dedup-identity.test.ts`), and "the two scanners" refers back to them the way a pronoun does. `all five` was different in kind: a standalone assertion that an enumeration was exhaustive, which my PR falsified by adding a sixth member elsewhere in the tree.

That difference decides the drift question. A third scanner importing `strip-comments.ts` has to be added to the enumeration two lines up — and the moment it is, "The two scanners" reads as broken English against the three names directly above it. The count and its antecedent can't be edited apart without the result being visibly wrong to whoever is making the edit. `all five` could, and did, because its members live in a different file.

So the consistency argument ("then the whole map is enumeration-only") is real but aesthetic, and the failure mode it's protecting against doesn't apply to this instance. Taking it would also be me applying a rule past the reasoning that justified it, which is a habit worth not acquiring on round thirteen.

**And that's me done with hardening on this PR.** Not because the findings stopped being correct — this one is a fair observation, I just think it's wrong on the merits — but because the bar I set two rounds ago was "something that isn't self-referential", and a nit about the phrasing of a comment adjacent to a comment I edited is as self-referential as it gets. If a behavioural issue surfaces I'll act on it immediately. Otherwise the PR is where I'd like a human to look at it.

3,588 tests, all ten `check:*` green, same one pre-existing root/`chmod` failure.

---

